PR #76, opened 2018-06-27
Fix/missing callback contexts

### LINKED ISSUE #45: Add context to callbacks that are lacking it
- [ ]  handle_prepare
- [ ]  handle_play
- [ ]  handle_stop
- [ ]  handle_other

### LINKED ISSUE #46: Add current playback state to contexts
It is held in state in some elements, what unnecesarily complicates code

bblaszkow06:
Closes #45 
Closes #46


lib/membrane/element/callback_context/caps.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  @type t :: %__MODULE__{
+          playback_state: Membrane.Mixins.Playback.state_t(),
+          caps: Membrane.Caps.t() | nil
+        }
+
+  defstruct playback_state: nil, caps: nil
+
+  @impl true
+  def from_state(state, entries) do
```
mat-hek: I would consider changing this to a macro. It would enable compile-time fields verification and slightly speed up structs creation. The same applies to all other contexts.

lib/membrane/element/callback_context.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  @doc """
+  Provides a proper Context struct for a callback.
+  """
+  @spec construct!(
+          context :: callback_t(),
+          state :: Membrane.Element.Manager.State.t(),
+          entries :: keyword()
+        ) :: struct()
+  def construct!(callback, state, entries \\ []) do
```
mat-hek: This is probably unused and can be removed.

PR #79, opened 2018-07-10
Refactor/element manager

mat-hek:


lib/membrane/core/element/pad_spec_handler.ex
```diff
@@ -0,0 +1,78 @@
...
+
+      {:ok, %State{state | pads: pads}}
+    else
+      {:error, reason} -> warn_error("Error parsing pads", reason, state)
+    end
+  end
+
+  defp handle_known_pads(known_pads_fun, direction, module) do
+    known_pads =
+      cond do
```
bblaszkow06: I know its copied, but since we're doing refactor... `cond` shuldn't be used when there are only 2 options, classic `if` would suffice. Even linter moans about it.

test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
bblaszkow06: Since we have ExUnit tests now, please add `mix test` to `.travis.yml`

mat-hek: Done

lib/membrane/core/playback_handler.ex
```diff
@@ -0,0 +1,173 @@
...
+
+  defp switch_target_lock(playbackable) do
+    playbackable
+    |> Playbackable.update_playback(&%Playback{&1 | target_locked?: not &1.target_locked?})
+  end
+
+  defp next_state(current, target) do
+    with {:ok, curr_pos} <-
+           @states_positions[current]
+           |> Helper.wrap_nil(:invalid_current_playback_state),
```
bblaszkow06: I see that helper for the 1st time and it's name is a complete :hankey:

bblaszkow06: ... but we can fix it outside of this PR

PR #82, opened 2018-08-06
Feature/payload types

### LINKED ISSUE #68: Support for different payload types
Currently we use only one payload type - erlang binary. We should support also a native C payload type, mutable and passed by pointer. This is necessary for processing large amount of data (such as high quality video).
We should consider:
- Way of passing pointer to different NIFs
- Monitoring lifecycle of payload
- Adding payload type to caps
- Size of each payload type should be accessible from elixir with possibly low overhead, because its used frequently to handle demand-related stuff.
- Payload should be automatically converted by core if type is not supported by element.

bblaszkow06:
Closes #68


lib/membrane/buffer.ex
```diff
@@ -8,17 +8,19 @@ defmodule Membrane.Buffer do
...
   alias __MODULE__
+  alias Membrane.Payload
 
-  @type payload_t :: bitstring
   @type metadata_t :: map
 
   @type t :: %Buffer{
-          payload: payload_t,
+          payload: Payload.t(),
+          type: :binary | :shm | :pointer,
```
mat-hek: Why is this field necessary? I don't think Core should be aware of all payload types. Also name `type` is a bit misleading, because it determines the type of payload, not type of buffer. Maybe it could be replaced with `type` function in `Payload` protocol, returning atom.

lib/membrane/payload.ex
```diff
@@ -0,0 +1,50 @@
+defprotocol Membrane.Payload do
+  @moduledoc """
+  """
+
+  @type t :: struct()
```
mat-hek: This protocol can be implemented for anything, and it is already for binary, so I would say `any`

lib/membrane/payload.ex
```diff
@@ -0,0 +1,50 @@
...
+  def split_at(payload, at_pos)
+
+  @doc """
+  Converts payload into binary
+  """
+  @spec to_binary(t()) :: binary()
+  def to_binary(payload)
+end
+
+defimpl Membrane.Payload, for: BitString do
```
mat-hek: I would move this to a separate file

lib/membrane/payload.ex
```diff
@@ -0,0 +1,50 @@
+defprotocol Membrane.Payload do
+  @moduledoc """
```
mat-hek: A sentence of explanation here would be a plus ;)

lib/membrane/payload.ex
```diff
@@ -0,0 +1,50 @@
...
+end
+
+defimpl Membrane.Payload, for: BitString do
+  @spec size(payload :: binary()) :: non_neg_integer
+  def size(data) when is_binary(data) do
+    data |> byte_size()
+  end
+
+  @spec split_at(binary(), non_neg_integer) :: {binary(), binary()}
+  def split_at(data, 0) do
```
mat-hek: This is already covered by the last clause. I am also not sure if we should require supporting this in protocol, as we do not use this, and implementing for other payloads might be problematic.

lib/membrane/payload.ex
```diff
@@ -0,0 +1,50 @@
...
+  @doc """
+  Returns total size of paload in bytes
+  """
+  @spec size(payload :: t()) :: non_neg_integer()
+  def size(payload)
+
+  @doc """
+  Splits the payload at given position (1st part has the size equal to `at_pos` argument)
+  """
+  @spec split_at(payload :: t(), at_pos :: non_neg_integer()) :: {t(), t()}
```
mat-hek: I would change `at_pos` type to `pos_integer`

lib/membrane/payload.ex
bblaszkow06: I followed the convention that an implementation has to be either where the struct is defined or the protocol is defined. This is how its done in Elixir itself (See: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/enum.ex)

lib/membrane/payload.ex
```diff
@@ -1,44 +1,51 @@
...
-  Returns total size of paload in bytes
+  Returns total size of payload in bytes
   """
   @spec size(payload :: t()) :: non_neg_integer()
   def size(payload)
 
   @doc """
   Splits the payload at given position (1st part has the size equal to `at_pos` argument)
+
+  `at_pos` has to be greater than 0 and smaller than the size of payload. This guarantees
```
mat-hek: It sounds like a user would have to guarantee that `at_pos` is valid. I would rather write "`at_pos` _is_ _always_ greater than 0..."

lib/membrane/payload.ex
```diff
@@ -1,44 +1,51 @@
...
+  returned payloads are never empty.
+
+  When such conditions are not met, the function should raise.
```
mat-hek: I would remove this sentence, as it is responsibility of the core to invoke these functions with proper arguments. Additional checks might be performed to be sure. Going current way, we should also write that function should raise e.g. if a string is passed instead of integer.

lib/membrane/payload.ex
```diff
@@ -1,44 +1,51 @@
...
   """
-  @spec split_at(payload :: t(), at_pos :: non_neg_integer()) :: {t(), t()}
-  def split_at(payload, at_pos)
+  @spec split_at!(payload :: t(), at_pos :: pos_integer()) :: {t(), t()}
+  def split_at!(payload, at_pos)
```
mat-hek: I think this exclamation is unnecessary. Specificaton of the function does not impose that any errors may happen

lib/membrane/payload.ex
```diff
@@ -1,44 +1,51 @@
...
   Converts payload into binary
   """
   @spec to_binary(t()) :: binary()
   def to_binary(payload)
+
+  @doc """
+  Returns an atom describing type of the payload.
+  """
+  @spec type(t()) :: atom()
+  def type(payload)
```
mat-hek: Shouldn't this function have no arguments?

lib/membrane/payload.ex
```diff
@@ -0,0 +1,58 @@
...
+  @doc """
+  Returns total size of payload in bytes
+  """
+  @spec size(payload :: t()) :: non_neg_integer()
+  def size(payload)
+
+  @doc """
+  Splits the payload at given position (1st part has the size equal to `at_pos` argument)
+
+  `at_pos` has to be greater than 0 and smaller than the size of payload. This guarantees
```
mat-hek: It sounds like a user would have to guarantee that at_pos is valid. I would rather write "at_pos is always greater than 0..."

PR #83, opened 2018-08-06
Added  option to a demand request from a filter

nobikik9:
Bugfix & added `set_to` option for demands from filters

### DIFF
--- a/lib/membrane_core/element/manager/filter.ex
+++ b/lib/membrane_core/element/manager/filter.ex
@@ -131,7 +131,6 @@ defmodule Membrane.Element.Manager.Filter do
     """)
   end
 
-
   def handle_buffer(:push, pad_name, buffers, state) do
     handle_process_push(pad_name, buffers, state)
   end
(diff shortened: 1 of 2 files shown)


mat-hek:
Please also fix typespec here: https://github.com/membraneframework/membrane-core/blob/v0.1/lib/membrane_core/element/action.ex#L94

lib/membrane_core/element/manager/filter.ex
```diff
@@ -106,8 +111,31 @@ defmodule Membrane.Element.Manager.Filter do
...
+    {:ok, state} = state |> update_sink_self_demand(pad_name, source, update_demand_func)
+
+    handle_process_pull(pad_name, source, buf_cnt, state)
+    |> or_warn_error("""
+    Demand of size #{inspect(buf_cnt)} on sink pad #{inspect(pad_name)}
+    was raised, and handle_process was called, but an error happened.
+    """)
+  end
+
+  def handle_self_demand(pad_name, source, :set, buf_cnt, state) do
```
mat-hek: This is now redundant ;)

PR #84, opened 2018-08-24
Changes required by Shm payload

bblaszkow06:


lib/membrane/payload.ex
```diff
@@ -23,36 +52,95 @@ defprotocol Membrane.Payload do
...
+  def split_at(payload, at_pos)
+      when is_binary(payload) and 0 < at_pos and at_pos < byte_size(payload) do
+    <<part1::binary-size(at_pos), part2::binary>> = payload
     {part1, part2}
   end
 
+  @impl true
+  @spec concat(left :: binary(), right :: binary()) :: binary()
+  def concat(left, right) when is_binary(left) do
+    left <> Payload.to_binary(right)
```
mat-hek: I think we should assume that both payloads are binaries, as pointed out in typespec

PR #91, opened 2018-09-07
Added possibility to create {:atom, id} element names

nobikik9:
It should be possible to create {:atom, id} element names instead of single atoms.

### DIFF
--- a/lib/membrane_core/pipeline.ex
+++ b/lib/membrane_core/pipeline.ex
@@ -278,21 +278,38 @@ defmodule Membrane.Pipeline do
     end
   end
 
+  defp child_name?(name) when is_atom(name), do: true
+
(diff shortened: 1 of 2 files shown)


lib/membrane_core/pipeline.ex
```diff
@@ -278,21 +278,38 @@ defmodule Membrane.Pipeline do
     end
   end
 
+  defp child_name?(name) when is_atom(name), do: true
```
bblaszkow06: I would define that as private guard (using `defguard`)

bblaszkow06: This will allow to only replace `is_atom` with `is_child_name` in the following `parse_child` calls

PR #92, opened 2018-09-07
add cool badges

mkaput:
I believe these are useful when a visitor first lands on GitHub page and then wants to check Hex page & API docs.


bblaszkow06:
Thank you stranger for your contribution!

PR #93, opened 2018-09-07
Deleted :indexed option for children

nobikik9:


lib/membrane_core/pipeline.ex
```diff
@@ -301,13 +301,8 @@ defmodule Membrane.Pipeline do
 
   defp resolve_children(children, state), do: children |> Enum.map_reduce(state, &resolve_child/2)
```
mat-hek: This is now unnecessary

mat-hek:
Also everything in `Membrane.Pipeline.State` that is related to children ids should be removed

PR #94, opened 2018-09-10
Feature/unified demands

### LINKED ISSUE #66: Consider unifying default behaviour of `demand` action and `handle_demand` callback between different element types
Currently:
- in filter, `handle_demand` receives always entire demand, and returning `demand` overrides previous demand
- in source, `handle_demand` receives only the new demand, except for being caused by `redemand` action, in which case it receives entire demand
- in sink, returning `demand` means adding it to existing demand

Imho we should change this, so that:
- entire demand is always passed to `handle_demand`
- returning `demand` always causes overriding existing demand

As in some cases different behaviour is required, we could:
- support passing lambda as a size to demand, such as `demand: {:sink, & &1+1}`
- pass the new demand to `handle_demand` in context

bblaszkow06:
Closes #66

### DIFF
--- a/lib/membrane/core/element/demand_handler.ex
+++ b/lib/membrane/core/element/demand_handler.ex
@@ -51,8 +51,8 @@ defmodule Membrane.Core.Element.DemandHandler do
   end
 
   @doc """
-  Based on the demand on given pad takes buffers and passes it to proper
-  controller.
+  Based on the demand on the given pad takes PullBuffer contents
+  and passes it to proper controllers.
   """
   @spec supply_demand(
           Pad.name_t(),
@@ -64,7 +64,7 @@ defmodule Membrane.Core.Element.DemandHandler do
   end
 
   @doc """
-  Supplies the demand requested on the given sink pad, if there are any.
+  Supplies the demand requested on the given sink pad, if there is any.
 
   In filters also triggers `handle_demand` callback when there is unsupplied demand
   on source pads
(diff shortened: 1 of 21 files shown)


lib/membrane/core/element/demand_handler.ex
```diff
@@ -20,47 +20,59 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-  def handle_demand(pad_name, :self, :increase, size, state) do
-    {total_size, state} =
-      PadModel.get_and_update_data!(
+  def update_demand(pad_name, size_fun, state) when is_function(size_fun) do
+    state =
+      PadModel.update_data!(
         pad_name,
         :demand,
-        fn demand -> (demand + size) ~> {&1, &1} end,
+        fn demand -> size_fun.(demand) ~> {:ok, &1} end,
```
mat-hek: we should check if resulting demand is non-negative

lib/membrane/core/element/demand_handler.ex
```diff
@@ -34,20 +34,25 @@ defmodule Membrane.Core.Element.DemandHandler do
...
+        end
+      end,
+      state
+    )
   end
 
   @doc """
-  Based on the demand on given pad takes buffers, passes it to proper
-  controller and checks if source demand has been suppplied.
+  Based on the demand on given pad takes buffers and passes it to proper
```
mat-hek: It takes not only buffers but also events and caps

lib/membrane/core/element/demand_handler.ex
```diff
@@ -59,7 +64,10 @@ defmodule Membrane.Core.Element.DemandHandler do
   end
 
   @doc """
-  Handles demands requested on given sink pad, if there are any.
+  Supplies the demand requested on the given sink pad, if there are any.
```
mat-hek: *there is any

PR #100, opened 2018-09-19
Fix/pad ref

mat-hek:
Closes #44

PR #102, opened 2018-09-20
Feature/pad directions names

mtsznowak:
should be merged after PR #100


spec/support/element/trivial_sink.ex
```diff
@@ -7,7 +7,7 @@ defmodule Membrane.Support.Element.TrivialSink do
 
   use Membrane.Element.Base.Sink
 
-  def_sink_pads sink: [caps: :any, demand_in: :buffers]
+  def_input_pads in: [caps: :any, demand_in: :buffers]
```
mat-hek: Maybe we should rename `demand_in` to `demand_unit`, as `in` suffix can be misleading when we have `in` pads

PR #103, opened 2018-09-21
Refactor/events

mat-hek:


lib/membrane/event_protocol.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  @doc """
+  Determines whether event is synchronized with buffers. Defaults to true.
+
+  Buffers and synchronized events are always received in the same order they are
+  sent. Non-synchronized events skip buffers waiting in internal buffers
+  (such as `Membrane.PullBuffer`) and are handled first. In other words,
+  they have priority over buffers.
+  """
+  @spec synchronized?(t) :: boolean
```
bblaszkow06: I would rather call is `synchronous` or negate its meaning and rename to `async`

PR #105, opened 2018-09-27
 returning :redemand from handle_process/write in filters

mtsznowak:


test/membrane/demands_test.exs
```diff
@@ -4,8 +4,18 @@ defmodule Membrane.Integration.DemandsTest do
   alias Membrane.Integration.{TestingFilter, TestingSource, TestingSink, TestingPipeline}
   alias Membrane.Pipeline
 
+  def assert_message(pattern) do
```
bblaszkow06: I believe a line of comment why this is used instead of `assert_receive` would be appreciated

lib/membrane/core/element/action_handler.ex
```diff
@@ -161,12 +161,13 @@ defmodule Membrane.Core.Element.ActionHandler do
...
-        actions |> join_buffers(),
-        callback,
-        handler_params,
-        state
-      )
+      case super(actions |> join_buffers(), callback, handler_params, state) do
+        {:ok, state} ->
+          invoke_redemands(state, handler_params)
+
+        {:error, reason} ->
```
bblaszkow06: Invalid pattern, misses state

lib/membrane/core/element/event_controller.ex
```diff
@@ -98,4 +101,15 @@ defmodule Membrane.Core.Element.EventController do
   end
 
   defp handle_special_event(_pad_ref, _event, state), do: {{:ok, :handle}, state}
+
+  defp post_callback_exec(%Event.EndOfStream{}, %{type: :sink} = state) do
+    ActionHandler.handle_action(
```
bblaszkow06: DIalyzer yells here about invalid types because `{:notification, _}` is not included in succsess typing for 1st argument (which looks like `Action.t()`
```elixir
{:buffer
   | :caps
   | :demand
   | :event
   | :forward
   | :notify
   | :playback_change
   | :redemand
   | :split, _},
```
)

PR #108, opened 2018-10-12
Refactor/message

mat-hek:


lib/membrane/core/message.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Membrane.Core.Message do
+  @moduledoc false
+  # Record representing membrane internal message
+
+  require Record
+
+  Record.defrecord(:message, __MODULE__, type: nil, args: [])
```
bblaszkow06: Isn't that atom too generic? Shouldn't it be prefixed with `membrane_`?

PR #110, opened 2018-10-18
Pads context

mat-hek:
This should be merged after #109


lib/membrane/element/callback_context/caps.ex
```diff
@@ -3,22 +3,6 @@ defmodule Membrane.Element.CallbackContext.Caps do
...
-  defmacro from_state(state, entries) do
-    quote do
-      %unquote(__MODULE__){
-        unquote_splicing(entries),
-        playback_state: unquote(state).playback.state
-      }
-    end
-  end
+  use Membrane.Element.CallbackContext,
+    caps: Membrane.Caps.t() | nil
```
bblaszkow06: Does keeping the caps here make any sense considering they can be accessed via `:pads_data`

bblaszkow06: Applies to all contexts

lib/membrane/core/element/caps_controller.ex
```diff
@@ -34,9 +34,9 @@ defmodule Membrane.Core.Element.CapsController do
   @spec exec_handle_caps(Pad.ref_t(), Caps.t(), params :: map, State.t()) ::
           State.stateful_try_t()
   def exec_handle_caps(pad_ref, caps, params \\ %{}, state) do
-    %{accepted_caps: accepted_caps, caps: old_caps} = PadModel.get_data!(pad_ref, state)
+    %{accepted_caps: accepted_caps} = PadModel.get_data!(pad_ref, state)
 
-    context = CallbackContext.Caps.from_state(state, caps: old_caps)
+    context = CallbackContext.Caps.from_state(state)
```
bblaszkow06: Consider bringing back old_caps field into this context

lib/membrane/element/callback_context/caps.ex
```diff
@@ -1,7 +1,11 @@
 defmodule Membrane.Element.CallbackContext.Caps do
   @moduledoc """
   Structure representing a context that is passed to the element when receiving
-  information about new caps for given pad
+  information about new caps for given pad.
+
+  The `pads[pad].caps` field is set to the old caps and has the same value as
```
bblaszkow06: This sentence suggests it changes something in `pads[]`. I would rather say old_caps is set to `pads[pad].caps`

PR #111, opened 2018-10-25
Add default value for an options in generated docs

bblaszkow06:


bblaszkow06:
This is how the generated docs look: image

lib/membrane/element/base/mixin/common_behaviour.ex
```diff
@@ -199,8 +199,19 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
...
       options
       |> Enum.map_join("\n", fn {k, v} ->
-        "* `#{Atom.to_string(k)}`: #{Keyword.get(v, :description, "\n")}"
-        |> String.trim()
+        desc =
+          v
+          |> Keyword.get(:description, "")
+          |> String.trim()
+
+        tail =
```
mat-hek: `tail` is not the best name here,  maybe `desc_default` would be better

lib/membrane/element/base/mixin/common_behaviour.ex
```diff
@@ -199,8 +199,19 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
...
+          if Keyword.has_key?(v, :default) do
+            "\n\n  Defaults to `#{inspect(v[:default])}`"
+          else
+            ""
+          end
+
+        "* `#{Atom.to_string(k)}`: #{desc}" <> tail
```
mat-hek: use either string interpolation or concatenation, mixing them looks messy

PR #125, opened 2018-12-04
Added new testing facilities.

Hajto:
Added TestingConfigurablePipeline and modified test sink to place demands autoamtically when configured to do so.


mat-hek:
Maybe before we merge this write some tests in some element (UDP?) using these utils to check if they work (I think writing tests for tests is a bit overkill as for now)

mat-hek:
Maybe before we merge this write some tests in one element (UDP?) using these utils to check if they work (I think writing tests for tests is a bit overkill as for now)

mat-hek:
Maybe before we merge this write some tests in one element (UDP?) using these utils to check if they work (I think writing tests for tests is a bit overkill as for now)

.gitignore
```diff
@@ -127,6 +127,7 @@ tags
 !.vscode/launch.json
 !.vscode/extensions.json
 .history
+.vscode/settings.json
```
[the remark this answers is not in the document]
Hajto: My derp. The git was seing this file because it was previously explictly unignored. I will ammend this changes to reduce clutter.

lib/membrane/integration/testing_configurable_pipeline.ex
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  @doc """
+  Sends message to a child by Element name.
+
+  ## Examples
+
+      message_child(pipeline, :sink, {:message, "to handle"})
+  """
+
```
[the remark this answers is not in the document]
Hajto: Which empty line? The one between examples and code?

lib/membrane/integration/testing_configurable_pipeline/assertions.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  The `pattern` argument must be a match pattern. Flunks with `failure_message`
+  if a message matching `pattern` is not received.
+  """
+  defmacro receive_message(
+             pattern,
+             timeout \\ Application.fetch_env!(:ex_unit, :assert_receive_timeout),
+             failure_message \\ nil
+           ) do
+    import ExUnit.Assertions, only: [assert_receive: 1]
```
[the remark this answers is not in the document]
Hajto: This results with warning because I messed up contexts. It does import ExUnit.Assertions into Assertions context not into caller context.

I think that my options are:

  - Importing inside `quote`, but that would import it every time
  - Write using macro which imports both `Pipeline.Assertions` and `ExUnit.Assertions`

What would be preferable solution?

lib/membrane/integration/testing_configurable_pipeline.ex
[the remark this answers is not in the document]
bblaszkow06: The one @mat-hek left the comment on. Or speaking more precisely: between docs for a function and the function itself. I don't like it either. If anywhere, that empty line could be after the example but **before** triple quote ending the docs

lib/membrane/integration/testing_configurable_pipeline/assertions.ex
```diff
@@ -0,0 +1,22 @@
...
+
+    quote do
+      assert_receive {Membrane.Integration.TestingConfigurable.Pipeline, unquote(pattern)},
```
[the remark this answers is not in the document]
bblaszkow06: There is no such module as `Membrane.Integration.TestingConfigurable.Pipeline` but there is `Membrane.Integration.TestingConfigurablePipeline`. A dot makes the difference.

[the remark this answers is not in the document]
bblaszkow06: As for the name itself, I really hate the current naming, it reminds me of Java and that's not a good thing. I would remove `Integration` instead and go for something like `Membrane.Testing.ConfigurablePipeline` or with `Test` instead of `Testing` @mat-hek

lib/membrane/integration/testing_data_source.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Membrane.Element.TestingDataSource do
```
bblaszkow06: Name of this element seems off. I believe all the testing elements should be a submodule of the same module. So far, other elements and pipeline are inside `Integration`.

lib/membrane/integration/testing_configurable_pipeline/assertions.ex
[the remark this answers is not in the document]
bblaszkow06: And one more: why `ConfigurablePipeline`? Is there a non-cofigurable version available for users? This one should be called TestingPipeline to be consistent with elemets in that module. Of course, that would mean the `TestingPipeline` from support should be renamed, but it will be now test-specific and I would rename it to sth like `Membrane.DemandTest.(Testing)Pipeline`

lib/membrane/integration/testing_configurable_pipeline/assertions.ex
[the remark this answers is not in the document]
bblaszkow06: Since both espec and ExUnit provide `assert_received` I would go for the last solution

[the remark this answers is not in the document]
Hajto: By `that would import it every time` I meant that each `assert_receive_message` call would generate single `import` expression. 

Illustrated by http://tpcg.io/LAiJM3

.gitignore
```diff
@@ -88,7 +88,7 @@ erl_crash.dump
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
```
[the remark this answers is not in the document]
Hajto: I disabled clrf and `curl` fresh `.gitignore` from master. The is no sign in the `changed files` section that there is missing \r\r

Hajto:
> Maybe before we merge this write some tests in some element (UDP?) using these utils to check if they work (I think writing tests for tests is a bit overkill as for now)

These elements are based on what I wrote for UDP element

bblaszkow06:
Fix the build first https://travis-ci.com/membraneframework/membrane-core/builds/94304322

lib/membrane/testing/pipeline.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Membrane.Testing.Pipeline do
+  @moduledoc """
+  Provides a testing utility pipeline which can be easily configured from test.
```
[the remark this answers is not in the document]
Hajto: @bblaszkow06 Is the documentation extensive enough now?

lib/membrane/testing/pipeline.ex
```diff
@@ -77,6 +141,17 @@ defmodule Membrane.Testing.Pipeline do
     :ok
   end
 
+  @doc """
+  Starts the Testing Pipeline and links it to the current process.
+
+  Proxy of `Membrane.Pipeline.start_link/3`.
```
bblaszkow06: Rather proxy for

lib/membrane/testing/pipeline.ex
```diff
@@ -1,6 +1,64 @@
...
+  ```
+  options = %Membrane.Testing.Pipeline.Options {
+    monitored_callbacks: [:handle_prepared_to_playing],
+    test_process: pid
+    ...
+  }
+  ```
+
+  First, you need to configure which callbacks invocation you want to know about.
+  You can do that by putting their name into `monitored_callbacks` field of `Options` struct.
```
bblaszkow06: their names

lib/membrane/testing/pipeline.ex
```diff
@@ -1,6 +1,64 @@
...
+  You also need to pass `PID` of the process that will receive messages.
```
bblaszkow06: **a** pid

lib/membrane/testing/pipeline.ex
```diff
@@ -1,6 +1,64 @@
...
+
+  ```
+  import Membrane.Testing.Pipeline.Assertions
+  assert_receive_message :handle_prepared_to_playing
+  ```
+
+  Then it is matter of waiting for right message to come. You can do that easily by using `Membrane.Testing.Pipeline.Assertions.assert_receive_message/3`.
```
bblaszkow06: **a** matter of waiting for **the** right...
Plus, split the line

lib/membrane/caps/matcher.ex
```diff
@@ -126,7 +126,6 @@ defmodule Membrane.Caps.Matcher do
 
   When `:any` is used as specs, caps can by anything (i.e. they can be invalid)
   """
-  @spec match?(:any, any()) :: true
```
bblaszkow06: This line should not be removed

lib/membrane/testing/source.ex
```diff
@@ -13,8 +13,10 @@ defmodule Membrane.Testing.Source do
                 type: :function,
                 spec: (non_neg_integer, non_neg_integer -> [Membrane.Action.t()]),
                 default: &__MODULE__.default_buf_gen/2,
-                description:
-                  "Action generator takes two arguments. First is counter which is incremented by 1 every call and second argument represents size of demand."
+                description: """
+                Action generator takes two arguments.
```
bblaszkow06: I would add an info that it's invoked on each handle_demand

lib/membrane/testing/pipeline.ex
```diff
@@ -25,9 +84,12 @@ defmodule Membrane.Testing.Pipeline do
     If links is not present or set to nil it will be populated automatically based on elements order using default pad names.
     """
 
-    @enforce_keys [:elements, :test_process]
-    defstruct @enforce_keys ++ [:monitored_callbacks, :links]
+    @enforce_keys [:elements]
+    defstruct @enforce_keys ++ [:monitored_callbacks, :links, :test_process]
```
bblaszkow06: You've introduced a potential bug here. If `:test_process` is nil, the send from `notify_parent` will fail. Therefore, `notify_parent` needs a guard for `nil` in `:test_process`

Hajto: I will add clause to init that returns error in case someone puts `monitor_callbacks` and does not provide `pid`

lib/membrane/integration/testing_configurable_pipeline/assertions.ex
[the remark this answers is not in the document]
mat-hek: Actually, we all got it wrong :P the import is placed in the proper place and works just as I described above. The problem is that specified arity of `assert_receive` is invalid (should be 3) @Hajto @bblaszkow06

test/support/demand_test/filter.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Membrane.Integration.TestingFilter do
+defmodule Membrane.DemandTest.Filter do
```
mat-hek: I would change the name to `Membrane.Support.DemandsTest.Filter`
- because now we have tests split into lib, test/support and actual tests so when the module name doesn't correspond to its location it's hard to determine what is being used
- to be consistent with `Membrane.Integration.DemandsTest`

test/support/demand_test/pipeline.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Membrane.Integration.TestingPipeline do
+defmodule Membrane.DemandTest.Pipeline do
```
mat-hek: `Membrane.Support.DemandsTest.Pipeline`

test/membrane/demands_test.exs
```diff
@@ -1,7 +1,10 @@
 defmodule Membrane.Integration.DemandsTest do
```
mat-hek: Move to `test/membrane/integration/demands_test`

lib/membrane/testing/pipeline.ex
```diff
@@ -0,0 +1,231 @@
...
+
+      iex> Pipeline.populate_links([el1: MembraneElement1, el2: MembraneElement2])
+      %{{:el1, :output} => {:el2, :input}}
+  """
+  @spec populate_links(elements :: Spec.children_spec_t()) :: Spec.links_spec_t()
+  def populate_links(elements) do
+    [first_name | element_names] = Enum.map(elements, fn {name, _} -> name end)
+
+    {links, _} =
+      Enum.reduce(element_names, {%{}, first_name}, fn x, {links, previous_element} ->
```
mat-hek: You can leave it as is, but there is a nice function for handling such cases
```
iex(5)> [1,2,3,4] |> Enum.chunk_every(2, 1, :discard)
[[1, 2], [2, 3], [3, 4]]
```

Hajto: Thanks, I like this approach much better.

test/support/demand_test/filter.ex
mat-hek: This is still called DemandTest instead of Demand**s**Test

test/support/demand_test/pipeline.ex
mat-hek: This is still called DemandTest instead of Demand**s**Test

Hajto: This time I made sure there is that `s` on the end :)

PR #134, opened 2018-12-14
added :default_formatter option parameter and Membrane.Time.pretty_du…

mat-hek:
…ration/1

### DIFF
--- a/lib/membrane/element/base/mixin/common_behaviour.ex
+++ b/lib/membrane/element/base/mixin/common_behaviour.ex
@@ -7,9 +7,11 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
 
   For more information on implementing elements, see `Membrane.Element.Base`.
   """
-  alias Membrane.{Action, Core, Element, Event}
+  alias Membrane.{Action, Core, Element, Event, Time}
   alias Core.CallbackHandler
   alias Element.{Action, CallbackContext, Pad}
+
+  use Bunch
 
   @typedoc """
   Type that defines all valid return values from most callbacks.
@@ -148,31 +150,14 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
   """
   @callback handle_shutdown(state :: Element.state_t()) :: :ok
 
-  @default_quoted_specs %{
-    atom:
-      quote do
-        atom()
-      end,
-    boolean:
-      quote do
-        boolean()
-      end,
-    string:
-      quote do
-        String.t()
-      end,
-    keyword:
-      quote do
-        keyword()
-      end,
-    struct:
-      quote do
-        struct()
-      end,
-    caps:
-      quote do
-        struct()
-      end
+  @default_types_params %{
+    atom: [spec: quote_expr(atom)],
+    boolean: [spec: quote_expr(boolean)],
+    string: [spec: quote_expr(String.t())],
+    keyword: [spec: quote_expr(keyword)],
+    struct: [spec: quote_expr(struct)],
+    caps: [spec: quote_expr(struct)],
+    time: [spec: quote_expr(Time.t()), inspector: &Time.pretty_duration/1]
   }
 
   @doc """
@@ -185,12 +170,12 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
 
     * `type:` atom, used for parsing
     * `spec:` typespec for value in struct. If ommitted, for types:
-      `#{inspect(Map.keys(@default_quoted_specs))}` the default typespec is provided.
+      `#{inspect(Map.keys(@default_types_params))}` the default typespec is provided.
       For others typespec is set to `t:any/0`
     * `default:` default value for option. If not present, value for this option
       will have to be provided each time options struct is created
-    * `default_formatter:` function converting value provided as `default` to a string.
-      Used when creating documentation instead of `inspect/1`
+    * `inspector:` function converting fields' value to a string. Used when
+      creating documentation instead of `inspect/1`
     * `description:` string describing an option. It will be present in value returned by `options/0`
       and in typedoc for the struct.
   """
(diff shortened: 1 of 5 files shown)


lib/membrane/time.ex
```diff
@@ -60,6 +60,34 @@ defmodule Membrane.Time do
   """
   defguard is_native_t(value) when is_integer(value)
 
+  @doc """
+  Returns duration as a string with unit. Chosen unit is the biggest possible one
+  that doesn't involve precission loss.
+
+  ## Examples
+
+      iex> 10 |> #{inspect(__MODULE__)}.milliseconds() |> #{inspect(__MODULE__)}.pretty_duration()
```
bblaszkow06: You can format the tests a bit nicer:
```elixir
      iex> 10
      ...> |> #{inspect(__MODULE__)}.milliseconds()
      ...> |> #{inspect(__MODULE__)}.pretty_duration()
      "10 ms"

      iex> 60_000_000
      ...> |> #{inspect(__MODULE__)}.microseconds()
      ...> |> #{inspect(__MODULE__)}.pretty_duration()
      "1 min"

```

lib/membrane/time.ex
```diff
@@ -60,6 +60,34 @@ defmodule Membrane.Time do
...
+      "10 ms"
+      iex> 60_000_000 |> #{inspect(__MODULE__)}.microseconds() |> #{inspect(__MODULE__)}.pretty_duration()
+      "1 min"
```
bblaszkow06: Add a test for days like this to at least ensure the whole array is traversed:
```elixir
      iex> 6
      ...> |> #{inspect(__MODULE__)}.days()
      ...> |> #{inspect(__MODULE__)}.pretty_duration()
      "6 d"

```

lib/membrane/time.ex
```diff
@@ -60,6 +60,34 @@ defmodule Membrane.Time do
...
+
+  """
+  @spec pretty_duration(t) :: String.t()
+  def pretty_duration(time) when is_t(time) do
+    [ns: 1000, us: 1000, ms: 1000, s: 60, min: 60, h: 24]
```
bblaszkow06: As we've discussed, I don't like this function very much, but it's OK. Anyway, here's my solution:
```elixir
  def pretty_duration(time) when is_t(time) do
    [ns: 1, us: 1000, ms: 1000, s: 1000, min: 60, h: 60, d: 24]
    |> Enum.scan(fn {unit, mult}, {_, acc} -> {unit, acc * mult} end)
    |> Enum.reverse()
    |> Enum.find_value(fn {unit, divisor} ->
      rem(time, divisor) == 0 && "#{time |> div(divisor)} #{unit}"
    end)
  end
```
I know it's a bit less efficient, but takes half of the lines compared to yours :stuck_out_tongue:
Do whatever you want with it :smile:

lib/membrane/element/base/mixin/common_behaviour.ex
```diff
@@ -203,8 +205,11 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
 
         default_val_desc =
           if Keyword.has_key?(v, :default) do
+            default_formatter =
+              v |> Keyword.get(:default_formatter, quote(do: &"`#{inspect(&1)}`"))
```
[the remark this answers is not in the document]
mat-hek: Backticks suggest that there's a term between them, while formatter may return kind of description of what the default value is, which may not be a term. For example '2 ms' returned by `Membrane.Time.pretty_duration/1` shouldn't be considered a term to me.

bblaszkow06:
Backticks are not just for terms but for anything you want to present using
monospace font. In docs I usually put values like this in backticks. In
this case I would also use them to keep the style for different options
cosistent. Anyway, It's a matter of opinion so we probably need a 3rd vote
😉

pon., 17 gru 2018, 10:55: Mateusz Front <notifications@github.com>
napisał(a):

mat-hek:
> Backticks are not just for terms but for anything you want to present using monospace font. In docs I usually put values like this in backticks. In this case I would also use them to keep the style for different options cosistent. Anyway, It's a matter of opinion so we probably need a 3rd vote 😉
I think that it's not really important whether the font is monospaced or not in this case, but it's important to mark terms specially so they could be easily distinguished from usual words

lib/membrane/element/base/mixin/common_behaviour.ex
```diff
@@ -148,31 +150,14 @@ defmodule Membrane.Element.Base.Mixin.CommonBehaviour do
...
-        struct()
-      end
+  @default_types_params %{
+    atom: [spec: quote_expr(atom)],
+    boolean: [spec: quote_expr(boolean)],
+    string: [spec: quote_expr(String.t())],
+    keyword: [spec: quote_expr(keyword)],
+    struct: [spec: quote_expr(struct)],
+    caps: [spec: quote_expr(struct)],
+    time: [spec: quote_expr(Time.t()), inspector: &Time.pretty_duration/1]
```
bblaszkow06: Although this `pretty_duration` is ok, I don't think it is a good way to present the default value. IMHO it should present the call to (`Time.some_unit`) because otherwise, it would be quite misleading

lib/membrane/time.ex
```diff
@@ -50,6 +50,16 @@ defmodule Membrane.Time do
   @type non_neg_t :: non_neg_integer
   @type native_t :: integer
 
+  @units %{
+    days: :d,
```
bblaszkow06: Is there a reason why the units are atoms, not strings?

lib/membrane/time.ex
```diff
@@ -76,9 +87,41 @@ defmodule Membrane.Time do
...
+      iex> import #{inspect(__MODULE__)}
+      iex> 10 |> milliseconds() |> to_code() |> Macro.to_string()
+      quote do 10 |> Membrane.Time.milliseconds() end |> Macro.to_string()
+      iex> 60_000_000 |> microseconds() |> to_code() |> Macro.to_string()
+      quote do 1 |> Membrane.Time.minutes() end |> Macro.to_string()
+      iex> 2 |> nanoseconds() |> to_code() |> Macro.to_string()
+      quote do 2 |> #{inspect(__MODULE__)}.nanoseconds() end |> Macro.to_string()
+
+  """
+  @spec pretty_duration(t) :: Macro.t()
```
bblaszkow06: copy paste went wrong :(

lib/membrane/time.ex
```diff
@@ -50,6 +50,16 @@ defmodule Membrane.Time do
   @type non_neg_t :: non_neg_integer
   @type native_t :: integer
 
+  @units %{
```
bblaszkow06: This name clashes with units function. If the values here were strings I would name it `time_unit_suffixes` or sth like that. You may also consider some other name for `units` function.

lib/membrane/time.ex
mat-hek: You need only one colon instead of a pair of quotes :D

PR #136, opened 2019-01-11
Fix error when unlinking element

bblaszkow06:


lib/membrane/core/element/pad_controller.ex
```diff
@@ -111,6 +111,22 @@ defmodule Membrane.Core.Element.PadController do
...
+    with {:ok, state} <- generate_eos_if_needed(pad_ref, state),
+         {:ok, state} <- handle_pad_removed(pad_ref, state),
+         {:ok, state} <- PadModel.delete_data(pad_ref, state) do
+      {:ok, state}
+    end
+  end
+
+  defp do_handle_unlink(:static, pad_ref, state) do
+    with {:ok, state} <- generate_eos_if_needed(pad_ref, state) do
+      # TODO: some changes in pad data should be considered
```
mat-hek: I think the data can just be deleted because the pad is assumed not to be linked anymore

lib/membrane/core/element/pad_controller.ex
```diff
@@ -74,17 +74,17 @@ defmodule Membrane.Core.Element.PadController do
   end
 
   @doc """
-  Executes `handle_pad_removed` callback for dynamic pads and removes pad data.
+  Handles situation where pad has been unlinked (e.g. when connected element has been removed from pipline)
+
+  Sends an EoS if unlinked pad was an input.
```
mat-hek: It doesn't send anything, rather simulates or something

lib/membrane/core/element/pad_controller.ex
```diff
@@ -74,17 +74,17 @@ defmodule Membrane.Core.Element.PadController do
...
+  Executes `handle_pad_removed` and removes pad data callback if the pad was dynamic.
```
mat-hek: 'removes pad data callback'?

spec/support/element/dynamic_filter.ex
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  @impl true
+  def handle_pad_added(_ref, _ctx, state) do
+    {:ok, state}
+  end
+
+  @impl true
+  def handle_pad_removed(ref, _ctx, state) do
+    send(self(), {:pad_removed, ref})
```
mat-hek: To make this more reusable, pid might be passed through options and stored in the state.

spec/support/element/dynamic_filter.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  def_input_pads input: [caps: :any, availability: :on_request, demand_unit: :buffers]
+
+  @impl true
+  def handle_init(_options) do
+    {:ok, %{pads: MapSet.new()}}
+  end
+
+  @impl true
+  def handle_pad_added(_ref, _ctx, state) do
```
mat-hek: if there is a message sent on pad removal, maybe it should also be sent here

spec/support/element/dynamic_filter.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Membrane.Support.Element.DynamicFilter do
+  @moduledoc """
+  This is a mock filter with dynamic inputs for use in specs.
+
+  Modify with caution as many specs may depend on its shape.
+  """
+
+  use Membrane.Element.Base.Filter
+
+  def_output_pads output: [caps: :any]
```
mat-hek: Output may also be dynamic

bblaszkow06: It's not that easy to create such element if it's supposed to look like it's working, I don't want to waste time on that

PR #138, opened 2019-01-18
Fix error handling in handle_buffer_pull

bblaszkow06:
If PullBuffer.store returned an error, handle_buffer_pull returned
error-reason tuple without state, so the error shown from pipeline was
not clear.

### DIFF
--- a/lib/membrane/core/element/buffer_controller.ex
+++ b/lib/membrane/core/element/buffer_controller.ex
@@ -68,24 +68,17 @@ defmodule Membrane.Core.Element.BufferController do
   defp handle_buffer_pull(pad_ref, buffers, state) do
     PadModel.assert_data!(pad_ref, %{direction: :input}, state)
 
-    with {{:ok, was_empty?}, state} <-
-           PadModel.get_and_update_data!(
-             pad_ref,
-             :buffer,
-             fn pb ->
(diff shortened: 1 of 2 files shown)


lib/membrane/core/element/buffer_controller.ex
```diff
@@ -68,24 +68,17 @@ defmodule Membrane.Core.Element.BufferController do
   defp handle_buffer_pull(pad_ref, buffers, state) do
     PadModel.assert_data!(pad_ref, %{direction: :input}, state)
 
-    with {{:ok, was_empty?}, state} <-
-           PadModel.get_and_update_data!(
-             pad_ref,
-             :buffer,
-             fn pb ->
-               was_empty? = pb |> PullBuffer.empty?()
+    with {:ok, old_pb} <- PadModel.get_data(pad_ref, :buffer, state),
```
mat-hek: This does not have corresponding `else` clause. I think wrapping below clause with `{x, state}` tuple would solve the problem without the need for `else`.

bblaszkow06: 1. In theory, this can return `{:error, reason}` as well. The assert above prevents that from happening but dialyzer might not know that.
2. This is what else is for, wrapping the result here seems like an antipattern to me and would impair readability.

PR #141, opened 2019-02-06
fixes in Membrane.Log

### LINKED ISSUE #140: Membrane.Log.info leaks Bunch import
I had a surprising error - I used `withl` macro without importing/using Bunch, yet there was no error. It appeared after I muted all logs using config and the line before `withl` there was `info("some info")`.

mat-hek:
Closes #140


mat-hek:
@bblaszkow06 check if this fixes your case

PR #146, opened 2019-02-27
Refactor inspired by credo

bblaszkow06:


lib/membrane/log/router.ex
```diff
@@ -38,47 +38,50 @@ defmodule Membrane.Log.Router do
    - :info
    - :warn
   """
-  @spec level_to_val(atom) :: 0 | 1 | 2
+  @spec level_to_val(:debug | :info | :warn) :: 0 | 1 | 2
```
mat-hek: There should be something like `Log.level_t`

PR #148, opened 2019-03-01
Define pads one by one and allow pad options to be specified on definition

bblaszkow06:


lib/membrane/core/element/options_spec_parser.ex
```diff
@@ -0,0 +1,196 @@
...
+  @spec def_options(module(), nil | Keyword.t()) :: Macro.t()
+  def def_options(module, options) do
+    {typedoc, opt_typespecs, escaped_opts} = parse_opts(options)
+    opt_typespec_ast = {:%{}, [], Keyword.put(opt_typespecs, :__struct__, module)}
+    # opt_typespec_ast is equivalent of typespec %__CALLER__.module{key: value, ...}
+    quote do
+      @typedoc """
+      Struct containing options for `#{inspect(__MODULE__)}`
+      """
+      @type t :: unquote(opt_typespec_ast)
```
mat-hek: this can be replaced with `%unquote(module){unquote_splicing(opt_typespecs)}`, and `opt_typespec_ast` can be removed

lib/membrane/core/element/options_spec_parser.ex
```diff
@@ -0,0 +1,196 @@
...
+          |> Keyword.get(
+            :inspector,
+            @default_types_params[opt_definition[:type]][:inspector] || quote(do: &inspect/1)
+          )
+
+        quote do
+          "Default value: `#{unquote(inspector).(unquote(opt_definition)[:default])}`"
+        end
+      else
+        quote_expr("**Required**")
```
mat-hek: change to italics

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+        """
+        @type unquote(Macro.var(pad_opts_type_name, nil)) :: unquote(opt_typespecs)
+      end
+
+    {escaped_opts, type_definiton}
+  end
+
+  def generate_opts_doc(escaped_opts) do
+    escaped_opts
+    |> Enum.map(&generate_opt_doc(&1))
```
mat-hek: `&generate_opt_doc/1`

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+    |> Enum.reduce(fn x, acc ->
```
mat-hek: x -> opt_doc

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+      |> Enum.map(fn {k, v} ->
+        default_val = @default_types_params[v[:type]][:spec] || quote_expr(any)
+
+        {k, v |> Keyword.put_new(:spec, default_val)}
+      end)
+
+    # Actual AST with typespec for the option
+    opt_typespecs =
+      with_default_specs
+      |> Enum.map(fn {k, v} -> {k, v[:spec]} end)
```
mat-hek: Bunch.KVList.map_values

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+
+    # Options without typespec
+    escaped_opts =
+      with_default_specs
+      |> Enum.map(fn {k, v} ->
+        {k, v |> Keyword.put(:spec, Macro.to_string(v[:spec]))}
```
mat-hek: Bunch.KVList.map_values

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+
+      #{unquote(desc) |> String.trim() |> Bunch.Markdown.indent()}
+      """
+    end
+  end
+
+  defp parse_opts(kw) when is_list(kw) do
+    with_default_specs =
+      kw
+      |> Enum.map(fn {k, v} ->
```
mat-hek: Bunch.KVList.map_values

lib/membrane/core/element/pad_controller.ex
```diff
@@ -25,13 +24,14 @@ defmodule Membrane.Core.Element.PadController do
...
+        ) :: State.stateful_try_t(PadModel.pad_info_t())
   def handle_link(pad_ref, direction, pid, other_ref, other_info, props, state) do
     pad_name = pad_ref |> Pad.name_by_ref()
     info = state.pads.info[pad_name]
 
-    with :ok <- validate_pad_being_linked(pad_ref, direction, info, state),
-         :ok <- validate_dir_and_mode(info, other_info) do
+    with :ok <- validate_pad_being_linked!(pad_ref, direction, info, state),
+         :ok <- validate_dir_and_mode!({pad_ref, info}, {other_ref, other_info}),
+         props <- parse_link_props!(props, pad_name, state) do
```
mat-hek: this should reside in the block below

lib/membrane/core/element/pad_controller.ex
```diff
@@ -122,51 +122,104 @@ defmodule Membrane.Core.Element.PadController do
...
           PadModel.pad_info_t(),
           State.t()
-        ) :: Type.try_t()
-  defp validate_pad_being_linked(pad_ref, direction, info, state) do
+        ) :: :ok
+  defp validate_pad_being_linked!(pad_ref, direction, info, state) do
     cond do
       :ok == PadModel.assert_instance(state, pad_ref) ->
-        {:error, :already_linked}
+        raise LinkError, "Pad #{inspect(pad_ref)} has already been linked!"
```
mat-hek: Unnecessary exclamation mark

lib/membrane/core/element/pad_controller.ex
```diff
@@ -122,51 +122,104 @@ defmodule Membrane.Core.Element.PadController do
...
+    buffer_props = parse_buffer_props!(pad_name, props[:buffer])
+    [pad: pad_props, buffer: buffer_props]
+  end
+
+  defp parse_pad_props!(_pad_name, nil, nil) do
+    {:ok, nil}
+  end
+
+  defp parse_pad_props!(pad_name, nil, _props) do
+    raise LinkError, "Pad #{inspect(pad_name)} does not define any options!"
```
mat-hek: Unnecessary exclamation mark

lib/membrane/core/element/pad_specs.ex
```diff
@@ -0,0 +1,236 @@
...
+          unquote(acc)
+          unquote(pad_def)
+        end
+      end
+    )
+  end
+
+  @spec def_pad_docs(Pad.direction_t()) :: String.t()
+  def def_pad_docs(direction) do
+    dir_str = direction |> to_string()
```
mat-hek: to remove

lib/membrane/core/element/pad_specs.ex
```diff
@@ -0,0 +1,236 @@
...
+    quote do
+      @doc """
+      Returns pads descriptions for `#{inspect(__MODULE__)}`
+      """
+      @spec membrane_pads() :: [{unquote(Pad).name_t(), unquote(Pad).description_t()}]
+      def membrane_pads() do
+        @membrane_pad
+      end
+
+      @moduledoc """
```
mat-hek: you can check here if `moduledoc` is defined and not false

lib/membrane/core/element/pad_specs.ex
```diff
@@ -0,0 +1,236 @@
...
+                options: [default: nil]
+              ) do
+      {:ok, {name, Map.put(config, :direction, direction)}}
+    else
+      spec: spec -> {:error, {:invalid_pad_spec, spec}}
+      config: {:error, reason} -> {:error, {reason, pad: name}}
+    end
+  end
+
+  # Generates docs describing pads, based on pads specification.
```
mat-hek: such comments should be docs

lib/membrane/core/input_buffer.ex
```diff
@@ -64,6 +64,21 @@ defmodule Membrane.Core.InputBuffer do
           | {:fail_size, pos_integer()}
         ]
 
+  @spec parse_props(keyword()) :: {:error, reason :: any()} | {:ok, props_t()}
+  def parse_props(input) do
+    with input <- input |> List.wrap(),
```
mat-hek: this should be placed either before with or piped just before the call to Bunch.Config.Parse

lib/membrane/element.ex
```diff
@@ -167,7 +165,7 @@ defmodule Membrane.Element do
   """
   @spec link(link_spec :: %Link{}) :: :ok | {:error, any}
   def link(%Link{from: %Endpoint{pid: pid}, to: %Endpoint{pid: pid}}) when is_pid(pid) do
-    {:error, :loop}
+    raise LinkError, "Cannot link element with itself!"
```
mat-hek: Unnecessary exclamation mark

lib/membrane/core/element/options_specs.ex
```diff
@@ -0,0 +1,166 @@
...
+    boolean: [spec: quote_expr(boolean)],
+    string: [spec: quote_expr(String.t())],
+    keyword: [spec: quote_expr(keyword)],
+    struct: [spec: quote_expr(struct)],
+    caps: [spec: quote_expr(struct)],
+    time: [spec: quote_expr(Time.t()), inspector: &Time.to_code_str/1]
+  }
+
+  @spec options_doc() :: String.t()
+  def options_doc do
```
mat-hek: I would place this doc directly in CommonBehaviour

spec/membrane/element_spec.exs
```diff
@@ -23,19 +23,26 @@ defmodule Membrane.ElementSpec do
       to: %Endpoint{element: :b, pad_name: to_pad(), pad_ref: to_pad(), pid: to_pid()}
     }
 
+  defmacrop catch_exit(call) do
```
mat-hek: Maybe separate this to some helper?

lib/membrane/core/element/options_specs.ex
bblaszkow06: And I would not :stuck_out_tongue:
1. This way everything connected to options is in one place - if we change anything about how options are parsed (i.e. modify the parse_opts from this module), the doc that is supposed to change will be in the same module.
2. This function is only used here, but it shouldn't - docs for Pad require improvement and this function should be used there

spec/membrane/element_spec.exs
bblaszkow06: 1. It is ESpec-specific (ExUnit provides that)
2. It is quite test-specific - `capture_log` is not something you want to add in general helper and pattern in catch is crafted only for a specific case - failing call to GenServer
So until there's another use case - I don't think it's necessary

lib/membrane/core/element/options_specs.ex
mat-hek: agreed

bblaszkow06: I've made an issue for this #153

spec/membrane/element_spec.exs
mat-hek: the helper could reside somewhere in the support module. it would be nice to see it's already done once fixing tests after the next change like this

PR #149, opened 2019-03-12
Add proxy funs for Membrane.Pipeline to pipelines

bblaszkow06:


lib/membrane/pipeline.ex
```diff
@@ -183,6 +183,9 @@ defmodule Membrane.Pipeline do
     end
   end
 
+  @doc """
+  Changes pipeline's playback state to `:stopped` and terminates its process
+  """
   @spec stop_and_terminate(pipeline :: pid) :: :ok
   def stop_and_terminate(pipeline) do
```
mat-hek: Add also a delegate to this in `Pipeline.__using__`

PR #158, opened 2019-04-15
Allow to pass dynamic pad id when linking

bblaszkow06:


mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Membrane.Mixfile do
   use Mix.Project
 
-  @version "0.3.0"
+  @version "0.3.1"
```
mat-hek: I would rather bump version directly on master or in a separate PR

bblaszkow06: I'll leave it as-is and keep that in mind in future :wink:

bblaszkow06:
@Hajto This was merged, you might be interested in that.

PR #160, opened 2019-04-17
Synchronization and clock utilities

mat-hek:
Usage examples:
- SDL: https://github.com/membraneframework/membrane-element-sdl/blob/feature/sync/lib/membrane_element_sdl/sink.ex
- PortAudio: https://github.com/membraneframework/membrane-element-portaudio/blob/feature/sync/lib/membrane_element_portaudio/sink.ex


lib/membrane/sync.ex
```diff
@@ -0,0 +1,57 @@
...
+    GenServer.call(sync, :sync)
+  end
+
+  @impl true
+  def init(_opts) do
+    {:ok, %{state: :reg, registered: MapSet.new(), synced: MapSet.new()}}
+  end
+
+  @impl true
+  def handle_call(:reg, {pid, _ref}, %{state: :reg} = state) do
```
bblaszkow06: You should also handle a call in sync state with some error

lib/membrane/sync.ex
```diff
@@ -0,0 +1,57 @@
...
+  def start_link(options \\ []) do
+    GenServer.start_link(__MODULE__, %{}, options)
+  end
+
+  def register(sync) do
+    GenServer.call(sync, :reg)
+  end
+
+  def sync(sync) do
+    GenServer.call(sync, :sync)
```
bblaszkow06: I wonder how you plan to use it, but I would expect `sync` call to block until actually synced, so after GenServer call there could be `receive` block waiting for sync message.

mat-hek:
Example of usage: https://github.com/membraneframework/membrane-element-sdl/blob/feature/sync/lib/membrane_element_sdl/sink.ex

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
...
+
+  def start_link!(options \\ [], gen_server_options \\ []) do
+    {:ok, clock} = start_link(options, gen_server_options)
+    clock
+  end
+
+  @impl GenServer
+  def init(options) do
+    state =
+      %{ratio: 1, subscribees: MapSet.new()}
```
bblaszkow06: subscribers?

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
+defmodule Membrane.Clock do
+  use GenServer
```
bblaszkow06: Docs are really needed, even for preliminary review

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
...
+          _ ->
+            %{init_time: nil, recv_time: 0, till_next: nil, proxy: false}
+        end
+      )
+
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_info({:membrane_clock_update, till_next}, %{proxy: false} = state) do
```
bblaszkow06: Why these messages don't have `Message.new` helpers?

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
...
+    {:ok, clock} = start_link(options, gen_server_options)
+    clock
+  end
+
+  @impl GenServer
+  def init(options) do
+    state =
+      %{ratio: 1, subscribees: MapSet.new()}
+      |> Map.merge(
+        case {options[:proxy], options[:proxy_to]} do
```
bblaszkow06: Something cannot be `proxy to`. Choose on of these words: https://ell.stackexchange.com/questions/158129/the-difference-between-proxy-of-vs-proxy-for-proxy-by

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  defp handle_clock_update(till_next, %{init_time: nil} = state) do
+    %{state | init_time: Time.monotonic_time(), till_next: till_next}
+  end
+
+  defp handle_clock_update(new_till_next, state) do
+    use Ratio
+    %{till_next: {nom, denom}, recv_time: recv_time} = state
+    recv_time = recv_time + (Time.milliseconds(nom) <|> denom)
```
bblaszkow06: This operator is terrible. I remember I've already asked what it does, but I should not ask at all. Please use a normal call instead of magic

lib/membrane/clock.ex
```diff
@@ -0,0 +1,107 @@
...
+  def handle_info(Message.new(:clock_subscribe, pid), state) do
+    send(pid, {:membrane_clock_ratio, self(), state.ratio})
+    Process.monitor(pid)
+    state = state |> Map.update!(:subscribees, &MapSet.put(&1, pid))
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_info(Message.new(:clock_unsubscribe, pid), state) do
+    Process.demonitor(pid)
```
bblaszkow06: Unfortunately, `demonitor/1` accepts monitor ref, not pid

lib/membrane/clock.ex
mat-hek: Because they can be sent by non-core modules

lib/membrane/core/element/timer_controller.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  def stop_timer(id, state) do
+    {timer, state} = state |> Bunch.Access.pop_in([:timers, id])
+
+    if timer |> is_nil do
+      {{:error, {:unknown_timer, id}}, state}
+    else
+      :ok = timer |> Timer.stop()
+
+      unless state.timers |> Bunch.KVList.any_value?(&(&1.clock == timer.clock)) do
```
bblaszkow06: Maybe this check could be moved to `Clock.unsubscribe`? (That would require to keep counter inside Clock incremented on `Clock.subscribe`)

lib/membrane/pipeline.ex
```diff
@@ -318,12 +336,86 @@ defmodule Membrane.Pipeline do
     end
   end
 
-  @spec add_children([parsed_child_t], State.t()) :: Type.stateful_try_t(State.t())
-  defp add_children(children, state) do
+  defp setup_clock(module, pid) do
+    if Bunch.Module.loaded_and_function_exported?(module, :clock, 0) do
```
bblaszkow06: I haven't noticed any docs on how to define clocks :thinking:

lib/membrane/pipeline.ex
```diff
@@ -318,12 +336,86 @@ defmodule Membrane.Pipeline do
...
+    end
+  end
+
+  defp choose_clock(state) do
+    choose_clock([], nil, state)
+  end
+
+  defp choose_clock(children, provider, state) do
+    cond do
+      provider -> get_clock_from_provider(children, provider)
```
bblaszkow06: Explicit `!= nil` could be more readable

lib/membrane/pipeline.ex
```diff
@@ -318,12 +336,86 @@ defmodule Membrane.Pipeline do
...
+
+  @spec add_children([parsed_child_t], State.t()) :: Type.stateful_try_t(State.t())
+  defp add_children(children, state) do
+    {:ok, state} =
+      children
+      |> Bunch.Enum.try_reduce(state, fn {name, data}, state ->
+        state |> State.add_child(name, data)
+      end)
+
+    state
```
bblaszkow06: Returned value does not match spec

lib/membrane/pipeline.ex
```diff
@@ -365,22 +457,49 @@ defmodule Membrane.Pipeline do
...
-    end)
+  defp set_children_watcher(children_data) do
+    :ok = children_data |> Bunch.Enum.try_each(&Message.call(&1.pid, :set_watcher, self()))
+  end
 
-    :ok
+  defp set_children_stream_sync(children, links, stream_sync) do
+    children = children |> Map.new()
+
+    unless stream_sync do
```
bblaszkow06: `unless` with `else` is discouraged by credo, you can swap cases and use `if`

lib/membrane/pipeline.ex
```diff
@@ -365,22 +457,49 @@ defmodule Membrane.Pipeline do
...
+      [
+        children
+        |> Map.keys()
+        |> Enum.reject(fn cn -> links |> Enum.any?(&(&1.from.element == cn)) end)
```
bblaszkow06: Do we really want to enforce sync sinks by default? Shouldn't that be optional?

lib/membrane/pipeline.ex
```diff
@@ -512,8 +631,38 @@ defmodule Membrane.Pipeline do
   end
 
   @impl CallbackHandler
-  def handle_action({:forward, {elementname, message}}, _cb, _params, state) do
-    with {:ok, pid} <- state |> State.get_child_pid(elementname) do
+  def handle_actions(%Spec{} = spec, :handle_init, params, state) do
+    super([spec: spec], :handle_init, params, state)
```
bblaszkow06: Let's deprecate spec without proper action

lib/membrane/clock.ex
```diff
@@ -0,0 +1,154 @@
...
+      else
+        send_and_update_ratio(1, state)
+      end
+
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_info(Message.new(:clock_subscribe, pid), state) do
+    send(pid, {:membrane_clock_ratio, self(), state.ratio})
```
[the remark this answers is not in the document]
mat-hek:
```
iex(1)> Ratio.new(1, 1)
1
```

test/membrane/clock_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Membrane.ClockTest do
+  @module Membrane.Clock
+
+  use ExUnit.Case
+
+  test "should send ratio once a new subscriber connects" do
+    clock = @module.start_link!()
```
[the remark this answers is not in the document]
mat-hek: This is a part of the last, WIP commit :D

spec/membrane/pipeline_spec.exs
```diff
@@ -144,11 +132,6 @@ defmodule Membrane.PipelineSpec do
       expect(state.playback.state) |> to(eq :stopped)
     end
 
-    it "should send a message to initialize children asynchronously" do
```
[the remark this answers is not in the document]
mat-hek: no

spec/membrane/pipeline_spec.exs
```diff
@@ -173,46 +156,6 @@ defmodule Membrane.PipelineSpec do
...
-        |> to(eq spec().children |> Keyword.keys() |> Enum.sort())
-      end
-
-      it "should return new pipeline state containing state" do
-        {:noreply, state} = described_module().handle_info(message(), state())
-        expect(state.internal_state) |> to(eq internal_state())
-      end
-
-      it "should call element's handle_init" do
-        described_module().handle_info(message(), state())
```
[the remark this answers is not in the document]
mat-hek: no

lib/membrane/sync.ex
```diff
@@ -0,0 +1,216 @@
+defmodule Membrane.Sync do
+  @moduledoc """
+  Sync allows to synchronize multiple processes, so that they performed their jobs
+  at the same time.
+
+  The flow of usage goes as follows:
+  - Processes register themselves in Sync, using `register/2`.
```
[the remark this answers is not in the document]
mat-hek: This is explained in docs for `start_link/2`

lib/membrane/element/base.ex
```diff
@@ -201,6 +232,45 @@ defmodule Membrane.Element.Base do
...
 
+  @doc """
+  Defines that element exports a clock to pipeline.
+
+  Exporting clock allows pipeline to choose it as the pipeline clock, enableing other
+  elements to synchronize with it. Element's clock is accessible via `clock` field,
+  while pipeline's one - via `pipeline_clock` field in callback contexts. Both of
+  them can be used for starting timers.
+  """
+  defmacro def_clock do
```
bblaszkow06: There should be a way to add docs for a clock

lib/membrane/element/base.ex
```diff
@@ -201,6 +232,45 @@ defmodule Membrane.Element.Base do
...
+    quote do
+      @membrane_clock_moduledoc """
```
bblaszkow06: We should probably prevent defining more than one clock

lib/membrane/element/base.ex
bblaszkow06: Plus if this becomes `def_clock/1` it should be added to `.formatter.exs`

lib/membrane/clock.ex
```diff
@@ -0,0 +1,249 @@
+defmodule Membrane.Clock do
```
bblaszkow06: Edit `mix.exs` and add this module to proper category

lib/membrane/element/base.ex
```diff
@@ -201,6 +232,45 @@ defmodule Membrane.Element.Base do
...
+      def membrane_clock?, do: true
+    end
+  end
+
+  @doc false
+  def update_moduledoc() do
+    quote do
+      if @moduledoc != false do
+        @moduledoc [
+                     :membrane_options_moduledoc,
```
bblaszkow06: This implementation removes existing moduledoc!

lib/membrane/core/element/state.ex
```diff
@@ -54,19 +60,24 @@ defmodule Membrane.Core.Element.State do
...
+      type: apply(options.module, :membrane_element_type, []),
+      name: options.name,
       internal_state: nil,
       pads: nil,
       watcher: nil,
       controlling_pid: nil,
       playback: %Playback{},
       playback_buffer: PlaybackBuffer.new(),
       delayed_demands: %{},
+      timers: %{},
```
kanes115: Maybe putting all sync-related stuff under one key is a good idea?

lib/membrane/core/element/timer_controller.ex
```diff
@@ -0,0 +1,59 @@
...
+      :ok = timer |> Timer.stop()
+      timer.clock |> Clock.unsubscribe()
+      {:ok, state}
+    end
+  end
+
+  def handle_tick(timer_id, %State{} = state) do
+    context = &CallbackContext.Tick.from_state/1
+
+    with true <- state.timers |> Map.has_key?(timer_id) or {:ok, state},
```
kanes115: Wouldn't it be more clear if you remove `or {:ok, state}` and add else clause?

lib/membrane/element/base.ex
```diff
@@ -183,13 +191,36 @@ defmodule Membrane.Element.Base do
...
+            ) :: callback_return_t
+
   @doc """
   Callback invoked when element is shutting down just before process is exiting.
   Internally called in `c:GenServer.terminate/2` callback.
   """
   @callback handle_shutdown(reason, state :: Element.state_t()) :: :ok
             when reason: :normal | :shutdown | {:shutdown, any}
 
+  @optional_callbacks membrane_clock?: 0,
```
kanes115: Why do we have both default implementations and optional_callbacks?

lib/membrane/element/base.ex
```diff
@@ -201,6 +232,45 @@ defmodule Membrane.Element.Base do
...
+  @doc false
+  def update_moduledoc() do
+    quote do
+      if @moduledoc != false do
+        @moduledoc [
+                     :membrane_options_moduledoc,
+                     :membrane_clock_moduledoc,
+                     :membrane_pads_moduledoc
+                   ]
+                   |> Enum.map(&Module.get_attribute(__MODULE__, &1))
```
kanes115: I would consider wrapping the next 3 calls into a function.

lib/membrane/pipeline.ex
```diff
@@ -377,12 +430,70 @@ defmodule Membrane.Pipeline do
...
   end
 
-  @spec add_children([parsed_child_t], State.t()) :: Type.stateful_try_t(State.t())
-  defp add_children(children, state) do
+  defp choose_clock(state) do
+    choose_clock([], nil, state)
+  end
+
+  defp choose_clock(children, provider, state) do
+    cond do
```
kanes115: I would rather at least have the value returned by this `cond` be bound to a var so that the result is named and the reader knows what it is before going on with reading to `case` statement. `nil` does not say a lot.

lib/membrane/pipeline.ex
```diff
@@ -583,13 +701,49 @@ defmodule Membrane.Pipeline do
 
   @impl GenServer
   def terminate(reason, state) do
-    CallbackHandler.exec_and_handle_callback(:handle_shutdown, __MODULE__, [reason], state)
+    :ok = state.module.handle_shutdown(reason, state.internal_state)
     :ok
```
kanes115: No need for this line

lib/membrane/pipeline.ex
```diff
@@ -583,13 +701,49 @@ defmodule Membrane.Pipeline do
...
+    Returning bare spec from `handle_init` is deprecated.
+    Return `{{:ok, spec: spec}, state}` instead.
+    Found in `#{inspect(state.module)}.handle_init/1`.
+    """)
+
+    super([spec: spec], :handle_init, params, state)
+  end
+
+  @impl CallbackHandler
+  def handle_actions(actions, callback, params, state) do
```
kanes115: Do we need this function at all here if the only thing we do is calling super?

lib/membrane/sync.ex
```diff
@@ -1,40 +1,46 @@
...
 
+  The main purpose for Sync is to synchronize multiple streams within a pipeline.
   The flow of usage goes as follows:
-  - Processes register themselves in Sync, using `register/2`.
-  - When a process is ready to synchronize, it invokes `ready/1`. When a process
-  is ready, it is assumed it is going to invoke `sync/2` approximately at the same
-  time as all other ready processes.
-  - If a process becomes no longer ready, it should invoke `unready/1`.
+  - A Sync process is started.
+  - Processes register themselves (or are being registered) in the Sync, using
```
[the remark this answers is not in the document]
kanes115: I also don't think there is a need for "register themselves" and "are registered" is sufficient.

kanes115:
Out of curiosity: have we considered maybe measuring function call times? Draw some flame graphs to establish what accuracy we can achieve in time synchronisation? If a decent accuracy is achievable on membrane/BEAM?

lib/membrane/clock.ex
```diff
@@ -0,0 +1,249 @@
...
+  def init(options) do
+    state =
+      %{
+        ratio: 1,
+        subscribers: %{},
+        time_provider:
+          options |> Keyword.get(:time_provider, fn -> System.monotonic_time(:millisecond) end)
+      }
+      |> Map.merge(
+        case {options[:proxy], options[:proxy_for]} do
```
kanes115: I think this `case` is too long and complicated and thus should be moved to a seperate function.

lib/membrane/clock.ex
```diff
@@ -0,0 +1,249 @@
...
+  end
+
+  defp handle_clock_update(till_next, state) do
+    use Ratio
+
+    if till_next < 0 do
+      raise "Clock update time cannot be negative, received: #{inspect(till_next)}"
+    end
+
+    till_next = till_next * Time.millisecond(1)
```
kanes115: Would it make sense to put such an enigmatic instruction inside a function which name would tell the reader the reasoning behind it?

lib/membrane/clock.ex
```diff
@@ -0,0 +1,249 @@
...
+  end
+
+  defp broadcast_and_update_ratio(ratio, state) do
+    state.subscribers |> Bunch.KVList.each_key(&send_ratio(&1, ratio))
+    %{state | ratio: ratio}
+  end
+
+  defp send_ratio(pid, ratio) do
+    send(pid, {:membrane_clock_ratio, self(), ratio})
+    :ok
```
kanes115: Unused return value

test/membrane/clock_test.exs
```diff
@@ -0,0 +1,147 @@
+defmodule Membrane.ClockTest do
+  @module Membrane.Clock
+
+  use ExUnit.Case
+
+  test "should calculate proper ratio and send it to subscribers on each (but the first) update" do
```
kanes115: Magic numbers (?)

lib/membrane/clock.ex
```diff
@@ -0,0 +1,249 @@
...
+    ~> {:noreply, &1}
+  end
+
+  @impl GenServer
+  def handle_info({:membrane_clock_update, till_next}, %{proxy: false} = state) do
+    {:noreply, handle_clock_update(till_next, state)}
+  end
+
+  @impl GenServer
+  def handle_info({:membrane_clock_ratio, pid, ratio}, %{proxy: true, proxy_for: pid} = state) do
```
kanes115: For the two next caluses: why don't we use `gen_server:handle_cast/cast`? Why those messages are not wrapped in an API function?

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
+defmodule Membrane.SyncTest do
+  @module Membrane.Sync
+
+  use ExUnit.Case
+  use Bunch
+
+  describe "should sync processes" do
+    defp gen_times do
```
kanes115: I think this function:
* is too long itself
* its name is not descriptive enough (what times do we get?). It's doing more than just getting them.
* has too long anonymous functions that are nested

lib/membrane/sync.ex
```diff
@@ -0,0 +1,195 @@
...
+  - :empty_exit? - if true, Sync automatically exits when all the registered
+    processes exit; defaults to false
+
+  """
+  @spec start_link([empty_exit?: boolean], GenServer.options()) :: GenServer.on_start()
+  def start_link(options \\ [], gen_server_options \\ []) do
+    GenServer.start_link(__MODULE__, options, gen_server_options)
+  end
+
+  def start_link!(options \\ [], gen_server_options \\ []) do
```
kanes115: Do we really need this "throwing" version of start link? I think this is not a common practice to do so, usually that is the user of the GenServer module that takes care of checking the starting result.

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+      sync = @module.start_link!()
+
+      tasks =
+        1..10
+        |> Enum.map(fn i ->
+          Task.async(fn ->
+            :ok = receive do: (:cont -> :ok)
```
kanes115: I would extract this `:cont` atom to a module attribute that would tell the reader what message this is.

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+            Process.sleep(10 * i)
```
kanes115: Magic number. Also, what are we waiting here for? Could we put that information in code?

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+        |> Enum.unzip()
+
+      {request_times, sync_times}
+    end
+
+    test "synchronization should happen shortly after the last process wants to synchronize" do
+      {request_times, sync_times} = gen_times()
+      now = System.monotonic_time(:millisecond)
+      last_request = request_times |> Enum.min_by(&(now - &1))
+      sync_times |> Enum.each(&assert_in_delta(&1, last_request, 10))
```
kanes115: 10, 5 <- magic numbers. I think this is a good idea to put those into module attributes. This way if the tests are run i.e. on a weaker hardware you can adjust those values.

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+    end
+
+    test "synchronization should happen approximately at the same time in each process" do
+      {_request_times, sync_times} = gen_times()
+      sync_times |> Enum.min_max() ~> ({min, max} -> assert_in_delta(min, max, 5))
+    end
+  end
+
+  test "should sync only if active" do
```
kanes115: I would add something about immediately returning in the test description

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+        send(t1.pid, :cont)
+        :ok = sync |> @module.sync()
+        :ok
+      end)
+
+    :ok = Task.await(t1)
+    :ok = Task.await(t2)
+  end
+
+  test "should exit once all syncees exit iff :empty_exit? flag is present" do
```
kanes115: typo `iff`

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+    s1 = @module.start_link!()
+    s2 = @module.start_link!(empty_exit?: true)
+
+    [s1, s2]
+    |> Enum.each(fn sync ->
```
kanes115: Too much nesting, mapping list literal in `Enum.each` :D

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+        task
+      end)
+      |> Enum.map(fn task ->
+        send(task.pid, :cont)
+        task
+      end)
+      |> Enum.map(&Task.await/1)
+    end)
+
+    Process.sleep(20)
```
kanes115: Magic number again. Also put the information about why we wait in code (wrap inside a function or put a comment)

test/membrane/sync_test.exs
```diff
@@ -0,0 +1,224 @@
...
+    assert Process.alive?(s1)
+    refute Process.alive?(s2)
+  end
+
+  test "#{inspect(@module)}.no_sync/0 should sync at once" do
```
kanes115: "sync at once" IMO is a little bit confusing. There is no synchronisation (funny enough that is what the name says) so let's rephrase it to say "call to sync returns immediately with no synchronisation"?

test/membrane/sync_test.exs
bblaszkow06: That's not a typo, although I thought so some time ago: https://github.com/erlang/otp/pull/2037

test/membrane/clock_test.exs
```diff
@@ -3,11 +3,145 @@ defmodule Membrane.ClockTest do
...
+    assert_receive {:membrane_clock_ratio, ^clock, _ratio}
+    send(clock, {:DOWN, nil, :process, self(), nil})
+    send(clock, {:membrane_clock_update, 42})
+    refute_receive {:membrane_clock_ratio, ^clock, _ratio}
   end
 
-  test "should calculate proper ratio and send it to all subscribers on each update" do
+  test "should ignore unsubscribe from a non-subscribed process" do
     clock = @module.start_link!()
+    @module.unsubscribe(clock)
```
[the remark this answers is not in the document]
kanes115: My question is: subscribe/unsubscribe is a cast and not a call in the first place?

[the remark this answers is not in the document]
kanes115: You don't need an acknolegdment that you successfully subscribed?

lib/membrane/sync.ex
```diff
@@ -47,103 +53,97 @@ defmodule Membrane.Sync do
...
+  def handle_call({:sync, options}, {pid, _ref} = from, %{active?: true} = state) do
+    latency = options |> Keyword.get(:latency, 0)
+
+    case state.processes[pid] do
+      nil ->
+        {:reply, {:error, :not_found}, state}
 
-      {:ok, state} ->
-        {:reply, :ok, state}
+      %{status: :registered} = syncee ->
```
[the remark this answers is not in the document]
kanes115: ?

lib/membrane/sync.ex
```diff
@@ -47,103 +53,97 @@ defmodule Membrane.Sync do
...
+        state =
+          state
+          |> put_in([:processes, pid], %{syncee | status: :sync, latency: latency, reply_to: from})
```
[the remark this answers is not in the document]
kanes115: ?

lib/membrane/sync.ex
```diff
@@ -154,63 +154,42 @@ defmodule Membrane.Sync do
...
-        {:ok, state |> put_in([:syncees, ref, :level], new_level)}
-
-      true ->
-        {{:error, invalid_level: syncee.level.name}, state}
-    end
-  end
-
   defp check_and_handle_sync(state) do
-    unless all_syncees_level?(state.syncees, [:sync, :registered]) do
+    if state.active? and state.processes |> Bunch.KVList.any_value?(&(&1.status != :sync)) do
```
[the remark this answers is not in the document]
kanes115: Not really, @bblaszkow06, test "should finish syncing currently synced processes upon deactivation" tells you why. If you deactivate the sync we want to release all the locks in all subscribed processes. I have changed the condition a little for it to be more clear though.

lib/membrane/sync.ex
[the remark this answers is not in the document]
kanes115: I assume we don't like this name and want to get rid of it, but I need confirmation.

test/membrane/clock_test.exs
```diff
@@ -3,11 +3,145 @@ defmodule Membrane.ClockTest do
...
+
+  test "should send proper ratio when default time provider is used" do
+    clock = @module.start_link!()
+    @module.subscribe(clock)
+    assert_receive {:membrane_clock_ratio, ^clock, 1}
+    send(clock, {:membrane_clock_update, 100})
+    Process.send_after(clock, {:membrane_clock_update, 100}, 50)
+    Process.send_after(clock, {:membrane_clock_update, 1234}, 100)
+    assert_receive {:membrane_clock_ratio, ^clock, ratio}
+    assert_in_delta Ratio.to_float(ratio), 2, 0.3
```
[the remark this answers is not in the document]
kanes115: Actually on my computer it sometimes does not pass with this range.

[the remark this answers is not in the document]
kanes115: I think the 0.3 delta is OK. It's like 15% error. We are dealing with small times (50ms).

lib/membrane/sync.ex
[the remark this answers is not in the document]
bblaszkow06: Yep, the registered processes used to be called  `syncees` before the previous review

test/membrane/sync_test.exs
bblaszkow06: I would hurt @mat-hek, he's a die-hard fan of using as little characters as possible. Let's make him suffer

lib/membrane/clock.ex
```diff
@@ -0,0 +1,253 @@
...
+    {:noreply, broadcast_and_update_ratio(ratio, state)}
+  end
+
+  @impl GenServer
+  # When ratio from previously proxied clock comes in after unsubscribing
```
kanes115: Again: why isn't unsubscribing synchronous? (done with handle_call)?

test/membrane/clock_test.exs
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  test "all subscribers should receive the same ratio after each update" do
+    clock = @module.start_link!(time_provider: fn -> receive do: (time: t -> t) end)
+
+    fn ->
+      Task.start_link(fn ->
+        @module.subscribe(clock)
+        assert_receive {:membrane_clock_ratio, ^clock, 1}
+        assert_receive {:membrane_clock_ratio, ^clock, 2}
```
kanes115: Actually, why are we expecting integer and not float? Ration doesn't have to be a whole number, does it? I think we should always return float.

lib/membrane/element/base.ex
kanes115: What do you mean?

test/membrane/clock_test.exs
bblaszkow06: No, just the opposite as floats have limited precision! The ratio here is calculated using (Ratio)[https://github.com/Qqwy/elixir-rational] library that keeps rational numbers using integer numerator and denominator. It also simplifies expressions, so `Ratio.new(6, 3) == 2`

kanes115: Ah, sure

lib/membrane/element/base.ex
```diff
@@ -270,6 +270,12 @@ defmodule Membrane.Element.Base do
                      :membrane_pads_moduledoc
                    ]
                    |> Enum.map(&Module.get_attribute(__MODULE__, &1))
+                   |> Enum.concat([Module.get_attribute(__MODULE__, :moduledoc)])
```
bblaszkow06: This isn't the best way to to this (to say the least). :stuck_out_tongue:
1. You swapped the order of docs, original moduledoc will now appear at the bottom
2. Each call to update_docs duplicates the moduledoc entries

To see what's wrong generate docs and check out `Membrane.Testing.Sink/Source`

kanes115: @bblaszkow06, see f3ac00915d6348af938d9be4cc115ada0e256c51 and resolve this, if possible, please

lib/membrane/element/base.ex
kanes115: Not relevant anymore

lib/membrane/element/base.ex
bblaszkow06: Now you can only write `def_clock` and module dock will contain "this element provides a clock" (or similar). You might want to add docs saying "This element provides a clock from a soundcard" or something like this. To do this, `def_clock` would have to accept params.

lib/membrane/core/element/options_specs.ex
```diff
@@ -45,14 +45,17 @@ defmodule Membrane.Core.Element.OptionsSpecs do
       """
       @type t :: %unquote(module){unquote_splicing(opt_typespecs)}
 
-      @membrane_options_moduledoc """
-      ## Element options
+      with {_, previous_doc} <- Module.get_attribute(__MODULE__, :moduledoc) do
```
bblaszkow06: This is still broken. If an element's creator sets `@moduledoc false` false is rendered as a string and all docs are present

lib/membrane/element/base.ex
```diff
@@ -243,13 +243,18 @@ defmodule Membrane.Element.Base do
   defmacro def_clock do
     quote do
       if !Module.get_attribute(__MODULE__, :has_clock) do
-        @membrane_clock_moduledoc """
-        ## Clock
+        @has_clock true
```
bblaszkow06: This might be nitpicking, but for such generated module attribute I would use much longer name to reduce the risk of naming conflict. Please use some prefix like `membrane_element_`

lib/membrane/core/element/options_specs.ex
```diff
@@ -45,14 +45,17 @@ defmodule Membrane.Core.Element.OptionsSpecs do
...
+        @moduledoc """
+        #{previous_doc}
 
-      Passed via struct `t:#{inspect(__MODULE__)}.t/0`
+        ## Element options
```
bblaszkow06: It's not a big thing, but it would be nice to keep the generated docs consistent. Right now the order of appearance of docs for options and clock depends on the order of macro calls. This can be handled by generating `moduledoc` only once - in `before_compile` callback

test/membrane/clock_test.exs
kanes115: I am now going with not naming those numbers, but represent the asserted value as additions of partial clock updates

lib/membrane/clock.ex
kanes115: It might be easier to send normal messages from NIFs which is a common use case for clocks.

lib/membrane/core/element/options_specs.ex
kanes115: We will leave this as is until [this bug](https://github.com/elixir-lang/elixir/issues/9334) is fixed.

lib/membrane/sync.ex
kanes115: Unrelevant, right, @bblaszkow06?

lib/membrane/core/element/options_specs.ex
kanes115: Appears that the order of `before_compile` execution stays as is.

If we want to preserve the order between docs for different element modules we have to make on of the tradeoffs:
* We can lose the order of pads `pads_specs |> Enum.sort_by(fn {_, config} -> config[:direction] end)` (because we would need to generate the docs "on the fly" of compiling `pad_specs.ex` (one pad doc at a time).
* We can generate pads docs in `Base` module which makes the code ugly (?)

lib/membrane/core/element/pad_specs.ex
```diff
@@ -169,17 +162,22 @@ defmodule Membrane.Core.Element.PadsSpecs do
...
+        #{x}
         """
-      end
-    end)
+      end)
+
+    """
+    ## Pads
+
+    #{pads_docs}
```
bblaszkow06: By not using `quote/unquote` at any point in docs generation you prevent using module attributes, aliased modules etc. for anything that will be used in docs (like default value of some option, caps)

lib/membrane/core/element/state.ex
bblaszkow06: bump

lib/membrane/core/element/timer_controller.ex
bblaszkow06: bump

lib/membrane/element/action.ex
```diff
@@ -157,6 +157,28 @@ defmodule Membrane.Element.Action do
...
+          {:start_timer,
+           {id :: any, interval :: Time.t(), clock :: Clock.t()}
+           | {id :: any, interval :: Time.t()}}
+
+  @typedoc """
+  Stops a timer started with `t:timer_t` action.
+
+  This action is atomic: stopping timer guarantees that no ticks will arrive from it.
+  """
+  @type stop_timer_t :: {:timer, id :: any}
```
bblaszkow06: This does not seem correct, shouldn't there be `:stop_timer`?

lib/membrane/element/base.ex
bblaszkow06: If you don't use def_clock macro, there will be no `membrane_clock?` callback, for others I'm guessing it is a matter of docs generation - if callback is marked as optional the user knows he doesn't need to implement it, but I'm not sure if that's a proper solution.

lib/membrane/element/base.ex
```diff
@@ -201,15 +232,79 @@ defmodule Membrane.Element.Base do
...
 
+  @doc """
+  Defines that element exports a clock to pipeline.
+
+  Exporting clock allows pipeline to choose it as the pipeline clock, enabling other
+  elements to synchronize with it. Element's clock is accessible via `clock` field,
+  while pipeline's one - via `pipeline_clock` field in callback contexts. Both of
+  them can be used for starting timers.
+  """
+  defmacro def_clock(doc \\ nil) do
```
bblaszkow06: Not added to `.formatter.exs`

lib/membrane/element/base.ex
```diff
@@ -201,15 +232,79 @@ defmodule Membrane.Element.Base do
...
+    user_doc_part =
+      if doc do
+        "\n#{doc}"
```
bblaszkow06: This newline is not needed if you use `user_doc_part` as you do right now. You are generating 2 empty lines if the user does not define docs. Since that's ok for Markdown, why the default value for the doc isn't empty string?

lib/membrane/sync.ex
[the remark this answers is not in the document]
bblaszkow06: Rename this variable, it is leftover of the previous, quite poor name for the registered processes

lib/membrane/sync.ex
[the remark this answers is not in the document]
bblaszkow06: Rename this variable, it is leftover of the previous, quite poor name for the registered processes

test/membrane/clock_test.exs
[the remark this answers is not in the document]
bblaszkow06: That's a fair point, I don't know why is cast used instead of call

lib/membrane/pipeline/spec.ex
```diff
@@ -43,6 +43,31 @@ defmodule Membrane.Pipeline.Spec do
...
+  results in synchronizing all sinks in the pipeline, while passing list of groups
+  of elements synchronizes all elements in each group. It is worth mentioning
+  that to keep the stream synchronized all involved elements need to rely on
+  the same clock.
+
+  By default, no elements are synchronized.
+
+  Sample definitions:
+    
+    %Spec{stream_sync: [[:element1, :element2], [:element3, :element4]]}
```
bblaszkow06: This is not surrounded by ```

lib/membrane/clock.ex
```diff
@@ -0,0 +1,247 @@
...
+  which are messages containing amount of time until the next update.
+  For example, a sink playing audio to the sound card can send an update before
+  each write to the soundcard buffer (for practical reasons that can be done every
+  100 or 1000 writes). Although it might be more intuitive to send updates with
+  the time passed, in practice the described approach turns out to be more convenient,
+  as it simplifies the first update.
+
+  Basing on updates, Clock calculates a _ratio_ between a reference time and
+  the time provided by the custom clock. The reference time can be configured with
+  `:time_provider` option. The ratio is broadcasted (see t:ratio_t)
```
bblaszkow06: `t:ratio_t` should be wrapped with ``

test/membrane/pipeline_test.exs
```diff
@@ -5,7 +5,7 @@ defmodule Membrane.PipelineTest do
   use ExUnit.Case
 
   def state(_ctx) do
-    [state: %State{}]
+    [state: struct(State)]
```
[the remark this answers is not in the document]
kanes115: Clearly it' s done to avoid the constraint put by `@enforce_key`. As these are unit tests, that is fine but I think it is better to explicitly put `%State{module: nil, clock_proxy: nil}`

lib/membrane/core/element/state.ex
```diff
@@ -29,6 +29,9 @@ defmodule Membrane.Core.Element.State do
           playback: Playback.t(),
           playback_buffer: PlaybackBuffer.t(),
           delayed_demands: %{{Pad.ref_t(), :supply | :redemand} => :sync | :async},
+          synchronization: %{
+            timers: %{Timer.id_t() => Timer.t()}
```
bblaszkow06: This typespec is not complete

lib/membrane/core/element/state.ex
```diff
@@ -54,19 +58,26 @@ defmodule Membrane.Core.Element.State do
   @doc """
   Initializes new state.
   """
-  @spec new(module, Element.name_t()) :: t
-  def new(module, name) do
+  @spec new(%{module: module, name: Element.name_t(), clock: Clock.t()}) :: t
```
bblaszkow06: No sync field in typespec

lib/membrane/core/element/action_handler.ex
```diff
@@ -27,6 +27,11 @@ defmodule Membrane.Core.Element.ActionHandler do
 
   @spec do_handle_action(Action.t(), callback :: atom, params :: map, State.t()) ::
           State.stateful_try_t()
+  defp do_handle_action({action, _}, :handle_init, _params, state)
+       when action not in [:latency] do
```
bblaszkow06: This new action lacks  doc and typespec in `Membrane.Element.Action`

lib/membrane/core/element/action_handler.ex
```diff
@@ -138,6 +143,23 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+
+  defp do_handle_action({:start_timer, {id, interval}}, cb, params, state) do
+    do_handle_action({:start_timer, {id, interval, state.pipeline_clock}}, cb, params, state)
+  end
+
+  defp do_handle_action({:stop_timer, id}, _cb, _params, state) do
+    TimerController.stop_timer(id, state)
+  end
+
+  defp do_handle_action({:latency, latency}, _cb, _params, state) do
```
bblaszkow06: This new action lacks  doc and typespec in `Membrane.Element.Action`

lib/membrane/sync.ex
```diff
@@ -0,0 +1,199 @@
...
+    end
+  end
+
+  defp ready_to_sync?(processes) do
+    not (processes |> Bunch.KVList.any_value?(&(&1.status != :sync)))
+  end
+
+  defp ensure_syncs_released(processes) do
+    processes_data = processes |> Map.values()
+    max_latency = processes_data |> Enum.map(& &1.latency) |> Enum.max(fn -> 0 end)
```
bblaszkow06: `Enum.max_by` can be used instead

lib/membrane/element/action.ex
```diff
@@ -157,6 +157,28 @@ defmodule Membrane.Element.Action do
...
+
+  Timers use `Process.send_after/3` under the hood.
+  """
+  @type start_timer_t ::
+          {:start_timer,
+           {id :: any, interval :: Time.t(), clock :: Clock.t()}
+           | {id :: any, interval :: Time.t()}}
+
+  @typedoc """
+  Stops a timer started with `t:timer_t` action.
```
bblaszkow06: Invalid type, should be `t:start_timer_t/0`

lib/membrane/element/action.ex
```diff
@@ -157,6 +157,28 @@ defmodule Membrane.Element.Action do
   """
   @type playback_change_t :: {:playback_change, :suspend | :resume}
 
+  @typedoc """
+  Starts a timer that will invoke `c:Membrane.Element.Base.handle_tick/3` callback
+  every `interval` according to the given `clock`.
+
+  If no clock is passed, pipeline clock is chosen. The timer's `id` is passed
+  to the `c:Membrane.Element.Base.handle_tick/3` callback and can be used for
+  stopping it (`t:stop_timer_t`).
```
bblaszkow06: Invalid type, should be `t:stop_timer_t/0`

kanes115: It is `t:stop_timer_t`, isn't it?

lib/membrane/sync.ex
kanes115: Well, but eventually we want to have latency and not a map in which the latency is the biggest, so I guess that is fine.

lib/membrane/core/element/action_handler.ex
kanes115: And also lacks tests, I think.

lib/membrane/sync.ex
bblaszkow06: Getting latency from map after `Enum.max_by` saves one iteration over entries in map :wink: But if you think this is more readable I can live with that

lib/membrane/clock.ex
```diff
@@ -0,0 +1,247 @@
...
+  @typedoc """
+  Update message received by the Clock. It should contain the time till the next
+  update.
+  """
+  @type update_t ::
+          {:membrane_clock_update,
+           milliseconds ::
+             non_neg_integer
+             | Ratio.t()
+             | {nominator :: non_neg_integer, denominator :: pos_integer}}
```
bblaszkow06: numerator/denominator

lib/membrane/core/element/timer_controller.ex
```diff
@@ -0,0 +1,63 @@
...
+             ActionHandler,
+             %{context: context},
+             [timer_id],
+             state
+           ) do
+      state
+      |> Bunch.Access.update_in([:synchronization, :timers, timer_id], &Timer.tick/1)
+      ~> {:ok, &1}
+    else
+      false -> {:ok, state}
```
bblaszkow06: Error from `:handle_tick` is not properly handled

lib/membrane/core/timer.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Membrane.Core.Timer do
+  @moduledoc false
+  require Membrane.Core.Message
+  alias Membrane.Core.Message
+  alias Membrane.Time
+
+  @enforce_keys [:interval, :clock, :init_time, :id]
```
bblaszkow06: No typespec

lib/membrane/sync.ex
kanes115: Well, with `max_by` we would need to do something like this:
`%{latency: max_latency} = processes_data |> Enum.max_by(& &1.latency, fn -> %{latency: 0} end)`
which I don't like (this artificial default value) and the performance is definitely negligible.

lib/membrane/element/base.ex
```diff
@@ -183,13 +191,43 @@ defmodule Membrane.Element.Base do
               state :: Element.state_t()
             ) :: callback_return_t
 
+  @doc """
+  Callback invoked upon each timer tick. A timer can be started with `Action.timer_t`
```
bblaszkow06: Full name is needed to generate proper link in docs, namely `t:Membrane.Element.Action.timer_t/0`

lib/membrane/pipeline.ex
```diff
@@ -583,18 +567,31 @@ defmodule Membrane.Pipeline do
...
+    children_data = state |> State.get_children() |> Map.values()
+    children_pids = children_data |> Enum.map(& &1.pid)
+    children_pids |> Enum.each(&change_playback_state(&1, new))
+    :ok = toggle_syncs_active(old, new, children_data)
     state = %{state | pending_pids: children_pids |> MapSet.new()}
     PlaybackHandler.suspend_playback_change(state)
   end
 
+  defp toggle_syncs_active(:prepared, :playing, children_data) do
+    do_toggle_syncs_active(children_data, &Sync.activate/1)
```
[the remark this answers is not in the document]
bblaszkow06: Bump https://github.com/membraneframework/membrane-core/pull/160/files#r316291648

lib/membrane/sync.ex
```diff
@@ -0,0 +1,199 @@
+defmodule Membrane.Sync do
```
bblaszkow06: Not asigned to a category in mix.exs (together with `Membrane.PlaybackState` and `Membrane.Core.Playback`)

test/membrane/clock_test.exs
```diff
@@ -0,0 +1,161 @@
...
+
+  test "all subscribers should receive the same ratio after each update" do
+    clock = @module.start_link!(time_provider: fn -> receive do: (time: t -> t) end)
+
+    fn ->
+      Task.start_link(fn ->
+        @module.subscribe(clock)
+        assert_receive {:membrane_clock_ratio, ^clock, @initial_ratio}
+        # TODO why do we get integers and not floats?
+        ratio = trunc(20 / (13 - 3))
```
bblaszkow06: Answering TODO: because that's how `Ratio` works and we use it to keep precision. However I belive it makes more sense to make this line `Ratio.new(20, (13-3))`

PR #161, opened 2019-04-25
Fix/buffer action

bblaszkow06:


lib/membrane/core/element/action_handler.ex
```diff
@@ -259,7 +245,8 @@ defmodule Membrane.Core.Element.ActionHandler do
       state
     )
 
-    withl pad: :ok <- PadModel.assert_data(state, pad_ref, %{direction: :output}),
+    withl pad: {:ok, pad_data} <- PadModel.get_data(state, pad_ref),
+          direction: %{direction: :output} <- pad_data,
           do: accepted_caps = PadModel.get_data!(state, pad_ref, :accepted_caps),
```
mat-hek: This line can be removed and `pad_data.accepted_caps` used instead

lib/membrane/core/element/action_handler.ex
```diff
@@ -259,7 +245,8 @@ defmodule Membrane.Core.Element.ActionHandler do
...
           caps: true <- Caps.Matcher.match?(accepted_caps, caps) do
       {%{pid: pid, other_ref: other_ref}, state} =
```
mat-hek: These values can be extracted from `pad_data`, and `PadModel.set_pad_data` can be used below

lib/membrane/core/element/action_handler.ex
```diff
@@ -150,22 +127,25 @@ defmodule Membrane.Core.Element.ActionHandler do
...
-  defp do_handle_action(_action, _callback, _params, state) do
-    {{:error, :invalid_action}, state}
+  defp do_handle_action(action, callback, _params, state) do
+    raise CallbackError, kind: :invalid_action, action: action, callback: {state.module, callback}
   end
 
   @impl CallbackHandler
   def handle_actions(actions, callback, handler_params, state) do
-    actions_after_redemand =
+    {redemand, actions_after_redemand} =
```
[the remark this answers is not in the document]
bblaszkow06: There can be no `:redemand` action and therefore it can be an empty list

lib/membrane/core/element/action_handler.ex
```diff
@@ -150,22 +127,25 @@ defmodule Membrane.Core.Element.ActionHandler do
...
         _ -> true
       end)
-      |> Enum.drop(1)
+      |> Enum.split(1)
 
     if actions_after_redemand != [] do
-      {{:error, :actions_after_redemand}, state}
+      raise ActionError,
+        reason: :actions_after_redemand,
+        action: redemand |> Enum.at(0),
```
mat-hek: hd

lib/membrane/core/element/action_handler.ex
```diff
@@ -133,21 +133,26 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+      end)
 
-    if actions_after_redemand != [] do
-      raise ActionError,
-        reason: :actions_after_redemand,
-        action: redemand |> Enum.at(0),
-        callback: {state[:module], callback}
-    else
-      super(actions |> join_buffers(), callback, handler_params, state)
+    case {redemands, actions_after_redemands} do
```
mat-hek: I would pipe here and place names inside the match, bcwyp

PR #162, opened 2019-05-06
Refactor/input buffer

bblaszkow06:


lib/membrane/core/element/buffer_controller.ex
```diff
@@ -68,11 +68,11 @@ defmodule Membrane.Core.Element.BufferController do
   defp handle_buffer_pull(pad_ref, buffers, state) do
     PadModel.assert_data!(state, pad_ref, %{direction: :input})
 
-    with {:ok, old_pb} <- PadModel.get_data(state, pad_ref, :buffer),
-         {:ok, pb} <- old_pb |> InputBuffer.store(buffers) do
-      state = PadModel.set_data!(state, pad_ref, :buffer, pb)
+    with {:ok, old_buffer} <- PadModel.get_data(state, pad_ref, :buffer),
```
mat-hek: Referring to the input buffer with just 'buffer' makes it easily confusable with `Membrane.Buffer`. I would consider having `input_buf` everywhere, including the field name in pad data.

lib/membrane/core/input_buffer.ex
```diff
@@ -235,38 +222,35 @@ defmodule Membrane.Core.InputBuffer do
...
-           current_size: size,
-           preferred_size: pref_size,
-           demand: demand,
-           min_demand: min_demand
-         } = pb,
-         new_demand
-       )
-       when size < pref_size and demand + new_demand > 0 do
-    to_demand = max(demand + new_demand, min_demand)
+  @spec send_demands(t(), pid(), Pad.ref_t()) :: t()
```
mat-hek: I would remove this and make `new` and `take` return the demand. It can be then easily sent by the caller

bblaszkow06: I'm not a big fan of this solution:
1. It would complicate the returned value
2. Every caller would have to match on a returned result and conditionally send a demand (if > 0)
3. Every caller would have done "know" how the message looks (if we added a new argument there would be many places to modify)

mat-hek: So maybe we should move the `send_demands` function to DemandHandler? This would fix 2. and 3., while 1. is IMHO not that painful

lib/membrane/core/element/demand_handler.ex
```diff
@@ -133,10 +133,12 @@ defmodule Membrane.Core.Element.DemandHandler do
 
   @spec do_supply_demand(Pad.ref_t(), pos_integer, State.t()) :: State.stateful_try_t()
   defp do_supply_demand(pad_ref, size, state) do
-    pb_output =
-      state |> PadModel.get_and_update_data(pad_ref, :buffer, &(&1 |> InputBuffer.take(size)))
+    pad_data = state |> PadModel.get_data!(pad_ref)
+    {buffer_output, new_buffer} = pad_data.buffer |> InputBuffer.take(size)
```
mat-hek: Fix confusing 'buffer' variable name

lib/membrane/core/element/demand_handler.ex
```diff
@@ -133,10 +133,12 @@ defmodule Membrane.Core.Element.DemandHandler do
...
 
-    with {{:ok, {_pb_status, data}}, state} <- pb_output,
+    with {:ok, {_buffer_status, data}} <- buffer_output,
```
mat-hek: Fix confusing 'buffer' variable name

lib/membrane/core/element/demand_handler.ex
```diff
@@ -133,10 +133,12 @@ defmodule Membrane.Core.Element.DemandHandler do
...
+         new_buffer = new_buffer |> InputBuffer.send_demands(pad_data.pid, pad_data.other_ref),
```
mat-hek: Fix confusing 'buffer' variable name

lib/membrane/core/element/buffer_controller.ex
```diff
@@ -68,11 +68,11 @@ defmodule Membrane.Core.Element.BufferController do
   defp handle_buffer_pull(pad_ref, buffers, state) do
     PadModel.assert_data!(state, pad_ref, %{direction: :input})
 
-    with {:ok, old_pb} <- PadModel.get_data(state, pad_ref, :buffer),
-         {:ok, pb} <- old_pb |> InputBuffer.store(buffers) do
-      state = PadModel.set_data!(state, pad_ref, :buffer, pb)
+    with {:ok, old_input_buf} <- PadModel.get_data(state, pad_ref, :buffer),
```
mat-hek: I would also change this `:buffer` in pad data to something like `:input_buf`

lib/membrane/core/input_buffer.ex
bblaszkow06: As we discussed, the way to go is to get back to sending messages directly from `take` and `new`, but pass pid and pad name as arguments to these functions

PR #164, opened 2019-05-06
Improve Testing API

### LINKED ISSUE #157: Improve Membrane.Testing API
Testing API is now quite confusing. Here are some reasons:
- It's difficult to determine from the name and arguments how does `assert_receive_message` differ from usual `assert_receive`
- Checking for something as simple and common as End Of Stream requires subscribing on `handle_notification`
- Messages from testing elements are received with `assert_receive`, while messages passed through the pipeline - with `assert_receive_message`
- There is no way to distinguish pipeline or element that sent a message if there are many of them

Suggested changes are:
- Make all testing elements send notifications to pipeline instead of the testing process
- By default subscribe on all pipeline callbacks
- Instead of `assert_receive_message`, make separate assertions like below:
  - `assert_pipeline_notified(pipeline_pid, element_name, message)`
  - `assert_pipeline_playback_changed(pipeline_pid, previous_playback_state, current_playback_state)`
  - `assert_pipeline_received(pipeline_pid, message)` for `:handle_other`
- Create proper subscription options for these assertions (`:notification` / `:playback` / `:other`)

Hajto:
Closes #157 

# Testing

For elements Sink and Source of `Membrane.Testing` workspace there are defined unit tests. Pipeline is tested in integration with Assertions, but Assertions do have sad path tests which are unit tests. Assertions are tested not only to test the behavior but also to force macros residing in Assertions module. If they are not compiled at least once there is no guarantee they will work.

## Possibilities of future improvement

`Membrane.Testing.Assertions.assert_message_receive/4` could be replaced with custom assert receive so that messages can be tailored to needs. For example `assert_sink_processed_buffer/4` could list only buffers that got received and not other messages in case of failure.

# Suggested review order
 - [ ] testing API interface
 - [ ] testing pipeline implementation
 - [ ] test Source and Sink implementation
 - [ ] assertions implementation and tests

Please read comments annotated by `# Note:` and comments on github.


lib/membrane/testing/assertions.ex
```diff
@@ -0,0 +1,229 @@
...
+  Asserts that pipeline's playback state changed from one to another.
+
+  Supplied change must be a valid one. Currently playback has 3
+  `t:Membrane.Pipeline.playback_state_t/0`, namely:
+
+   - `:stopped`
+   - `:prepared`
+   - `:playing`
+
+   those states must change in a sequence so change from `:stopped` to
```
Hajto: This description could be better, but I do lack ideas. I want to communicate to a programmer that might use this and doesn't know membrane internally to not do stupid thigs.

lib/membrane/testing/sink.ex
```diff
@@ -1,17 +1,29 @@
 defmodule Membrane.Testing.Sink do
   @moduledoc """
-  Sink Element that will send every buffer it receives to the pid passed as argument.
+  Sink Element that allows asserting on buffers it receives.
```
mat-hek: - This is true only if the element is used in the testing pipeline. Mention what the element does itself.
- Not only buffers, but also events

lib/membrane/testing/source.ex
```diff
@@ -7,33 +7,44 @@ defmodule Membrane.Testing.Source do
...
-                It is an action generator that takes two arguments.
-                First argument is counter which is incremented by 1 every call
-                and second argument represents size of demand.
+                If `output` is an enumerable with `Membrane.Payload.t()` then
+                buffer containing those payloads will be sent through the
+                `:output` pad and followed by `Membrane.Event.EndOfStream`.
+
+                If `output` is a function then it will be invoked each time
+                `handle_demand` is invoked. It is an action generator that takes
+                two arguments. The first argument is counter which is incremented by
```
mat-hek: This is true only in case of the default implementation. In general, it is a state of a generator

lib/membrane/testing/source.ex
```diff
@@ -7,33 +7,44 @@ defmodule Membrane.Testing.Source do
...
+  alias Membrane.Event.EndOfStream
+  require Bunch.Code
 
   def_output_pad :output, caps: :any
 
-  def_options actions_generator: [
-                type: :function,
-                spec: (non_neg_integer, non_neg_integer -> [Action.t()]),
+  def_options output: [
+                spec: (non_neg_integer, non_neg_integer -> {[Action.t()], integer()}) | Enum.t(),
```
mat-hek: The first argument is not necessarily an integer

lib/membrane/testing/source.ex
```diff
@@ -7,33 +7,44 @@ defmodule Membrane.Testing.Source do
...
               ]
 
   @impl true
-  def handle_init(opts) do
-    {:ok, opts |> Map.merge(%{cnt: 0})}
+  def handle_init(%__MODULE__{output: output} = opts) do
+    opts = Map.from_struct(opts)
+
+    if is_function(output) do
+      {:ok, opts |> Map.merge(%{cnt: 0})}
```
mat-hek: cnt -> generator_state, init with `nil`

lib/membrane/testing/pipeline.ex
```diff
@@ -30,39 +30,55 @@ defmodule Membrane.Testing.Pipeline do
...
+  ## Example usage
 
-  In some cases, you want to get a notification when `Pipeline` reaches a certain state.
-  You can achieve that by monitoring pipeline callbacks.
+  Once options are created we can start the pipeline.
 
-  ```
-  options = %Membrane.Testing.Pipeline.Options {
-    monitored_callbacks: [:handle_prepared_to_playing],
-    test_process: pid
```
mat-hek: Why did you remove this?

lib/membrane/testing/pipeline.ex
```diff
@@ -202,15 +196,8 @@ defmodule Membrane.Testing.Pipeline do
...
-  defp get_callback_name(name) when is_atom(name), do: name
-  defp get_callback_name({name, _}), do: name
-
-  defp notify_parent(message, state) do
-    %{test_process: parent, monitored_callbacks: monitored_callbacks} = state
-
-    if get_callback_name(message) in monitored_callbacks do
-      send(parent, {__MODULE__, message})
-    end
+  defp notify_parent(message, %{test_process: parent} = state) do
```
mat-hek: parent -> test_process

lib/membrane/testing/pipeline.ex
```diff
@@ -75,49 +91,35 @@ defmodule Membrane.Testing.Pipeline do
...
-    @typedoc """
-    Defines supported callback names.
-    """
-    @type pipeline_callback ::
-            :handle_notification
-            | :handle_playing_to_prepared
-            | :handle_prepared_to_playing
-            | :handle_prepared_to_stopped
-            | :handle_stopped_to_prepared
+    defstruct @enforce_keys ++ [:links, :test_process]
```
mat-hek: Maybe override start_link, so that it assigned `self()` to test_process if it is `nil`?

lib/membrane/testing/source.ex
```diff
@@ -7,33 +7,44 @@ defmodule Membrane.Testing.Source do
   use Bunch
   alias Membrane.Buffer
   alias Membrane.Element.Action
+  alias Membrane.Event.EndOfStream
+  require Bunch.Code
```
mat-hek: why?

lib/membrane/testing/assertions.ex
```diff
@@ -0,0 +1,229 @@
...
+  You can match for exact value:
+
+      assert_sink_processed_buffer(pid, :the_sink ,%Membrane.Buffer{payload: ^specific_payload})
+
+  You can also use pattern to extract data from the buffer:
+
+      assert_sink_processed_buffer(pid, :sink, %Buffer{payload: <<data::16>> <> <<255>>})
+      do_something(data)
+  """
+  defmacro assert_sink_processed_buffer(
```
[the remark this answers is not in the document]
Hajto: `assert_sink_buffer` was chosen by our traditional means.

lib/membrane/testing/assertions.ex
Hajto: Last part is hard(ish) to achieve because we need to pass some form of pattern to underlaying `ExUnit.Assertions.assert_receive/3`, but:
 - it rejects `c when c in [:handle..., :handle] because:
```
   cannot find or invoke local when/2 inside match. 
Only macros can be invoked in a match and they must be defined before their invocation. 
Called as: callback when :erlang."=:="(callback, :handle_prepared_to_playing)
```
 - it rejects `c when c == :a or c == :b` with reason:
```
invalid expression in match, or is not allowed on patterns such as function clauses, 
case clauses or on the left side of the = operator
```

I am open to suggestions.

lib/membrane/testing/pipeline.ex
Hajto: I thought I already overriden start_link and it automagically bound to self of the caller.

lib/membrane/testing/source.ex
Hajto: Forgot to delete it.

lib/membrane/testing/pipeline.ex
```diff
@@ -1,23 +1,23 @@
 defmodule Membrane.Testing.Pipeline do
   @moduledoc """
-  This Pipeline was created to reduce testing boilerplate and ease communication with it's elements.
-  It also provides utility for receiving messages when `Pipeline` playback state changes
-  and notifications it receives.
-
-  ## Starting Pipeline
+  This Pipeline was created to reduce testing boilerplate and ease communication
+  with its elements. It also provides utility for receiving messages when
+  `Pipeline` playback state changes and notifications it receives.
```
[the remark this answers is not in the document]
Hajto: Neither I nor the Grammarly find anything off about grammar of these two sentences. Could you be a bit more specific?

lib/membrane/testing/pipeline.ex
Hajto: I've added the default clause for when `test_process: nil`, but I must remain non-enforced key (since struct default values seem to be calculated compile time).

test/membrane/integration/demands_test.exs
```diff
@@ -4,54 +4,52 @@ defmodule Membrane.Integration.DemandsTest do
...
 
     0..(demand - 1)
-    |> Enum.each(fn i ->
-      pattern = pattern_gen.(i)
-      assert_message(pattern)
-    end)
+    |> assert_buffers_received(pid)
 
     pattern = pattern_gen.(demand)
     refute_receive ^pattern
```
Hajto: This won't work anymore. There is a need for `refute_sink_buffer` and a like then.

lib/membrane/testing/assertions.ex
```diff
@@ -0,0 +1,229 @@
...
+  """
+  require ExUnit.Assertions
+
+  @default_timeout 2000
+
+  # Note: pattern value_x = unquote(x) is used to workaround allowing both
+  # Note: both function, value and variable to be passed as an argument.
+
+  @doc false
+  defmacro assert_message_receive(
```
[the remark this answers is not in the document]
Hajto: @mat-hek 

If I call this `assert_receive_pipeline` wouldn't it slightly clash with `assert_pipeline_receive`? Those two names are different by they are painstakingly similar and will sooner or later they might cause issues if we ever need to discuss anything about this module. 

My suggestion would be naming it something like `base_assert`, `base_testing_assertion` because it will never be exposed or `positive_assertion` since there will be need of adding `refute_xxx` which could base on `negative_assertion` (there are use cases in `demands_test.exs`, which is now broken).

lib/membrane/pipeline.ex
```diff
@@ -655,7 +655,11 @@ defmodule Membrane.Pipeline do
       @impl true
       def handle_spec_started(_new_children, state), do: {:ok, state}
 
-      defoverridable start: 2,
+      defoverridable start: 0,
```
Hajto: This was added to solve warning when overriding `start` and `start_link`

lib/membrane/testing/assertions.ex
```diff
@@ -3,160 +3,210 @@ defmodule Membrane.Testing.Assertions do
...
-          previous_playback_state,
-          current_playback_state
-        )
+    is_previous_ignored = match?({:_, _, _}, previous_state)
+    is_current_ignored = match?({:_, _, _}, current_state)
+    are_arguments_values = !is_previous_ignored and !is_current_ignored
 
-      assert_message_receive(pipeline_pid, ^callback_name, timeout, nil)
-    else
+    if are_arguments_values and {previous_state, current_state} not in valid_changes do
```
mat-hek: This won't work when a variable is passed, like`_variable` or `^variable`. Better check if previous and current state are atoms

mat-hek:
Can you also update tests using this in an element? e.g. Tee?

Hajto:
> Can you also update tests using this in an element? e.g. Tee?

Yes

lib/membrane/testing/assertions.ex
Hajto: Wouldn't be `^variable` considered a pattern? This macro does not support receiving patterns as arguments. I all macros in this modules i explicitly stated when to pass pattern as argument. If someones wants to pass (for any reason) state as variable he can achieve that by:

```
assert_pipeline_playback_changed(pipeline, previous, current)
```

and using `_variable` would trigger compiler to emit a warning since _ is used to prefix variables that won't be used.

lib/membrane/testing/pipeline.ex
```diff
@@ -30,94 +30,109 @@ defmodule Membrane.Testing.Pipeline do
...
+  We can also assert that the `Membrane.Testing.Sink` processed a specific
+  buffer.
+
+      assert_sink_buffer(pipeline, :sink ,%Membrane.Buffer{payload: 1})
+
+  ## Assertions
 
-  Then it is a matter of waiting for the right message to come.
-  You can do that easily by using `Membrane.Testing.Pipeline.Assertions.assert_receive_message/3`.
+  Using this module enables usage of various assertions which are described in
```
kanes115: "Using enables usage"? I would go for something that make it clear this is about keyword `use` maybe?

lib/membrane/testing/assertions.ex
```diff
@@ -0,0 +1,353 @@
...
+  Such expression will flunk if `the_sink` received or will receive caps with
+  property equal to value of `val` variable.
+  """
+  defmacro refute_sink_caps(pipeline, element_name, caps_pattern, timeout \\ @default_timeout) do
+    do_sink_caps(&refute_receive_from_pipeline/3, pipeline, element_name, caps_pattern, timeout)
+  end
+
+  defp do_sink_caps(assertion, pipeline, sink_name, caps, timeout) do
+    quote do
+      element_name_value = unquote(sink_name)
```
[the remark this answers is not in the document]
mat-hek: This may be problematic as `bind_quoted` disables unquoting, which is probably necessary in this case

lib/membrane/testing/pipeline.ex
```diff
@@ -170,47 +176,45 @@ defmodule Membrane.Testing.Pipeline do
...
+  end
 
   @impl true
   def handle_other({:for_element, element, message}, state) do
     {{:ok, forward: {element, message}}, state}
   end
 
   def handle_other(message, state),
-    do: notify_parent({:handle_other, message}, state)
+    do: notify_test_process({:handle_other, message}, state)
```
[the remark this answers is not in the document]
mat-hek: `notify_test_process` already wraps messages with `{__MODULE__, ...}` tuple

lib/membrane/testing/pipeline.ex
mat-hek: This is not about `use` keyword ;) Tried to write it more clearly

PR #168, opened 2019-07-08
Testing pipeline

DominikWolek:
# Custom module option

Element Pipeline of `Membrane.Testing` have `:module` option now, which allows to pass some custom callbacks.


lib/membrane/testing/pipeline.ex
```diff
@@ -45,6 +45,19 @@ defmodule Membrane.Testing.Pipeline do
     }
     ```
 
+  You can also pass custom pipeline module, by using `:module` field of
```
Hajto:
```
You can also pass **a** custom pipeline module, by using `:module` field of
  `Membrane.Testing.Pipeline.Options` struct. Every callback of **the** module will be executed before **the** callbacks of Testing.Pipeline.
```

You can use a/an instead of the or vice versa. Up to you to decide.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+      {{reaction, actions}, state} -> {{reaction, actions}, state}
+    end
+  end
+
+  defp pipeline_eval(
+         :handle_init,
+         _custom_args,
+         _function,
+         _args,
+         %{module: nil} = _state
```
Hajto: No need to assign to `_state` if you are not using it anywhere.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+    case result do
+      {:ok, state} -> {{:ok, []}, state}
+      {{reaction, actions}, state} -> {{reaction, actions}, state}
+    end
+  end
+
+  defp pipeline_eval(
+         :handle_init,
+         _custom_args,
+         _function,
```
Hajto: I would call this `_fun` just to make it shorter.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+      Map.put(
+        new_state,
+        :module_state,
+        pipeline_eval(:handle_init, args, nil, nil, new_state)
+      )
+
     {{:ok, spec}, new_state}
   end
 
+  defp wrap_result(result) do
```
Hajto: I would personally do `defp wrap_result({:ok, state})....)` instead of `case` in this particular example, but it's up to you to decide.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+    case result do
+      {:ok, state} -> {{:ok, []}, state}
+      {{reaction, actions}, state} -> {{reaction, actions}, state}
+    end
+  end
+
+  defp pipeline_eval(
```
Hajto: Another possible solution that takes into account both compactness and readability of the code is to use body less definition.

```
def fun(arg_name, arg2_name)
def fun(1 ,_ ,_), do: :blah
def fun(2,_,_), do: :blah2
```

Hajto: But since there is not a lot of cases I would use this only if argument names are too long to fit into single line.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+       do: nil
+
+  defp pipeline_eval(
+         :handle_init,
+         custom_args,
+         _function,
+         _args,
+         %{module: module} = _state
+       ) do
+    with _custom_result = {{:ok, _spec}, state} <-
```
Hajto: If you don't use `_custom_result` then just omit it. It is not needed by any means.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+         %{module: nil} = _state
+       ),
+       do: nil
+
+  defp pipeline_eval(
+         :handle_init,
+         custom_args,
+         _function,
+         _args,
+         %{module: module} = _state
```
Hajto: Same as above.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+           |> wrap_result,
+         do: state
+  end
+
+  defp pipeline_eval(
+         _custom_function,
+         _custom_args,
+         function,
+         args,
+         %{module: nil} = _state
```
Hajto: Same as above.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+      {l, {:ok, _actions_r}} -> l
+      {l, _r} -> l
+    end
+  end
+
+  defp combine_results({actions_l, state_l}, {actions_r, state_r}),
+    do: {combine_actions(actions_l, actions_r), Map.put(state_l, :custom_pipeline_state, state_r)}
+
   @impl true
   def handle_stopped_to_prepared(state),
```
Hajto: In cases like this (function definition with `,do: ` and a single function call that spans o lot of lines I would actually switch to `do end` syntax.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,38 +193,174 @@ defmodule Membrane.Testing.Pipeline do
...
+
+  defp pipeline_eval(
+         custom_function,
+         custom_args,
+         function,
+         args,
+         %{module: module} = state
+       ) do
+    with custom_result = {{:ok, _actions}, _state} <-
+           apply(module, custom_function, custom_args ++ state[:custom_pipeline_state])
```
Hajto: 1. You can use state.custom_pipeline_state
2. Is it possible that state.custom_pipeline_state could return nil?

lib/membrane/testing/pipeline.ex
DominikWolek: Done

lib/membrane/testing/pipeline.ex
DominikWolek: Changed name to `eval`.

lib/membrane/testing/pipeline.ex
DominikWolek: Done

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
DominikWolek: 1. Done
2. Yes, i.e. when `state.module` return `nil`.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,44 +218,160 @@ defmodule Membrane.Testing.Pipeline do
...
+  defp eval(:handle_init, custom_args, _, _, %State{module: module}) do
+    {{:ok, _spec}, state} = apply(module, :handle_init, custom_args) |> wrap_result
+    state
+  end
+
+  defp eval(_, _, function, args, %State{module: nil}),
+    do: apply(function, args)
+
+  defp eval(custom_function, custom_args, function, args, %State{module: module} = state) do
+    with custom_result = {{:ok, _actions}, _state} <-
```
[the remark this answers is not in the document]
DominikWolek: will work because `wrap_result {:ok, state} = {{:ok, []}, state}`

lib/membrane/testing/pipeline.ex
```diff
@@ -101,17 +114,46 @@ defmodule Membrane.Testing.Pipeline do
...
+    State of the pipeline defined by Module.
+    """
+
+    @enforce_keys [:test_process, :module]
+    defstruct @enforce_keys ++ [:custom_pipeline_state]
+
+    @type t :: %__MODULE__{
+            test_process: pid() | nil,
+            module: module() | nil,
+            custom_pipeline_state: list() | map() | nil
```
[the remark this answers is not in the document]
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,44 +218,160 @@ defmodule Membrane.Testing.Pipeline do
       links: links
     }
 
-    new_state = Map.take(args, [:test_process])
+    new_state = %State{test_process: args.test_process, module: args.module}
+
+    new_state = %State{
+      new_state
+      | custom_pipeline_state: eval(:handle_init, args, nil, nil, new_state)
```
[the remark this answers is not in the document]
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,44 +218,160 @@ defmodule Membrane.Testing.Pipeline do
...
 
-  def handle_other(message, state),
-    do: notify_test_process({:handle_other, message}, state)
+  @impl true
+  def handle_spec_started(elements, %State{} = state),
+    do:
+      eval(
+        :handle_spec_started,
+        [elements],
+        &do_nothing/2,
```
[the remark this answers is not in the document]
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
```diff
@@ -176,44 +218,160 @@ defmodule Membrane.Testing.Pipeline do
...
 
   @impl true
-  def handle_prepared_to_playing(state),
-    do: notify_playback_state_changed(:prepared, :playing, state)
+  def handle_stopped_to_prepared(%State{} = state),
+    do:
+      eval(
+        :handle_stopped_to_prepared,
+        [],
+        &notify_playback_state_changed/3,
```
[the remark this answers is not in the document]
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
```diff
@@ -234,34 +229,39 @@ defmodule Membrane.Testing.Pipeline do
...
     end
   end
 
+  defp eval(custom_function, custom_args, result)
+
+  defp eval(_, _, {_, %State{module: nil}} = result),
+    do: result
+
+  defp eval(:handle_init, custom_args, {_, %State{module: module}} = result) do
+    {{:ok, _spec}, custom_state} = apply(module, :handle_init, custom_args) |> wrap_result
```
mat-hek: In the case when a module is provided, I would take the spec from handle init and forbid passing it through options

lib/membrane/testing/pipeline.ex
```diff
@@ -219,13 +220,7 @@ defmodule Membrane.Testing.Pipeline do
...
-    new_state = %State{
-      new_state
-      | custom_pipeline_state: eval(:handle_init, args, nil, nil, new_state)
-    }
-
-    {{:ok, spec}, new_state}
+    eval(:handle_init, args, {{:ok, spec}, new_state})
   end
 
   defp wrap_result({:ok, state}),
```
mat-hek: Move all the private function to one place, preferably to the end of the file, and reorder them so that helpers were first, then helpers to helpers etc

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
DominikWolek: Done.

lib/membrane/testing/pipeline.ex
```diff
@@ -206,73 +205,24 @@ defmodule Membrane.Testing.Pipeline do
...
-      {l, :ok} -> l
-      {{:ok, actions_l}, {:ok, actions_r}} -> {:ok, actions_l ++ actions_r}
-      {_l, r} -> r
-    end
-  end
-
-  defp combine_results({custom_actions, custom_state}, {actions, state}) do
-    {combine_actions(custom_actions, actions),
-     Map.put(state, :custom_pipeline_state, custom_state)}
+  def handle_init(options) do
```
mat-hek: Assert here that there are no children nor links and module is present

DominikWolek: Done

lib/membrane/buffer/metric.ex
```diff
@@ -8,7 +8,7 @@ defmodule Membrane.Buffer.Metric do
 
   @type unit_t :: :buffers | :bytes
 
-  @callback pullbuffer_preferred_size() :: pos_integer
+  @callback input_buf_preferred_size() :: pos_integer
```
Hajto: I would prefer full `buffer` instead of `buf`.

mat-hek: This shows due to merging master, has not changed in this PR

lib/membrane/testing/pipeline.ex
```diff
@@ -163,59 +204,166 @@ defmodule Membrane.Testing.Pipeline do
...
+  defp eval(custom_function, custom_args, function, %State{module: module} = state) do
+    with custom_result = {{:ok, _actions}, _state} <-
+           apply(module, custom_function, custom_args ++ [state.custom_pipeline_state])
+           |> wrap_result do
+      result = function.()
+      combine_results(custom_result, result)
+    end
+  end
+
+  defp eval(custom_function, custom_args, result)
```
mat-hek: all these clauses can be merged to one doing `eval(custom_function, custom_args, fn -> result, state)`

lib/membrane/testing/pipeline.ex
```diff
@@ -163,59 +204,166 @@ defmodule Membrane.Testing.Pipeline do
...
   end
 
-  defp notify_test_process(message, %{test_process: test_process} = state) do
+  defp notify_test_process(message, %State{test_process: test_process} = state) do
     send(test_process, {__MODULE__, self(), message})
 
     {:ok, state}
   end
+
+  defp wrap_result({:ok, state}),
```
mat-hek: would rather call this something like unify_result

lib/membrane/testing/pipeline.ex
```diff
@@ -163,59 +204,166 @@ defmodule Membrane.Testing.Pipeline do
...
+  def handle_other(message, %State{} = state),
+    do:
+      eval(
+        :handle_other,
+        [message],
+        fn -> notify_test_process({:handle_other, message}, state) end,
+        state
+      )
+
+  defp default_options(%Options{test_process: nil} = options),
```
mat-hek: supply_default_options ?

DominikWolek: This function name was already specified in `new-testing-api` and merged to master

lib/membrane/testing/pipeline.ex
DominikWolek: Done

lib/membrane/testing/pipeline.ex
DominikWolek: done

lib/membrane/testing/pipeline.ex
mat-hek: ok, anyway you can rename it ;)

PR #170, opened 2019-07-16
Fix order of generated docs for pads and options

bblaszkow06:


lib/membrane/core/element/pad_specs.ex
```diff
@@ -92,7 +92,7 @@ defmodule Membrane.Core.Element.PadsSpecs do
       """
       @spec membrane_pads() :: [{unquote(Pad).name_t(), unquote(Pad).description_t()}]
       def membrane_pads() do
-        @membrane_pad
+        unquote(pads |> Macro.escape())
```
mat-hek: `@membrane_pad |> Enum.reverse()` would be less confusing. BTW I think `@membrane_pad` should be called `@membrane_pads`

PR #172, opened 2019-07-17
replaced travis with circleci

mat-hek:


.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
...
+        environment:
+          MIX_ENV: test
+
+    working_directory: ~/app
+
+    steps:
+      - checkout
+      - run: mix deps.get
+      - run: mix format --check-formatted
+      - run: mix test
```
bblaszkow06: You forgot to add espec

PR #174, opened 2019-07-25
Stream management callbacks

kanes115:
TODO:
- [x] Create action `end_of_stream`
- [ ] Should watcher be informed about start of stream synchronously or asynchronously?

### DIFF
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -28,7 +28,7 @@ defmodule Membrane.Core.Element.ActionHandler do
   @spec do_handle_action(Action.t(), callback :: atom, params :: map, State.t()) ::
           State.stateful_try_t()
   defp do_handle_action({action, _}, cb, _params, %State{playback: %{state: :stopped}} = state)
-       when action in [:buffer, :event, :caps, :demand, :redemand, :forward] and
+       when action in [:buffer, :event, :caps, :demand, :redemand, :forward, :end_of_stream] and
               cb != :handle_stopped_to_prepared do
     {{:error, {:playback_state, :stopped}}, state}
   end
@@ -125,6 +125,16 @@ defmodule Membrane.Core.Element.ActionHandler do
        )
        when is_pad_ref(pad_ref) and is_demand_size(size) and type in [:sink, :filter] do
     supply_demand(pad_ref, size, cb, params[:supplying_demand?] || false, state)
+  end
+
+  defp do_handle_action(
+         {:end_of_stream, pad_ref},
+         _cb,
+         _params,
+         %State{type: type} = state
+       )
+       when is_pad_ref(pad_ref) and type != :sink do
+    send_event(pad_ref, %Event.EndOfStream{}, state)
   end
 
   defp do_handle_action(action, callback, _params, state) do
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -129,7 +129,7 @@ defmodule Membrane.Core.Element.ActionHandler do
 
   defp do_handle_action(
          {:end_of_stream, pad_ref},
-         _cb,
+         _callback,
          _params,
          %State{type: type} = state
        )
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -98,15 +98,15 @@ defmodule Membrane.Core.Element.ActionHandler do
               :handle_end_of_stream
             ] do
     {action, dir} =
-      case {cb, params} do
-        {:handle_process_list, _} ->
+      case cb do
+        :handle_process_list ->
           {:buffer, :output}
 
-        {:handle_caps, _} ->
+        :handle_caps ->
           {:caps, :output}
 
-        {ev_cb, %{direction: dir}} when ev_cb in [:handle_event, :handle_end_of_stream] ->
-          {:event, Pad.opposite_direction(dir)}
+        ev_cb when ev_cb in [:handle_event, :handle_end_of_stream] ->
+          {:event, Pad.opposite_direction(params.direction)}
       end
 
     pads = state |> PadModel.filter_data(%{direction: dir}) |> Map.keys()
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -131,7 +131,7 @@ defmodule Membrane.Core.Element.ActionHandler do
          {:end_of_stream, pad_ref},
          _callback,
          _params,
-         %State{type: type} = state
+         %State{type: type, playback: %{state: :playing}} = state
        )
        when is_pad_ref(pad_ref) and type != :sink do
     send_event(pad_ref, %Event.EndOfStream{}, state)
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -109,11 +109,18 @@ defmodule Membrane.Core.Element.ActionHandler do
           {:event, Pad.opposite_direction(params.direction)}
       end
 
+    result_data =
+      if data == :end_of_stream do
+        %Event.EndOfStream{}
+      else
+        data
+      end
+
     pads = state |> PadModel.filter_data(%{direction: dir}) |> Map.keys()
 
     pads
     |> Bunch.Enum.try_reduce(state, fn pad, st ->
-      do_handle_action({action, {pad, data}}, cb, params, st)
+      do_handle_action({action, {pad, result_data}}, cb, params, st)
     end)
   end
 
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -90,13 +90,23 @@ defmodule Membrane.Core.Element.ActionHandler do
   end
 
   defp do_handle_action({:forward, data}, cb, params, %State{type: :filter} = state)
-       when cb in [:handle_caps, :handle_event, :handle_process_list] do
+       when cb in [
+              :handle_caps,
+              :handle_event,
+              :handle_process_list,
+              :handle_start_of_stream,
+              :handle_end_of_stream
+            ] do
     {action, dir} =
       case {cb, params} do
-        {:handle_process_list, _} -> {:buffer, :output}
-        {:handle_caps, _} -> {:caps, :output}
-        {:handle_event, %{direction: :input}} -> {:event, :output}
-        {:handle_event, %{direction: :output}} -> {:event, :input}
+        {:handle_process_list, _} ->
+          {:buffer, :output}
+
+        {:handle_caps, _} ->
+          {:caps, :output}
+
+        {ev_cb, %{direction: dir}} when ev_cb in [:handle_event, :handle_end_of_stream] ->
+          {:event, Pad.opposite_direction(dir)}
       end
 
     pads = state |> PadModel.filter_data(%{direction: dir}) |> Map.keys()
--- a/lib/membrane/core/element/event_controller.ex
+++ b/lib/membrane/core/element/event_controller.ex
@@ -7,6 +7,7 @@ defmodule Membrane.Core.Element.EventController do
   alias Core.Element.{ActionHandler, PadModel, State}
   alias Element.{CallbackContext, Pad}
   require CallbackContext.Event
+  require CallbackContext.StreamManagement
   require PadModel
   use Core.Element.Log
   use Bunch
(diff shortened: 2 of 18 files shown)


lib/membrane/core/element/action_handler.ex
```diff
@@ -127,6 +127,16 @@ defmodule Membrane.Core.Element.ActionHandler do
     supply_demand(pad_ref, size, cb, params[:supplying_demand?] || false, state)
   end
 
+  defp do_handle_action(
+         {:end_of_stream, pad_ref},
+         _cb,
```
Hajto: If formatter decided to expand it to multiple lines consider renaming cb => callback

lib/membrane/core/element/event_controller.ex
```diff
@@ -50,6 +54,20 @@ defmodule Membrane.Core.Element.EventController do
 
   @spec do_exec_handle_event(Pad.ref_t(), Event.t(), params :: map, State.t()) ::
           State.stateful_try_t()
+  defp do_exec_handle_event(pad_ref, event, params, state)
+       when event in [%Event.StartOfStream{}, %Event.EndOfStream{}] do
+    data = PadModel.get_data!(state, pad_ref)
+    context = CallbackContext.Event.from_state(state)
+
+    CallbackHandler.exec_and_handle_callback(
```
Hajto: I would peronally opt for spliting this function call 
```
callback = stream_event_to_callback(event)
params = params |> Map.merge(%{direction: data.direction})
whatever_this_is = [pad_ref, context]
```
and using those variables to call the function.

For me it would me easier to read and comprehend, but since this is style thing do as you wish.

test/membrane/element/element_test.exs
```diff
@@ -0,0 +1,107 @@
...
+            filter: %TestFilter{target: self()},
+            sink: Testing.Sink
+          ]
+        })
+
+      [pipeline: pipeline]
+    end
+
+    test "causes handle_start_of_stream/3 to be called", %{pipeline: pipeline} do
+      Pipeline.play(pipeline)
```
Hajto: Since last version if any pipeline has use Pipeline it has proxy functions. Which means you don't have to explictly call Membrane.Pipeline.play/1 you can use Any.Pipeline.play/1 or in fact Membrane.Testing.Pipeline.play/1

test/membrane/element/element_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      assert_pipeline_playback_changed(pipeline, _, :playing)
+
+      TestFilter.assert_callback_called(:handle_start_of_stream)
+    end
+
+    test "does not trigger calling callback handle_event/3", %{pipeline: pipeline} do
+      Pipeline.play(pipeline)
```
Hajto: Same as above.

test/membrane/element/element_test.exs
```diff
@@ -0,0 +1,107 @@
...
+            filter: %TestFilter{target: self()},
+            sink: Testing.Sink
+          ]
+        })
+
+      [pipeline: pipeline]
+    end
+
+    test "causes handle_end_of_stream/3 to be called", %{pipeline: pipeline} do
+      Pipeline.play(pipeline)
```
Hajto: same as above

test/membrane/element/element_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      assert_pipeline_playback_changed(pipeline, _, :playing)
+
+      TestFilter.assert_callback_called(:handle_end_of_stream)
+    end
+
+    test "does not trigger calling callback handle_event/3", %{pipeline: pipeline} do
+      Pipeline.play(pipeline)
```
Hajto: same as above.

test/membrane/element/element_test.exs
```diff
@@ -0,0 +1,107 @@
...
+
+    @impl true
+    def handle_demand(_, size, _, _ctx, state), do: {{:ok, demand: {:input, size}}, state}
+
+    @impl true
+    def handle_process(_pad, _buffer, _context, state), do: {:ok, state}
+  end
+
+  describe "Start of stream" do
+    setup do
```
Hajto: Why both start and end of stream can't be tested with same pipeline?

lib/membrane/element/base.ex
```diff
@@ -189,6 +189,25 @@ defmodule Membrane.Element.Base do
   """
   @callback handle_shutdown(state :: Element.state_t()) :: :ok
 
+  @doc """
+  Callback invoked when element receives `Membrane.Event.StartOfStream` event.
```
Hajto: when `AN` element or `THE` element.

lib/membrane/element/base.ex
```diff
@@ -189,6 +189,25 @@ defmodule Membrane.Element.Base do
...
+  """
+  @callback handle_start_of_stream(
+              pad :: Pad.ref_t(),
+              context :: CallbackContext.Event.t(),
+              state :: Element.state_t()
+            ) :: callback_return_t
+
+  @doc """
+  Callback invoked when element receives `Membrane.Event.EndOfStream` event
```
Hajto: Same as above.

lib/membrane/core/element/event_controller.ex
```diff
@@ -50,6 +54,20 @@ defmodule Membrane.Core.Element.EventController do
 
   @spec do_exec_handle_event(Pad.ref_t(), Event.t(), params :: map, State.t()) ::
           State.stateful_try_t()
+  defp do_exec_handle_event(pad_ref, event, params, state)
+       when event in [%Event.StartOfStream{}, %Event.EndOfStream{}] do
```
mat-hek: better do `do_exec_handle_event(pad_ref, %event{}, ...) when event in [Event.StartOfStream, Event.EndOfStream] ...`

lib/membrane/element/action.ex
```diff
@@ -157,6 +157,14 @@ defmodule Membrane.Element.Action do
   """
   @type playback_change_t :: {:playback_change, :suspend | :resume}
 
+  @typedoc """
+  Sends EndOfStream event through a pad (output) that triggers
+  callback `end_of_stream/3` at the receiver element.
+
+  Forbidden when playback state is stopped.
```
mat-hek: I would allow that only if playing

lib/membrane/element/base.ex
```diff
@@ -189,6 +189,25 @@ defmodule Membrane.Element.Base do
   """
   @callback handle_shutdown(state :: Element.state_t()) :: :ok
 
+  @doc """
+  Callback invoked when element receives `Membrane.Event.StartOfStream` event.
+  """
+  @callback handle_start_of_stream(
+              pad :: Pad.ref_t(),
+              context :: CallbackContext.Event.t(),
```
mat-hek: I would create a separate context for this as differences may occur in the future

lib/membrane/element/base.ex
```diff
@@ -239,15 +258,16 @@ defmodule Membrane.Element.Base do
       def handle_pad_removed(_pad, _context, state), do: {:ok, state}
 
       @impl true
-      def handle_event(pad, %Event.StartOfStream{}, _context, state),
+      def handle_event(_pad, _event, _context, state), do: {:ok, state}
+
+      @impl true
+      def handle_start_of_stream(pad, _context, state),
         do: {{:ok, notify: {:start_of_stream, pad}}, state}
```
mat-hek: We should inform the pipeline under the hood, not allowing to override this. Also, the pipeline should have separate callbacks like `handle_element_*_of_stream` (remember to update assertions in Membrane.Testing).

lib/membrane/element/base.ex
```diff
@@ -239,15 +258,16 @@ defmodule Membrane.Element.Base do
...
 
       @impl true
-      def handle_event(pad, %Event.EndOfStream{}, _context, state),
+      def handle_end_of_stream(pad, _context, state),
         do: {{:ok, notify: {:end_of_stream, pad}}, state}
```
mat-hek: - Same as for the start of stream.
- In filters, this should be passed to all outputs by default.
- In the future, we should add an option to defer notifying about the end of stream (which you can make an issue of).

test/membrane/element/element_test.exs
kanes115: I know that, but in this particular case I would like to be agnostic when it comes to what pipeline module has been used as I get pipeline pid from setup? I mean now if for example testing piepeline  name changed, we would only need to change one place in test code instead of two

test/membrane/element/element_test.exs
kanes115: Well, start_of_stream does not require any payload to be sent in the test while end of stream does (?). This is like a different set of test pipeline and elements. Does it make sense?

lib/membrane/element/base.ex
kanes115: I must say I was a little bit following the rules that were followed when writing the rest of this file's documentation. Anyway I think that we can assume that `element` here is a general idea which would not require an article :)

lib/membrane/element/base.ex
kanes115: Same as above

test/membrane/element/element_test.exs
kanes115: Same as above

test/membrane/element/element_test.exs
kanes115: Same as above

test/membrane/element/element_test.exs
kanes115: Same as above

test/membrane/element/element_test.exs
Hajto: In this case payloads are sent. If you don't provide list of payloads, `Testing.Source` will output buffers based on default buffer generator.

test/membrane/element/element_test.exs
Hajto: In this case payloads are sent. If you don't provide list of payloads, Testing.Source will output buffers based on default buffer generator. So buffers will be sent regardless.

test/membrane/element/element_test.exs
Hajto: You are right, but if you were to use a different pipeline that overridden `play/1` you would want to use that pipeline's implementation anyways. I will leave that up to your conscience, because both options have it's pros. Choose whichever you think is best.

Hajto: And if the name of the pipeline changed you would have to change name in one place, because then it you would alias `Membrane.Testing.Pipeline`.

Hajto:
Also please try `mix compile --force` to check the warnings and fix those if possible.

lib/membrane/element/base.ex
kanes115: Ad2 - what is `this` in your sentence?
Ad3 - yes, commits from 88e8828605a88671b98d9553dd80456d234e6ee1 to b495cbed99a221906c25d2dbe89179ebde814fc3 handle this.

test/membrane/element/element_test.exs
kanes115: Fair point, let's switch to the specific module

kanes115:
Rebased with master

lib/membrane/element/base.ex
mat-hek: `this` is the end of stream event

kanes115: Is [this](eddfa8ad98a61f62e7477b588e20c9090f7bcaf2) what you meant?

lib/membrane/core/element/action_handler.ex
```diff
@@ -391,4 +401,7 @@ defmodule Membrane.Core.Element.ActionHandler do
     Message.send(watcher, :notification, [name, notification])
     {:ok, state}
   end
+
+  defp opposite_direction(:input), do: :output
```
mat-hek: Can be moved to Membrane.Pad

lib/membrane/core/element/action_handler.ex
```diff
@@ -90,13 +90,23 @@ defmodule Membrane.Core.Element.ActionHandler do
...
-       when cb in [:handle_caps, :handle_event, :handle_process_list] do
+       when cb in [
+              :handle_caps,
+              :handle_event,
+              :handle_process_list,
+              :handle_end_of_stream,
+              :handle_start_of_stream
+            ] do
     {action, dir} =
       case {cb, params} do
```
mat-hek: Matching on params is now unneeded

mat-hek:
I have just realized that the start/end of stream callbacks make no sense when there are no input pads, so they can be moved to `Element.WithInputPads`

kanes115:
> I have just realized that the start/end of stream callbacks make no sense when there are no input pads, so they can be moved to `Element.WithInputPads`

True. Have changed that. (b01a1e36de756cb0b2e1ba94fe6964515d03cf04)

PR #175, opened 2019-07-29
Bins

kanes115:


lib/membrane/core/parent_action.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Membrane.Core.ParentAction do
+  @moduledoc """
+  This module consists of common for bin and pipeline types and functions connected to actions.
```
Hajto: common what? Common functions? callbacks? handlers?

Maybe you meant `This module consists of common types and functions connected to actions for bin and pipeline .`

lib/membrane/core/parent_action.ex
```diff
@@ -0,0 +1,67 @@
...
+  alias Membrane.Core.Message
+  alias Membrane.Notification
+  alias Membrane.Pipeline
+  alias Membrane.Bin
+  alias Membrane.Core.ParentUtils
+
+  use Bunch
+
+  @typedoc """
+  Action that sends a message to element identified by name.
```
Hajto: **a** name or **the** name?

lib/membrane/core/parent_utils.ex
```diff
@@ -0,0 +1,131 @@
+defmodule Membrane.Core.ParentUtils do
+  use Bunch
+
+  alias Membrane.Element
```
Hajto: This could be ordered alphabetically and put into single alias `Membrane.{A,B,C}`

lib/membrane/core/parent_utils.ex
```diff
@@ -0,0 +1,131 @@
...
+  alias Membrane.Bin
+  alias Membrane.ParentError
+  alias Membrane.Core.ParentState
```
Hajto: Same as above.

lib/membrane/core/parent_utils.ex
```diff
@@ -0,0 +1,131 @@
+defmodule Membrane.Core.ParentUtils do
+  use Bunch
```
Hajto: we usually keep uses together. Personally i keep try sticking to elixir style guide and stick to this order:

```
@moduledoc
@behaviour
use
import
alias
require
@module_attribute
defstruct
@type
@callback
@macrocallback
@optional_callbacks
```

But that is not enforced by any means. Do what you please and think is right.

lib/membrane/core/parent_utils.ex
```diff
@@ -0,0 +1,131 @@
...
+  @spec check_if_children_names_unique([parsed_child_t], Bin.State.t() | Pipeline.State.t()) ::
+          Type.try_t()
+  def check_if_children_names_unique(children, state) do
+    %{children: state_children} = state
+
+    children
+    |> Enum.map(& &1.name)
+    |> Kernel.++(Map.keys(state_children))
+    |> Bunch.Enum.duplicates()
+    ~> (
```
Hajto: We usually use elixir syntax wherever it is possible. Here it can be replaced by 
```
|> case do
  [] -> :hurray
  dups -> :sad
end
```

lib/membrane/pipeline.ex
```diff
@@ -24,44 +27,68 @@ defmodule Membrane.Pipeline do
...
-  @type pipeline_options_t :: any
+  defmodule Spec do
+    @moduledoc """
+    Structure representing topology of a pipeline.
 
-  @typedoc """
-  Action that sends a message to element identified by name.
-  """
-  @type forward_action_t :: {:forward, {Element.name_t(), Notification.t()}}
+    It can be returned from
```
Hajto: Maybe from **the** {name} callback?

lib/membrane/pipeline.ex
```diff
@@ -24,44 +27,68 @@ defmodule Membrane.Pipeline do
...
-  """
-  @type spec_action_t :: {:spec, Spec.t()}
+    Children that should be spawned when the pipeline starts can be defined
+    with the `:children` field.
 
-  @typedoc """
-  Action that stops, unlinks and removes specified child/children from pipeline.
-  """
-  @type remove_child_action_t :: {:remove_child, Element.name_t() | [Element.name_t()]}
+    You have to set it to a keyword list, where keys are valid element names (`t:Membrane.ParentUtils.chidl_name_t/0`)
```
Hajto: link to type does not work, maybe you meant child not child**l**

lib/membrane/pipeline.ex
```diff
@@ -24,44 +27,68 @@ defmodule Membrane.Pipeline do
...
+    that are unique within this pipeline and values are either element's module or
+    struct of that module.
 
-  @typedoc """
-  Type describing actions that can be returned from pipeline callbacks.
+    Sample definitions:
+
+    [
```
Hajto: Please indend and surround by tripple ticks to format it properly.

lib/membrane/pipeline.ex
```diff
@@ -24,44 +27,68 @@ defmodule Membrane.Pipeline do
...
+    keyword list at the end of a tuple (See `t:endpoint_options_t/0`).
+    Children names have to match names given to the `:children` field.
+
+    Once it's done, pipeline will ensure that links are present.
+
+    Sample definition:
+
+    ```
+    %{
+    {:source_a, :output} => {:converter, :input, buffer: [preferred_size: 20_000]},
```
Hajto: please indent map contents one level

lib/membrane/pipeline.ex
```diff
@@ -24,44 +27,68 @@ defmodule Membrane.Pipeline do
...
+    can be defined with the `:links` field.
 
-  Returning actions is a way of pipeline interaction with its elements and
-  other parts of framework.
+    You have to set it to a map, where both keys and values are tuples of
+    `{child_name, pad_name}`. Entries can also have additional options passed by
+    keyword list at the end of a tuple (See `t:endpoint_options_t/0`).
+    Children names have to match names given to the `:children` field.
+
+    Once it's done, pipeline will ensure that links are present.
```
Hajto: that these links are present

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  alias Membrane.Element
+  alias Membrane.Core.{PadModel, PadsSpecs}
+  alias Membrane.Pipeline.Link
+  alias Membrane.Bin
+  alias Membrane.Core.Bin.{State, LinkingBuffer}
+  alias Membrane.{CallbackError, Core, Element, Notification, BinError}
+  alias Core.Pad
+  alias Core.{Message, Playback}
+  alias Bunch.Type
+  alias Membrane.Core.PadController
```
Hajto: All `Membrane.Core.{PadSpecHandler, ...}` can be bunched together.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+    For bins boundries there are special links allowed. User should define links
+    between bin's input and first child's input (input-input type) and first
+    child's output and bin output (output-output type). In this case, callback module
+    creator should name endpoint of the bin with macro `this_bin()`
+
+    Sample definition:
+
+    ```
+    %{
+    {this_bin(), :input} => {:filter1, :input, buffer: [preferred_size: 10]},
```
Hajto: indent this one level?

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  @type bin_options_t :: any
+
+  @typedoc """
+  Type that defines all valid return values from most callbacks.
+  """
+  @type callback_return_t ::
+          CallbackHandler.callback_return_t(ParentAction.t(), State.internal_state_t())
+
+  @doc """
+  Enables to check whether module is membrane bin
```
Hajto: missing dot at the end.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  """
+  @callback membrane_bin? :: true
+
+  @doc """
+  Callback invoked on initialization of bin process. It should parse options
```
Hajto: initialization of **the** bin? **and ** bin?

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  and initialize element internal state. Internally it is invoked inside
```
Hajto: missing possesive s. element **'s**

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  `c:GenServer.init/1` callback.
+  """
+  @callback handle_init(options :: bin_options_t) ::
+              {{:ok, Spec.t()}, State.internal_state_t()}
+              | {:error, any}
+
+  @doc """
+  Callback invoked when bin transition from `:stopped` to `:prepared` state has finished,
+  that is all of its elements are prepared to enter `:playing` state.
```
Hajto: it **'** s

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  """
+  @callback handle_stopped_to_prepared(state :: State.internal_state_t()) :: callback_return_t
+
+  @doc """
+  Callback invoked when bin transition from `:playing` to `:prepared` state has finished,
+  that is all of its elements are prepared to be stopped.
```
Hajto: same as above

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  """
+  @callback handle_playing_to_prepared(state :: State.internal_state_t()) :: callback_return_t
+
+  @doc """
+  Callback invoked when bin is in `:playing` state, i.e. all its elements
```
Hajto: same as above.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  are in this state.
+  """
+  @callback handle_prepared_to_playing(state :: State.internal_state_t()) :: callback_return_t
+
+  @doc """
+  Callback invoked when bin is in `:playing` state, i.e. all its elements
```
Hajto: same as above.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+
+  Useful for receiving ticks from timer, data sent from NIFs or other stuff.
+  """
+  @callback handle_other(message :: any, state :: State.internal_state_t()) :: callback_return_t
+
+  @doc """
+  Callback invoked when `Membrane.Bin.Spec` is linked and in the same playback
+  state as bin.
+
+  Spec can be started from `c:handle_init/1` callback or as `t:Membrane.Core.ParentAction.spec_action_t/0`
```
Hajto: This line seems to be too long.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+              state :: State.internal_state_t()
+            ) ::
+              callback_return_t
+
+  defguard is_bin_name(term)
+           when is_atom(term) or
+                  (is_tuple(term) and tuple_size(term) == 2 and is_atom(elem(term, 0)) and
+                     is_integer(elem(term, 1)) and elem(term, 1) >= 0)
+
+  defmacro this_bin_marker do
```
Hajto: Maybe few words about this macro?

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+
+  @doc PadsSpecs.def_bin_pad_docs(:input)
+  defmacro def_input_pad(name, spec) do
+    Pad.assert_public_name!(name)
+
+    input_spec = spec
+    output_spec = Keyword.drop(spec, @private_input_pad_spec_keys)
+
+    input = PadsSpecs.def_pad(name, :input, input_spec)
+
```
Hajto: This newline seems odd. Few lines lower you have almost same structure and it does not have new line between input and output.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+      if Module.get_attribute(__MODULE__, :bin_pads_pairs) == nil do
+        Module.register_attribute(__MODULE__, :bin_pads_pairs, accumulate: true)
+        @before_compile {unquote(__MODULE__), :generate_bin_pairs}
+      end
+
+      @bin_pads_pairs {{:private, unquote(name)}, unquote(name)}
+    end
+  end
+
+  defmacro generate_bin_pairs(env) do
```
Hajto: Maybe few words about this macro?

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  defp parse_links(links), do: links |> Bunch.Enum.try_map(&Link.parse/1)
+
+  @spec resolve_links([Link.t()], State.t()) :: [Link.resolved_t()]
+  defp resolve_links(links, state) do
+    {new_links, new_state} =
+      links
+      |> Enum.reduce(
+        {[], state},
+        fn link, {links_acc, state_acc} ->
+          {from, s1} = link.from |> resolve_link(state_acc)
```
Hajto: what does `s` stand for?

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+        %State{playback: %Playback{state: current_playback_state}, pending_pids: pending_pids} =
+          state
+      ) do
+    new_pending_pids = pending_pids |> MapSet.delete(pid)
+    new_state = %{state | pending_pids: new_pending_pids}
+
+    if new_pending_pids != pending_pids and new_pending_pids |> Enum.empty?() do
+      callback = PlaybackHandler.state_change_callback(current_playback_state, new_playback_state)
+
+      with {:ok, new_state} <-
```
Hajto: This looks like `~>>` use case.

lib/membrane/bin.ex
mat-hek: `its` is the correct syntax

Hajto: Mea maxima culpa, http://its-not-its.info/

lib/membrane/core/bin/linking_buffer.ex
```diff
@@ -0,0 +1,59 @@
...
+      {:error, {:unknown_pad, _}} ->
+        Map.put(buf, outgoing_pad, msg)
+    end
+  end
+
+  @doc """
+  Sends messages stored for a given outpud pad.
+  A link must already be available.
+  """
+  @spec eval_for_pad(t(), Pad.ref_t(), Bin.State.t()) :: t()
```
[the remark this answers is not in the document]
kanes115: I was following the convention in `PlaybackBuffer`

lib/membrane/core/bin/linking_buffer.ex
```diff
@@ -0,0 +1,59 @@
...
+        buf
+
+      {msg, new_buf} ->
+        do_eval(msg, pad, bin_state)
+        new_buf
+    end
+  end
+
+  defp do_eval(msg, outgoing_pad, bin_state) do
+    {:ok, %{pid: dest_pid, other_ref: other_ref}} = PadModel.get_data(bin_state, outgoing_pad)
```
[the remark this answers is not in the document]
kanes115: You mean the key in a map returned by PadModel? Why? Even if we agree on this, the change is out of the scope of this PR, I guess.

lib/membrane/core/parent_action.ex
kanes115: https://context.reverso.net/t%C5%82umaczenie/angielski-polski/identified+by+name
Seems okey without an article.

lib/membrane/pipeline.ex
kanes115: We do not normally use the definite article with names (unless it is one of the exceptions).

lib/membrane/pipeline.ex
kanes115: Why is it necessary?

lib/membrane/bin.ex
kanes115: It is analogous to docs of `c:Pipeline.handle_init/1 `.

lib/membrane/bin.ex
kanes115: Oh, it actually should be private.

lib/membrane/core/parent_action.ex
kanes115: I think the original sentence is correct. "Module consists of common " ++ "for bin and pipeline" (between what entities they are common) ++ "types and functions" (what is actually shared) ++ "connected to actions" (description of those shared entities).

And it rhymes, so it's cool :D

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+        state
+      end
+
+    {{:ok, :cont}, state}
+  end
+
+  defp handle_link(pid, args, state) do
+    case self() do
+      ^pid ->
+        {{:ok, _spec} = res, state} = apply(PadController, :handle_link, args ++ [state])
```
mat-hek: - use with
- passing args this way smells a bit, if there are too many args they should be passed as a struct/well-known map

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+        ~> %{state | linking_buffer: &1}
+      else
+        state
+      end
+
+    {{:ok, :cont}, state}
+  end
+
+  defp handle_link(pid, args, state) do
+    case self() do
```
mat-hek: you can check it in a guard, making separate clauses

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+      :stopped ->
+        {:stop, :normal, state}
+
+      _ ->
+        state = %{state | terminating?: true}
+
+        PlaybackHandler.change_and_lock_playback_state(:stopped, __MODULE__, state)
+        |> noreply(state)
+    end
+  end
```
mat-hek: ~100 lines above look like copied from the pipeline

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+  @doc """
+  Checks whether module is a bin.
+  """
+  @spec bin?(module) :: boolean
+  def bin?(module) do
+    module |> Bunch.Module.check_behaviour(:membrane_bin?)
+  end
+
+  @spec handle_spec(Spec.t(), State.t()) :: Type.stateful_try_t([Element.name_t()], State.t())
+  defp handle_spec(%Spec{children: children_spec, links: links}, state) do
```
mat-hek: separate this and helpers to something like `Bin.SpecController`

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+    PadController.handle_linking_finished(state)
+    |> reply()
+  end
+
+  def handle_call(Message.new(:set_watcher, watcher), _, state) do
+    {:ok, %{state | watcher: watcher}} |> reply()
+  end
+
+  @impl CallbackHandler
+  def handle_action({:forward, {elementname, message}}, _cb, _params, state) do
```
mat-hek: - separate to `Bin.ActionHandler`
- add @impl for all clauses

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+  use Membrane.Core.CallbackHandler
+  use GenServer
+  use Membrane.Core.PlaybackHandler
+  use Membrane.Core.PlaybackRequestor
+
+  import Membrane.Helper.GenServer
+
+  alias Membrane.Element
+
+  alias Membrane.Core.{
```
mat-hek: I would move all the parent-related stuff under `Core.Parent.*` and all the child-related stuff under `Core.Child.*`

lib/membrane/bin.ex
```diff
@@ -0,0 +1,733 @@
...
+  alias Membrane.Pipeline.Link
+  alias Membrane.Bin
+  alias Membrane.Core.Bin.{State, LinkingBuffer}
+  alias Membrane.{CallbackError, Core, Element, Notification, BinError}
+  alias Core.Pad
+  alias Core.{Message, Playback}
+  alias Bunch.Type
+  alias Membrane.Core.PadController
+  alias Membrane.Core.PadSpecHandler
+  alias Membrane.Core.ParentUtils
```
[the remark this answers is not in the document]
kanes115: Yea, I have it in TODO list: "Move Parent* modules to Parent.* namespace". I think so too

lib/membrane/bin.ex
```diff
@@ -0,0 +1,745 @@
...
+          module,
+          bin_options :: bin_options_t,
+          process_options :: GenServer.options()
+        ) :: GenServer.on_start()
+  def start_link(my_name, module, bin_options \\ nil, process_options \\ []),
+    do: do_start(:start_link, my_name, module, bin_options, process_options)
+
+  defp do_start(method, my_name, module, bin_options, process_options) do
+    with :ok <-
+           (if module |> bin? do
```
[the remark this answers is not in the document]
kanes115: I agree, I was just following the convention from pipeline.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,745 @@
...
+              :ok
+            else
+              :not_bin
+            end) do
+      apply(GenServer, method, [__MODULE__, {my_name, module, bin_options}, process_options])
+    else
+      :not_bin ->
+        {:not_bin, module}
```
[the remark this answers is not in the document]
kanes115: In pipeline there is only a warning printed.

lib/membrane/core/parent_state.ex
```diff
@@ -0,0 +1,56 @@
+defprotocol Membrane.Core.ParentState do
```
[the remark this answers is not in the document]
kanes115: How is that better? This is a very similar approach as for `Plabackable`.

lib/membrane/bin.ex
kanes115: Ad 2) `handle_link` is like `maybe_*` functions. It only calls a function with given arguments if predicate is true. In this case, I think, this is justified.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,746 @@
...
+
+    {new_links, new_state}
+  end
+
+  defp resolve_link(
+         %{element: this_bin_marker(), pad_name: pad_name, id: id} = endpoint,
+         %{module: bin_mod, name: name} = state
+       ) do
+    with {:ok, private_bin} <- bin_mod.get_corresponding_private_pad(pad_name),
+         {{:ok, pad_ref}, state} <- PadController.get_pad_ref(private_bin, id, state) do
```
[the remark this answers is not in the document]
kanes115: resolving as this is outdated (we don't use `module.get_corresponding_private_pad/1`)

lib/membrane/bin.ex
kanes115: What would you put under Core.Child.*?

kanes115: `ChildrenController` should go to `Parent.*` imo

lib/membrane/core/bin/spec_controller.ex
```diff
@@ -0,0 +1,154 @@
...
+
+  defp handle_link(pid, args, state) when pid == self() do
+    with {{:ok, _spec} = res, state} <- apply(PadController, :handle_link, args ++ [state]),
+         {:ok, state} <- PadController.handle_linking_finished(state),
+         do: {res, state}
+  end
+
+  defp handle_link(pid, args, state) do
+    Message.call(pid, :handle_link, args)
+    ~> {&1, state}
```
Hajto: We had long discussion about this pattern. The problem is that it does reduces performance (`~>` compiles into case expression) without any positive effects. In this particular case I don't think it will make too much difference, but I wanted to bring this fact to your attention.

In the long run I think `~>` should be optimized for such use cases.

lib/membrane/core/bin/state.ex
```diff
@@ -0,0 +1,54 @@
...
+          module: module,
+          children: children_t,
+          pending_pids: MapSet.t(pid),
+          terminating?: boolean,
+          name: Bin.name_t(),
+          bin_options: any,
+          pads: PadModel.pads_t() | nil,
+          watcher: pid,
+          controlling_pid: pid,
+          linking_buffer: LinkingBuffer.t()
```
Hajto: This field is initialized with value nil, should this be `LinkingBuffer.t() | nil` then?

Hajto: Also could this type definition have same order of fields as `defstruct`?

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,170 @@
+defmodule Membrane.Core.ChildrenController do
```
Hajto: I wonder wether this should be part of public api. I would personally mark this as `@moduledoc false`

lib/membrane/core/bin/spec_controller.ex
```diff
@@ -0,0 +1,154 @@
+defmodule Membrane.Core.Bin.SpecController do
```
Hajto: I wonder wether this should be part of public api. I would personally mark this as `@moduledoc false`

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,170 @@
...
+  use Bunch
+  use Membrane.Log, tags: :core
+  use Membrane.Core.PlaybackRequestor
+
+  alias Membrane.{Bin, Element, ParentError, Spec}
+  alias Membrane.Core.{Message, Parent}
+  alias Membrane.Core.ChildrenController
```
Hajto: This could be moved to line 7.

Hajto: `alias Membrane.Core.{ChildrenController, Message, Parent}`

lib/membrane/core/element/state.ex
```diff
@@ -6,9 +6,10 @@ defmodule Membrane.Core.Element.State do
 
   use Membrane.Log, tags: :core
   alias Membrane.{Core, Element}
-  alias Core.{Playback, Playbackable}
-  alias Core.Element.{PadModel, PadSpecHandler, PlaybackBuffer}
-  alias Element.Pad
+  alias Core.PadSpecHandler
+  alias Core.{Playback, Playbackable, PadModel}
```
Hajto:
```
alias Core.{PadModel, PadSpecHandler , Playback, Playbackable}
```

lib/membrane/core/pad.ex
```diff
@@ -143,4 +157,27 @@ defmodule Membrane.Element.Pad do
   """
   def name_by_ref({:dynamic, name, _id}) when is_pad_name(name), do: name
```
Hajto: If public function has multiple clauses it is nice touch to provide bodyless clause:
```
def f1(very, important)
def f1(:d, :p)
def f1(:u, :a)
```

Because then `ex_doc` will place argument names from that def without body in docs instead of generated names like `f1(arg1, arg2)`.

This is not really needed, but I think it is nice touch.

lib/membrane/core/pad.ex
```diff
@@ -143,4 +157,27 @@ defmodule Membrane.Element.Pad do
...
   def name_by_ref(ref) when is_pad_name(ref), do: ref
+
+  def get_corresponding_bin_pad({:dynamic, name, id}),
```
Hajto: Same as above, also optional.

lib/membrane/core/pad_data.ex
```diff
@@ -55,6 +56,7 @@ defmodule Membrane.Element.Pad.Data do
     :sticky_messages,
     :input_buf,
     :demand,
-    :options
+    :options,
```
Hajto: If think this makes sense to intilize with default empty list.

lib/membrane/core/pad_data.ex
```diff
@@ -55,6 +56,7 @@ defmodule Membrane.Element.Pad.Data do
...
+    :bin?
```
Hajto: I would prefer if this had default value that is boolean, otherwise typespec should be boolean or nil.

lib/membrane/core/pad_model.ex
```diff
@@ -28,7 +30,8 @@ defmodule Membrane.Core.Element.PadModel do
 
   @type unknown_pad_error_t :: {:error, {:unknown_pad, Pad.name_t()}}
 
-  @spec assert_instance(State.t(), Pad.ref_t()) :: :ok | unknown_pad_error_t
+  @spec assert_instance(Bin.State.t() | Element.State.t(), Pad.ref_t()) ::
```
Hajto: Bin or Element state seems to be very common in this module I would consider defining type alias (or some macro expansion shenanigans). (optional)

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Membrane.Core.Parent.Action do
+  @moduledoc """
+  This module consists of common for bin and pipeline types and functions connected to actions.
+  """
+  alias Membrane.CallbackError
+  alias Membrane.Core.Parent
+  alias Membrane.Core.Message
+  alias Membrane.Notification
```
Hajto: `alias Membrane.{Bin, Notification, Pipeline}`

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Membrane.Core.Parent.Action do
+  @moduledoc """
+  This module consists of common for bin and pipeline types and functions connected to actions.
+  """
+  alias Membrane.CallbackError
+  alias Membrane.Core.Parent
```
Hajto: ` alias Membrane.Core.{Message,Parent}`

lib/membrane/core/pipeline/state.ex
```diff
@@ -0,0 +1,36 @@
...
+  alias Membrane.Core.{Playback, Playbackable}
+  alias Membrane.Element
+  use Bunch
+
+  @derive Playbackable
+
+  @type t :: %__MODULE__{
+          internal_state: internal_state_t | nil,
+          playback: Playback.t(),
+          module: module,
```
Hajto: Theoretically or nil

test/membrane/core/bin_test.exs
```diff
@@ -0,0 +1,422 @@
+defmodule Membrane.Core.BinTest do
```
Hajto: This module contains a lot of helper elements and bins. I would consider putting some of more general entities to appropriate support folder to declutter this test module.

test/membrane/core/bin_test.exs
```diff
@@ -0,0 +1,422 @@
...
+    def handle_prepared_to_playing(_ctx, state), do: {:ok, state}
+
+    @impl true
+    def handle_demand(:output, size, _, _ctx, state), do: {{:ok, demand: {:input, size}}, state}
+
+    @impl true
+    def handle_process(_pad, buf, _, state), do: {{:ok, buffer: {:output, buf}}, state}
+
+    @impl true
+    def handle_event(_pad, event, _ctx, state), do: {{:ok, forward: event}, state}
```
Hajto: why is this needed? Isn't forward default action?

test/membrane/core/bin_test.exs
```diff
@@ -0,0 +1,422 @@
...
+
+    @impl true
+    def handle_init(opts) do
+      children = [
+        filter1: opts.filter1,
+        filter2: opts.filter2
+      ]
+
+      links = %{
+        {this_bin(), :input} => {:filter1, :input, buffer: [preferred_size: 10]},
```
Hajto: Why are buffer sizes preset to 10? Is this important for tests?

test/membrane/core/bin_test.exs
```diff
@@ -0,0 +1,422 @@
...
+    def handle_process(_pad, buf, _, state), do: {{:ok, buffer: {:output, buf}}, state}
+
+    @impl true
+    def handle_event(_pad, event, _ctx, state), do: {{:ok, forward: event}, state}
+
+    @impl true
+    def handle_shutdown(_reason, _state), do: :ok
+  end
+
+  describe "Starting and transmitting buffers" do
```
Hajto: This and subsequent test are very similar, the differentiating factor is pipeline configuration. Maybe test procedure could be generalized?

lib/membrane/core/bin/spec_controller.ex
```diff
@@ -0,0 +1,154 @@
...
+          new_link = %{link | from: from, to: to}
+          {new_link, state_acc}
+        end
+      )
+
+    {new_links, new_state}
+  end
+
+  defp resolve_link(
+         %{element: Bin.this_bin_marker(), pad_name: pad_name, id: id} = endpoint,
```
mat-hek: `Bin.this_bin_marker` returns `{Membrane.Core.Bin.SpecController, :this_bin}` in this case. Not sure how it can work in this shape

lib/membrane/bin.ex
```diff
@@ -641,44 +295,13 @@ defmodule Membrane.Bin do
...
-    Message.send(watcher, :notification, [name, notification])
-    {:ok, state}
-  end
-
   def set_controlling_pid(server, controlling_pid, timeout \\ 5000) do
     Message.call(server, :set_controlling_pid, controlling_pid, [], timeout)
   end
 
   defmacro __using__(_) do
+    marker = this_bin_marker()
```
mat-hek: `marker` will be always equal to `{Membrane.Bin, :this_bin}`

lib/membrane/core/bin/spec_controller.ex
```diff
@@ -0,0 +1,154 @@
...
+           |> Parent.State.get_children()
+           |> Bunch.Enum.try_each(fn {_pid, pid} -> pid |> Element.handle_linking_finished() end),
+         do: {:ok, state}
+  end
+
+  defp link(
+         %Link{from: %Link.Endpoint{pid: from_pid} = from, to: %Link.Endpoint{pid: to_pid} = to},
+         state
+       ) do
+    from_args = [
```
mat-hek: I remember I already requested a fix here, but it might've been marked as resolved because you separated this to another file.

lib/membrane/core/pad.ex
```diff
@@ -154,14 +158,26 @@ defmodule Membrane.Core.Pad do
...
-      raise CompileError,
-        file: __ENV__.file,
-        description: "#{inspect(name)} is not a proper pad name. Use public names only."
-    end
+  def get_corresponding_bin_pad({:dynamic, name, id}),
+    do: {:dynamic, get_corresponding_bin_name(name), id}
+
+  def get_corresponding_bin_pad(name), do: get_corresponding_bin_name(name)
+
+  def create_private_name(name) do
```
mat-hek: This shouldn't be public, as the name suggests

lib/membrane/core/bin/spec_controller.ex
mat-hek: `it does reduces performance (~> compiles into case expression)` <- that is not true. It may only slightly extend compilation time. See https://github.com/membraneframework/bunch/blob/master/lib/bunch.ex#L248

lib/membrane/core/bin/spec_controller.ex
kanes115: No, it returns `{Membrane.Bin, :this_bin}`

lib/membrane/bin.ex
kanes115: Yes, it will. The idea here is that there is a function available in the callback module `this_bin/0` that will mark endpoint as belonging to bin's API. We also need a way to to recognise that a particular endpoint belongs to bin api. We do it by calling macro `Bin.this_bin_marker()` (from [here](https://github.com/membraneframework/membrane-core/pull/175/files#diff-66ae7e58d645656945aedbf3a8c5cb10R43))

lib/membrane/core/bin/spec_controller.ex
kanes115: I've responded to that: https://github.com/membraneframework/membrane-core/pull/175#discussion_r313308564

lib/membrane/core/pad.ex
kanes115: There are two  mechanisms that we might mix up. There is a concept of `private pads` which are basically pads that don't have documentation and are used by Membrane somehow (not the user) and there is a concept of `bin pads` which are descriptions of pads that will lead to creation of two pads in the end (public and private one that are strictly connected to each other).
I believe creating private pads should be possible by Membrane from other modules than `Pad*` modules, for example by modules that deal with `bin pads` (as it's the case here). 
Thus, privacy of private pads means nothing more that they are not to be managed nor seen by the user.
Does it make sense, @mat-hek? Do I get you correctly?

kanes115: The thing I might consider useful is to put [this](https://github.com/membraneframework/membrane-core/pull/175/files#diff-b4462fc7a237adf6fe299ac5fbe2661fR56) piece of code somewhere in `Bin` namespace so that is clear these procedures are bin specific.

lib/membrane/bin.ex
```diff
@@ -693,10 +316,10 @@ defmodule Membrane.Bin do
       @impl true
       def membrane_bin?, do: true
 
-      defp this_bin, do: unquote(__MODULE__).this_bin_marker()
+      defp this_bin, do: unquote(marker)
```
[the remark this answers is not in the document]
kanes115: I feel strong preference towards `this_bin()`? Isn't this more intuitive? Like `this` from OOP?

lib/membrane/core/pad.ex
kanes115: Well, in this case ex_doc is capable of deriving the correct parameter name.

lib/membrane/core/pad.ex
kanes115: https://github.com/membraneframework/membrane-core/pull/175#discussion_r313755361

lib/membrane/core/bin/spec_controller.ex
Hajto: Then this must have changed since we last tested it.

test/membrane/core/bin_test.exs
kanes115: I can see that, but on the other hand those modules have only `handle_init` callback and at most two pad definitions. As they are small maybe it is better to have them in the same file? Otherwise, how do suggest to declutter this? Put every test bin module into a seperate file? All of them into one different file?

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Membrane.Core.Parent.Action do
+  @moduledoc """
+  This module consists of common for bin and pipeline types and functions connected to actions.
```
mat-hek: Membrane.Core.* modules shouldn't be public. Consider creating Membrane.* counterparts for docs, types and public functions

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,63 @@
...
+  alias Membrane.{CallbackError, Notification}
+  alias Membrane.Core.{Parent, Message}
+  alias Parent.ChildrenController
+
+  use Bunch
+
+  @typedoc """
+  Action that sends a message to element identified by name.
+  """
+  @type forward_action_t :: {:forward, {ChildrenController.child_name_t(), Notification.t()}}
```
mat-hek: ChildController is private, so type from there shouldn't be used here

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,63 @@
...
+  Children's playback state is changed to the current parent state.
+  `c:handle_spec_started` callback is executed once it happens.
+  """
+  @type spec_action_t :: {:spec, Spec.t()}
+
+  @typedoc """
+  Action that stops, unlinks and removes specified child/children from their parent.
+  """
+  @type remove_child_action_t ::
+          {:remove_child, ChildrenController.child_name_t() | [ChildrenController.child_name_t()]}
```
mat-hek: as above

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,63 @@
...
+  """
+  @type forward_action_t :: {:forward, {ChildrenController.child_name_t(), Notification.t()}}
+
+  @typedoc """
+  Action that instantiates elements and links them according to `Membrane.Spec`.
+
+  Children's playback state is changed to the current parent state.
+  `c:handle_spec_started` callback is executed once it happens.
+  """
+  @type spec_action_t :: {:spec, Spec.t()}
```
mat-hek: Spec is unaliased. Dialyzer should catch such errors

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Membrane.Core.Parent.MessageDispatcher do
```
mat-hek: I wouldn't call this message dispatcher, as this doesn't dispatch messages. Also, message dispatcher in element handles all the messages and dispatches them (and I feel it's a good way to go), while this is passed some messages only. I would call this LifecycleController and replace `handle_message` clauses with proper functions, leaving message handling to the Bin and the Pipeline

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -0,0 +1,143 @@
...
+        {error, state}
+    end
+  end
+
+  def handle_message(Message.new(:shutdown_ready, child), state, _) do
+    {{:ok, pid}, state} = Parent.State.pop_child(state, child)
+    {Element.shutdown(pid), state}
+  end
+
+  def handle_message(Message.new(:demand_unit, [demand_unit, pad_ref]), state, _) do
```
mat-hek: This is not related to parent

lib/membrane/core/parent/state.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Membrane.Core.Parent.State do
```
mat-hek: -> ChildrenModel

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,169 @@
...
+  require Message
+
+  @type child_name_t :: Element.name_t() | Bin.name_t()
+  @typep parsed_child_t :: %{name: child_name_t(), module: module, options: Keyword.t()}
+
+  @callback resolve_links([Link.t()], Parent.State.t()) :: [Link.resolved_t()]
+
+  @callback link_children([Link.resolved_t()], Parent.State.t()) :: Type.try_t()
+
+  @callback exec_handle_spec_started([child_name_t()], Parent.State.t()) ::
```
mat-hek: I'd opt for `action_handler_module` callback instead of this

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,169 @@
...
+
+  defp child_type(module) do
+    if module |> Bunch.Module.check_behaviour(:membrane_bin?) do
+      :bin
+    else
+      :element
+    end
+  end
+
+  # Recursion that starts children processes, case when both module and options
```
mat-hek: remove

lib/membrane/core/pad_controller.ex
```diff
@@ -184,8 +184,12 @@ defmodule Membrane.Core.Element.PadController do
 
   @spec parse_link_props!(Keyword.t(), Pad.name_t(), State.t()) :: Keyword.t()
   defp parse_link_props!(props, pad_name, state) do
-    opts_spec = state.module.membrane_pads()[pad_name].options
+    {_, pad} =
+      state.module.membrane_pads()
+      |> PadSpecHandler.add_bin_pads()
+      |> Enum.find(fn {k, _} -> k == pad_name end)
```
mat-hek: `Keyword.get(pad_name)` would work as well

lib/membrane/core/pad_controller.ex
```diff
@@ -184,8 +184,12 @@ defmodule Membrane.Core.Element.PadController do
 
   @spec parse_link_props!(Keyword.t(), Pad.name_t(), State.t()) :: Keyword.t()
   defp parse_link_props!(props, pad_name, state) do
-    opts_spec = state.module.membrane_pads()[pad_name].options
+    {_, pad} =
```
mat-hek: this is rather `pad_spec`

lib/membrane/core/pads_specs.ex
```diff
@@ -243,4 +273,32 @@ defmodule Membrane.Core.Element.PadsSpecs do
   defp generate_pad_property_doc(_k, v) do
     "`#{inspect(v)}`"
   end
+
+  defp documenation_string(direction, bin?) do
+    sentence_object =
```
mat-hek: maybe `subject`?

lib/membrane/spec.ex
```diff
@@ -1,48 +1,55 @@
...
+    For bins boundaries there are special links allowed. User should define links
+    between bin's input and first child's input (input-input type) and last
+    child's output and bin output (output-output type). In this case, callback module
+    creator should name endpoint of the bin with macro `this_bin()`
+
+    Sample definition:
+
+    ```
+    %{
+      {this_bin(), :input} => {:filter1, :input, buffer: [preferred_size: 10]},
```
mat-hek: remember to update

lib/membrane/with_input_pads.ex
```diff
@@ -32,11 +32,13 @@ defmodule Membrane.Element.WithInputPads do
   """
   @deprecated "Use def_input_pad/2 for each pad instead"
   defmacro def_input_pads(pads) do
+    for name <- pads, do: Pad.assert_public_name!(name)
```
mat-hek: if still needed, this should be moved to `PadSpecs.def_pads`

lib/membrane/core/pad_controller.ex
kanes115: It wouldn't as after we have changed pad name from `atom` to `atom | {:private, atom}` it's no longer a keyword list.

lib/membrane/core/pads_specs.ex
kanes115: Well, this is rather `object` than `subject`, isn't it? Anyway I've inlined it and don't need variable anymore.

lib/membrane/core/parent/message_dispatcher.ex
kanes115: I agree with the part that this, in fact, is not (only) a message dispatcher and should not be named this. The fact that it does not handle ALL of the messages is not a problem as we only want to handle parent-specific ones though. That being said I would rename this to `Parent.LifecycleController` and then also create a `Parent.MessageDispatcher` that would handle *common* for bin and pipeline messages. I will proceed with the former and wait for your reply with the latter.

kanes115: We will still be left with a few `handle_info/2` identical clauses in bin and pipeline which might be separated into a different file. Maybe we still would like to have a message dispatcher for parent? That being said I will create a lifecycle create and wait for your reply with creating the dispatcher.

kanes115: How about that: 88337ad8553d503f27b4c6c86dccc809358318b8

kanes115:
Rebased with master

lib/membrane/core/bin/linking_buffer.ex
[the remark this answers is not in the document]
Hajto: Yes and yes.

lib/membrane/core/pad_spec_handler.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Membrane.Core.PadSpecHandler do
+  @moduledoc false
+  # Module parsing pads specifications in elements.
+
+  alias Membrane.{Bin, Pad}
```
Hajto: This and subsequent alias could be merged together.

lib/membrane/core/pad_spec_handler.ex
```diff
@@ -0,0 +1,74 @@
...
+  alias Membrane.{Core, Element}
+  alias Core.Element
```
Hajto: This and subsequent alias could be merged together.

lib/membrane/core/pads_specs.ex
```diff
@@ -1,12 +1,13 @@
...
   # based on them.
-  alias Membrane.{Caps, Element}
-  alias Membrane.Core.Element.OptionsSpecs
-  alias Element.Pad
+  alias Membrane.Caps
+  alias Membrane.Core.OptionsSpecs
+  alias Membrane.Pad
   alias Bunch.Type
   use Bunch
+  require Membrane.Pad
```
Hajto: Maybe you could `require Pad` here? (i am not sure)

lib/membrane/core/pipeline/state.ex
```diff
@@ -0,0 +1,32 @@
...
+  # internally in Membrane.
+
+  alias Membrane.Core.{Playback, Playbackable}
+  alias Membrane.Element
+  use Bunch
+
+  @derive Playbackable
+
+  @type t :: %__MODULE__{
+          internal_state: internal_state_t | nil,
```
Hajto: Maybe order fields the same way in typespec as they were defined in struct.

test/membrane/core/element/action_handler_test.exs
```diff
@@ -1,9 +1,9 @@
 defmodule Membrane.Core.Element.ActionHandlerTest do
   use ExUnit.Case, async: true
-  alias Membrane.{ActionError, Buffer}
+  alias Membrane.{ActionError, Buffer, Pad.Data}
   alias Membrane.Core.{Message, Playback}
```
Hajto: Padmodel could get be aliased here.

lib/membrane/core/pads_specs.ex
```diff
@@ -70,21 +72,31 @@ defmodule Membrane.Core.Element.PadsSpecs do
...
+      end
+    end
+  end
+
   @doc """
   Generates `membrane_pads/0` function, along with docs and typespecs.
   """
   defmacro generate_membrane_pads(env) do
-    pads = Module.get_attribute(env.module, :membrane_pads) |> Enum.reverse()
+    pads = Module.get_attribute(env.module, :membrane_pads) |> nil_to([]) |> Enum.reverse()
```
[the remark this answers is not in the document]
kanes115: Bins can have no pads. Maybe we should check if they are not empty for elements still.

lib/membrane/core/bin/linking_buffer.ex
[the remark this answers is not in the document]
kanes115: True, if something is a buffer it should not perform any logic on the messages. Playbackbuffer does that hence the function name. Will change that but I suggest that we divide Playbackbuffer into buffer and logic module as well in the future.

lib/membrane/bin.ex
```diff
@@ -0,0 +1,392 @@
...
+  Spec can be started from `c:handle_init/1` callback or as
+  `t:Membrane.Core.Parent.Action.spec_action_t/0` action.
+  """
+  @callback handle_spec_started(
+              elements :: [Child.name_t()],
+              state :: State.internal_state_t()
+            ) ::
+              callback_return_t
+
+  defguard is_bin_name(term)
```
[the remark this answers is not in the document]
kanes115: Also should be moved to children_controller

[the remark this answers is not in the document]
kanes115: and be private :D

lib/membrane/pad.ex
```diff
@@ -83,9 +83,17 @@ defmodule Membrane.Element.Pad do
...
-  Describes how a pad should be declared in element.
+  Describes how a pad should be declared in element or bin.
   """
-  @type spec_t :: output_spec_t | input_spec_t
+  @type spec_t :: output_spec_t | input_spec_t | bin_spec_t
+
+  @typedoc """
+  For bins there are exactly the same options for both directions.
+  Specifying `:demand_unit` for bin's output pad means that you are
+  actually defining what demand units are going to be used against
```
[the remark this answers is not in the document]
kanes115: How would you phrase it?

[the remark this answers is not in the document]
bblaszkow06: `demand units will be used to make demands to bin's elements connected to its input pad`
On second thought this is kind of inconsistent with element's pad behaviour - I don't see why bin allows defining demand unit while element's pads don't. Either both should allow that or both shouldn't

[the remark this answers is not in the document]
kanes115: It is because this is not actually an output pad demand_unit but a private input pad demand unit option. Maybe we should call it differently to avoid confusion? The reason it is not designed in a better fashion is that we apparently are going to get rid of this option anyways (according to @mat-hek) so we established we don't care that much about it now.

lib/membrane/bin.ex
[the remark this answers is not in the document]
kanes115: Why do you think this is a good place to raise an error? I think both approaches are okey in this case, we're not deep in the callstack so it does not matter and I've done it similarly to the way pipeline does it. I would not change it in this PR unless you feel really strong about it.

The unnecessary line has been deleted, though.

lib/membrane/pad.ex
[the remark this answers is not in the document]
bblaszkow06: Ok, so let's just rephrase it (with one more tiny change): `demand units a bin will use to make demands from bin's elements connected to its input pad`

lib/membrane/bin.ex
```diff
@@ -0,0 +1,392 @@
...
+      state =
+        %State{
+          internal_state: internal_state,
+          bin_options: bin_options,
+          module: module,
+          name: my_name
+        }
+        |> PadSpecHandler.init_pads()
+
+      Message.self(:handle_spec, spec)
```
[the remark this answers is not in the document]
kanes115: If it does does not to be done in a specific synchronicity, then I think it is best to stay asynchronous. Even more so that this code is going to be changed.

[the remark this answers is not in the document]
bblaszkow06: But this code has to be changed as PR #160 will be merged in v0.4.0 window while this PR belongs to v0.5.0. There will be a lot of work as you have to implement synchronization and clocks for bins based on new Pipeline implementation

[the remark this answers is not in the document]
kanes115: Okey, still, I have to wait for it to be merged, don't I?

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,196 @@
...
+    |> Enum.each(&change_playback_state(&1, state.playback.state))
+
+    {{:ok, children_names}, state}
+  end
+
+  @spec change_playback_state(pid, Membrane.PlaybackState.t()) :: :ok
+  defp change_playback_state(pid, new_state)
+       when Membrane.PlaybackState.is_playback_state(new_state) do
+    alias Membrane.Core.Message
+    require Message
```
Hajto: You required message globally in line 13, shouldn't that suffice?

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,196 @@
...
+
+  defp child_type(module) do
+    if module |> Bunch.Module.check_behaviour(:membrane_bin?) do
+      :bin
+    else
+      :element
+    end
+  end
+
+  defp start_child_element(%{name: name, module: module, options: options}) do
```
Hajto: I have a pretty controversial one here. This function and `start_child_bin` are almost identical. You technically could pass either `Element` or `Bin` to DRY, but that would reduce readability and I don't know wether this is worth it.

lib/membrane/core/children_controller.ex
```diff
@@ -0,0 +1,196 @@
...
+  defp exec_handle_spec_started(spec_controller_module, children_names, state) do
+    callback_res =
+      CallbackHandler.exec_and_handle_callback(
+        :handle_spec_started,
+        spec_controller_module.action_handler_module,
+        [children_names],
+        state
+      )
+
+    with {:ok, _} <- callback_res do
```
Hajto: That kind looks like it would benefit from writing it as `case do` or rewrite as `~>>`. (~>> {{:error,_}_})

lib/membrane/core/pad_model.ex
```diff
@@ -1,9 +1,11 @@
-defmodule Membrane.Core.Element.PadModel do
+defmodule Membrane.Core.PadModel do
   @moduledoc false
   # Utility functions for veryfying and manipulating pads and their data.
 
-  alias Membrane.Element.Pad
+  alias Membrane.Pad
   alias Membrane.Core.Element.State
+  alias Membrane.Bin.State
+  alias Membrane.Element
```
Hajto:
```
  alias Membrane.{Element, Pad}
  alias Membrane.Bin.State
  alias Membrane.Core.Element.State

```

lib/membrane/core/pads_specs.ex
```diff
@@ -1,12 +1,13 @@
-defmodule Membrane.Core.Element.PadsSpecs do
+defmodule Membrane.Core.PadsSpecs do
   @moduledoc false
   # Functions parsing element pads specifications, generating functions and docs
   # based on them.
-  alias Membrane.{Caps, Element}
-  alias Membrane.Core.Element.OptionsSpecs
-  alias Element.Pad
+  alias Membrane.Caps
```
Hajto:
```
alias Bunch.Type
alias Membrane.{Caps,Pad}
# ordeer of rest aliases is ok
```

lib/membrane/core/parent.ex
```diff
@@ -0,0 +1,36 @@
...
+    {:ok, state}
+  end
+
+  def handle_playback_state_changed(_old, _new, state), do: {:ok, state}
+
+  @spec change_playback_state(pid, Membrane.PlaybackState.t()) :: :ok
+  def change_playback_state(pid, new_state)
+      when Membrane.PlaybackState.is_playback_state(new_state) do
+    alias Membrane.Core.Message
+    require Message
```
Hajto: This was required globally in this module, shouldn't that suffice?

Hajto: Few lines earlier you just call `Message.send/3` without re-requiring it, how is this function different.

lib/membrane/core/parent.ex
```diff
@@ -0,0 +1,36 @@
...
+  use Membrane.Core.PlaybackHandler
+
+  alias Membrane.Core
+  alias Core.{Message, PlaybackHandler}
+
+  require Membrane.PlaybackState
+  require Message
+
+  def handle_playback_state(_old, new, state) do
+    children_pids = state |> Core.Parent.ChildrenModel.get_children() |> Map.values()
```
Hajto: I think spanning this vertically into 3 lines would make it look better and easier to read.

lib/membrane/core/parent/action.ex
```diff
@@ -0,0 +1,33 @@
...
+      {:ok, state}
+    else
+      {:error, reason} ->
+        {{:error, {:cannot_forward_message, [element: element_name, message: message], reason}},
+         state}
+    end
+  end
+
+  def handle_remove_child(children, state) do
+    with {:ok, pids} <-
```
Hajto: This particular use of `with` bothered me for a while and I wandered a moment how to rewrite it better. I think you are using it a little bit against it's purpose. You are using `with` to flow control over shape of single value and only for executing side effects.

This may be controversial but I would rathar form this as:
```
result = computation()
if match?({:ok, _}, result do
  send some messages
end
{result, state}
```

I would be open to discussion on this one, because I am really curious what would be best course of action in this scenario.

kanes115: I consider using `match?` in such circumstances a bad idea. I agree, though, that the `with` clause here is too long and could be moved before `with` (assigned to a variable).

Like this:
```
x = computation()
with {:ok, res} <- x, do: action(res)
```

The reason being: pattern matching is an essential tool in elixir and usually should be preferred.

lib/membrane/core/children_controller.ex
kanes115: Well, this is a good suggestion but IMO would require us to create an "interface" for modules that can be passed to this function (that would define functions `start_link` and `set_controlling_pid`). This would probably be done by creating a module (something like `Membrane.Core.Child`) that would define these as callbacks. 

I am not sure (honestly, give me your opinion here) if this is worth the candle (the behaviour module would be very thin, we would introduce an abstraction that would actually make little good). On the other hand we already have `Parent` (i.e. `ParentError` in this function so we could be good with having Child).

Aside: It is sometimes fine to violate DRY to keep the code more readable or flexible (if the two repeated pieces of code are logically different).

Anyway, by creating the interface we stop being controversial :)

lib/membrane/core/pad_model.ex
kanes115: It's a pitty that formatter can't reorder and group top-level aliases.

kanes115: Or rather that there is no tool that would do that (maybe we should develop one) as formatter  does not want to change the AST.

lib/membrane/core/pads_specs.ex
```diff
@@ -123,10 +154,11 @@ defmodule Membrane.Core.Element.PadsSpecs do
   @spec parse_pad_specs!(
           specs :: Pad.spec_t(),
           direction :: Pad.direction_t(),
-          declaration_env :: Macro.Env.t()
+          declaration_env :: Macro.Env.t(),
```
kanes115: Spec is wrong! Arguments switch.

lib/membrane/core/pipeline/link.ex
```diff
@@ -1,7 +1,7 @@
-defmodule Membrane.Pipeline.Link do
+defmodule Membrane.Core.Link do
```
kanes115: File name does not match module name! It should be moved to `/lib/membrane/core/link.ex`

lib/membrane/spec.ex
```diff
@@ -1,48 +1,58 @@
...
+  creator should name endpoint of the bin with macro `this_bin()`
+
+  Sample definition:
+
+  ```
+  %{
+    {Bin.itself(), :input} => {:filter1, :input, buffer: [preferred_size: 10]},
+    {:filter1, :output} => {:filter2, :input, buffer: [preferred_size: 10]},
+    {:filter2, :output} => {Bin.itself(), :output, buffer: [preferred_size: 10]}
+  }
```
kanes115: Lacking closing ```

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  require Message
+
+  @type handlers :: %{
+          action_handler: module(),
+          playback_controller: module(),
+          spec_controller: module()
+        }
+
+  @spec handle_message(Message.t(), :info | :call | :other, State.t()) ::
```
kanes115: Wrong spec!

lib/membrane/core/pipeline/link.ex
kanes115: Corrected on a new PR.

lib/membrane/spec.ex
kanes115: Corrected on a new PR.

lib/membrane/core/parent/message_dispatcher.ex
kanes115: Corrected on a new PR.

kanes115:
Closing due to #201 PR that resolves merge conflicts.

PR #176, opened 2019-07-31
Handle_shutdown callback

DominikWolek:
# handle_shutdown

Added `handle_shutdown` (executed on `terminate`) callback to Pipeline.

Updated `handle_shutdown` callback in Element.

### DIFF
--- a/lib/membrane/element.ex
+++ b/lib/membrane/element.ex
@@ -127,7 +127,7 @@ defmodule Membrane.Element do
   It will wait for reply for amount of time passed as second argument
   (in milliseconds).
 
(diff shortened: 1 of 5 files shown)


lib/membrane/element.ex
```diff
@@ -127,7 +127,7 @@ defmodule Membrane.Element do
   It will wait for reply for amount of time passed as second argument
   (in milliseconds).
 
-  Will trigger calling `c:Membrane.Element.Base.handle_shutdown/1`
+  Will trigger calling `c:Membrane.Element.Base.handle_shutdown/2`
```
Hajto: calling the

PR #184, opened 2019-08-30
Fix monitoring processes from elements

bblaszkow06:


lib/membrane/core/element/state.ex
```diff
@@ -62,6 +64,7 @@ defmodule Membrane.Core.Element.State do
       pads: nil,
       watcher: nil,
       controlling_pid: nil,
+      parent_monitor: parent_monitor,
```
Hajto: Why is `parent_monitor` optional? Is there a case where element is not monitored?

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,57 @@
...
+
+  alias __MODULE__.SomeElement
+  alias Membrane.Element
+  alias Membrane.Core.Message
+
+  require Membrane.Core.Message
+
+  describe "Not linked element" do
+    test "should shutdown when pipeline is down" do
+      task =
```
Hajto: @bblaszkow06 
I think that spawn would suffice here. In that case you would have to stop relying on `Task.await` and just rely on assert_receive from line 32. Both of these operations are blocking. Right now you are kinda blocking execution so other blocking operation won't be blocking.

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,57 @@
...
+        Task.async(fn ->
```
Hajto: Both of these tests utilize this task. I would move it's creation into `setup` block.

bblaszkow06: And I wouldn't:
1. This is a one-liner, no need to extract a common part
2. When this process is spawned in a test It's more obvious what it does
3. Both tests use this process in a slightly different manner - it would make sense to move it to `setup` block if `on_exit` could be used for both

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,49 @@
+defmodule Membrane.Core.ElementTest do
+  use ExUnit.Case, async: true
+
+  defmodule SomeElement do
+    use Membrane.Element.Base.Source
```
kanes115: This is deprecated, use `Membrane.Source` instead

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,49 @@
...
+    def_output_pad :output, caps: :any
+
+    @impl true
+    def handle_other(msg, _ctx, state) do
+      {{:ok, notify: msg}, state}
+    end
+  end
+
+  alias __MODULE__.SomeElement
```
kanes115: Why do we need this alias?

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    test "should not assume pipeline is down when getting any monitor message" do
+      monitored_proc = spawn(fn -> receive do: (:exit -> :ok) end)
+      on_exit(fn -> send(monitored_proc, :exit) end)
+
+      {:ok, elem_pid} = Element.start(monitored_proc, SomeElement, :name, %{}, [])
+      :ok = Element.set_watcher(elem_pid, self())
+      ref = make_ref()
+      self = self()
```
kanes115: Is shadowing Kernel functions the best idea? Why not `me = self()`?

test/membrane/core/element_test.exs
bblaszkow06: This was merged to branch `v0.3`, `Membrane.Source` does not exist before `v0.4`

test/membrane/core/element_test.exs
bblaszkow06: You're right, it's not needed (nesting modules is not something I do often, haven't realised alias is not needed in such cases)

PR #189, opened 2019-09-17
Integration tests for Clocks and Timers

kanes115:
This PR adds integration tests for Clock and Timer module. It introduces some longer tests that will be run only on CircleCI by default.

### DIFF
--- a/test/membrane/integration/sync_test.exs
+++ b/test/membrane/integration/sync_test.exs
@@ -1,0 +1,45 @@
+defmodule Membrane.Integration.SyncTest do
+  use ExUnit.Case, async: false
+
+  alias Membrane.Support.Sync
+  alias Membrane.{Time, Testing}
+
+  @error 8
+  @timeout 500
+
+  @tag :long_running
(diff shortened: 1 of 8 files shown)


test/membrane/integration/sync_test.exs
```diff
@@ -0,0 +1,45 @@
...
+  use ExUnit.Case, async: false
+
+  alias Membrane.Support.Sync
+  alias Membrane.{Time, Testing}
+
+  @error 8
+  @timeout 500
+
+  @tag :long_running
+  test "When ration = 1 amount of lost ticks is roughly the same regardless of the time of transmission" do
```
bblaszkow06: `ration`? You meant `ratio`?

test/support/sync/sink.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Membrane.Support.Sync.Sink do
+  use Membrane.Sink
+
+  def_input_pad :input, caps: :any, demand_unit: :buffers
+
+  def_clock()
+
+  @impl true
+  def handle_init(_opts), do: {:ok, :ignored}
```
bblaszkow06: This callback can be removed.

PR #192, opened 2019-09-19
Added fail path for Testing Pipeline start

Hajto:
Before there was possibility to pass invalid options to `Membrane.Testing.Pipeline.start/_link/2`. For example:
 - no elements nor module
 - both module and elements provided

### DIFF
--- a/lib/membrane/testing/pipeline.ex
+++ b/lib/membrane/testing/pipeline.ex
@@ -170,18 +170,25 @@ defmodule Membrane.Testing.Pipeline do
     do_start(:start, pipeline_options, process_options)
   end
 
-  defp do_start(_type, %Options{elements: nil, module: nil}, _process_options),
-    do: {:error, :no_config}
+  defp do_start(_type, %Options{elements: nil, module: nil}, _process_options) do
+    raise """
+
+    You provided no information about pipeline contents. Please provide either:
+     - list of elemenst via `elements` field of Options struct with optional links between
+     them via `links` field of `Options` struct
+     - module that implements `Membrane.Pipeline` callbacks via `module` field of `Options`
+     struct
+    """
+  end
 
   defp do_start(_type, %Options{elements: elements, module: module}, _process_options)
        when is_atom(module) and module != nil and elements != nil do
-    warn("""
+    raise """
+
     When working with Membrane.Testing.Pipeline you can't provide both
     override module and elements list in the Membrane.Testing.Pipeline.Options
     struct.
-    """)
-
-    {:error, :wrong_options}
+    """
   end
 
   defp do_start(type, options, process_options) do
--- a/lib/membrane/testing/pipeline.ex
+++ b/lib/membrane/testing/pipeline.ex
@@ -97,6 +97,7 @@ defmodule Membrane.Testing.Pipeline do
   """
 
   use Membrane.Pipeline
+  use Membrane.Log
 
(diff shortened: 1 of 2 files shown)


lib/membrane/testing/pipeline.ex
```diff
@@ -162,11 +163,31 @@ defmodule Membrane.Testing.Pipeline do
...
+    do_start(:start_link, pipeline_options, process_options)
   end
 
   def start(pipeline_options, process_options \\ []) do
-    Pipeline.start(__MODULE__, default_options(pipeline_options), process_options)
+    do_start(:start, pipeline_options, process_options)
+  end
+
+  defp do_start(_type, %Options{elements: nil, module: nil}, _process_options),
+    do: {:error, :no_config}
```
Hajto: This name

lib/membrane/testing/pipeline.ex
```diff
@@ -162,11 +163,31 @@ defmodule Membrane.Testing.Pipeline do
...
+
+  defp do_start(_type, %Options{elements: elements, module: module}, _process_options)
+       when is_atom(module) and module != nil and elements != nil do
+    warn("""
+    When working with Membrane.Testing.Pipeline you can't provide both
+    override module and elements list in the Membrane.Testing.Pipeline.Options
+    struct.
+    """)
+
+    {:error, :wrong_options}
```
Hajto: And this may need discussion.

lib/membrane/testing/pipeline.ex
bblaszkow06: The solution for your naming problems is simple - don't use them at all. Instead of returning an error tuple, raise!

test/membrane/testing/pipeline_test.exs
```diff
@@ -52,12 +52,15 @@ defmodule Membrane.Testing.PipelineTest do
 
   describe "When starting Testing Pipeline does" do
     test "returns an error if a pipeline is started with both elements and module provided in options" do
-      assert {:error, :wrong_options} =
-               Pipeline.start(%Pipeline.Options{elements: [elem: Elem], module: Mod})
+      assert_raise RuntimeError, fn ->
```
bblaszkow06: A minor thing but still - test description is now invalid

bblaszkow06: Plus some assertion about returned error message should be there (it doesn't have to be a copy paste of whole message but some regex making sure it's the error you wanted to see)

test/membrane/testing/pipeline_test.exs
```diff
@@ -52,12 +52,15 @@ defmodule Membrane.Testing.PipelineTest do
...
+        Pipeline.start(%Pipeline.Options{elements: [elem: Elem], module: Mod})
+      end
     end
 
     test "returns an error if no means of generating spec are provided (no elements, no module)" do
-      assert {:error, :no_config} = Pipeline.start(%Pipeline.Options{elements: nil, module: nil})
+      assert_raise RuntimeError, fn ->
```
bblaszkow06: same as for the test above x2

test/membrane/testing/pipeline_test.exs
Hajto: raises not returns. gotcha

Hajto: I've asserted crucial part of the error message.

PR #194, opened 2019-09-19
Fix monitoring processes from elements

### LINKED ISSUE #185: Fix monitoring from element
Fix from ec4ba2f4790ec5239614704498bc1e93c6709ad5 has to applied to master

Hajto:
Closes #185

### DIFF
--- a/lib/membrane/core/element.ex
+++ b/lib/membrane/core/element.ex
@@ -130,8 +130,8 @@ defmodule Membrane.Core.Element do
 
   @impl GenServer
   def init(options) do
-    Process.monitor(options.pipeline)
-    state = State.new(options)
+    parent_monitor = Process.monitor(options.pipeline)
+    state = options |> Map.put(:parent_monitor, parent_monitor) |> State.new()
 
     with {:ok, state} <-
            MessageDispatcher.handle_message(
@@ -159,7 +159,7 @@ defmodule Membrane.Core.Element do
   end
 
   @impl GenServer
-  def handle_info({:DOWN, _ref, :process, _pid, reason}, state) do
+  def handle_info({:DOWN, ref, :process, _pid, reason}, %{parent_monitor: ref} = state) do
     {:ok, state} =
       MessageDispatcher.handle_message(Message.new(:pipeline_down, reason), :info, state)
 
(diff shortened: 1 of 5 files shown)


test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,68 @@
...
+  describe "Not linked element" do
+    test "should shutdown when pipeline is down" do
+      monitored_proc = spawn(fn -> receive do: (:exit -> :ok) end)
+
+      {:ok, elem_pid} =
+        monitored_proc
+        |> element_init_options
+        |> Element.start()
+
+      {:ok, _clock} = Message.call(elem_pid, :handle_watcher, self())
```
Hajto: This line works same as `:ok = Element.set_watcher(elem_pid, self())` (based on my archeology work).
First when making this PR I disabled it because I couldn't find the proper substitute at first and this test passed without this line. If I understand watcher correctly, element will notify the watcher when it gets a down message?

Maybe we should refute?

test/membrane/core/element_test.exs
```diff
@@ -0,0 +1,68 @@
...
+      ref = Process.monitor(elem_pid)
+      send(monitored_proc, :exit)
+      assert_receive {:DOWN, ^ref, :process, ^elem_pid, :normal}
+    end
+
+    test "should not assume pipeline is down when getting any monitor message" do
+      monitored_proc = spawn(fn -> receive do: (:exit -> :ok) end)
```
Hajto: Why are we creating here `monitored_proc` if it isn't used anywhere further in the test. Do we want to have process that is alive during test execution and keep test executor mailbox sane?

test/membrane/core/element_test.exs
bblaszkow06: There's no such wrapper in v0.4.0, you can create one ;) `:handle_watcher` is basically `:set_watcher` but returns a clock as well

test/membrane/core/element_test.exs
bblaszkow06: `monitored_proc` is needed for elemtent to set up watcher, but it can be `self()` as well

lib/membrane/core/element.ex
```diff
@@ -130,8 +130,8 @@ defmodule Membrane.Core.Element do
 
   @impl GenServer
   def init(options) do
-    Process.monitor(options.pipeline)
-    state = State.new(options)
+    parent_monitor = Process.monitor(options.pipeline)
```
bblaszkow06: Let's change `pipeline` to `parent` or `parent_process` to be consistent with state struct field name and make it ready for bins

PR #197, opened 2019-09-24
Dialyzer fixes after synchronization merge.

Hajto:


lib/membrane/core/callback_handler.ex
```diff
@@ -166,13 +161,13 @@ defmodule Membrane.Core.CallbackHandler do
   end
 
   @spec parse_callback_result(callback_return_t | any, module, callback :: atom, handler_params_t) ::
-          {:ok, Type.stateful_try_t(list, internal_state_t)}
+          {{:ok, list}, internal_state_t}
```
Hajto: As discussed IRL, only `:ok` is going to be returned, because erronous raises an error.

test/membrane/core/element/event_controller_test.exs
```diff
@@ -0,0 +1,96 @@
...
+
+      state = %{state | pads: pads}
+
+      assert {:ok, state} = EventController.handle_event(:input, %Event.EndOfStream{}, state)
+      assert state.pads.data.input.end_of_stream?
+    end
+  end
+
+  describe "Event controller handles normal events" do
+    test "handles ignore", %{state: state} do
```
bblaszkow06: If you join describe and test strings, the test name doesn't make much sense

test/membrane/pipeline_test.exs
```diff
@@ -1,15 +1,50 @@
...
-  use ExUnit.Case
 
   defp state(_ctx) do
     [state: %State{module: nil, clock_proxy: nil}]
   end
 
   setup_all :state
 
+  describe "Handle init" do
+    test "valid pipeline" do
```
bblaszkow06: You've swapped tests names

bblaszkow06: plus this name should be more similar to the other

lib/membrane/core/timer.ex
```diff
@@ -3,7 +3,9 @@ defmodule Membrane.Core.Timer do
   require Membrane.Core.Message
   alias Membrane.Core.Message
   alias Membrane.Time
+  alias Membrane.Clock
 
+  @type id_t :: any()
```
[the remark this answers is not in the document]
Hajto: I've put it back because it was used in other file previously. Now it is also used in struct.

lib/membrane/core/callback_handler.ex
Hajto: Ok my bad. I missread the code.

test/membrane/pipeline_test.exs
Hajto: renamed both tests

Hajto:
I also muted warnings about protocol implementations.

bblaszkow06:
> I also muted warnings about protocol implementations.

As discussed, this is not needed as dialyxir rc.7 have been released fixing this issue

PR #198, opened 2019-09-26
Fix/dynamically added elements

bblaszkow06:


lib/membrane/core/playback_handler.ex
```diff
@@ -73,6 +73,7 @@ defmodule Membrane.Core.PlaybackHandler do
   def state_change_callback(:prepared, :playing), do: :handle_prepared_to_playing
   def state_change_callback(:prepared, :stopped), do: :handle_prepared_to_stopped
 
+  @spec change_playback_state(PlaybackState.t(), module(), Playbackable.t()) :: handler_return_t()
```
Hajto: handler_return_t assumes that error can be any,
After a once over the file I found those as possible options:

- playback_already_unlocked
- invalid_current_playback_state
- invalid_target_playback_state
- target_and_current_states_equal

If error is not explicitly one those, maybe we could change handler_return_t and make the error reason at least an atom narrowing down the type?

lib/membrane/pipeline.ex
```diff
@@ -579,7 +579,12 @@ defmodule Membrane.Pipeline do
     children_pids |> Enum.each(&change_playback_state(&1, new))
     :ok = toggle_syncs_active(old, new, children_data)
     state = %{state | pending_pids: children_pids |> MapSet.new()}
-    PlaybackHandler.suspend_playback_change(state)
+
+    if children_pids |> Enum.empty?() do
```
Hajto: This is fix for no-elements = no state change problem?

test/membrane/core/element/action_handler_test.exs
```diff
@@ -533,7 +533,6 @@ defmodule Membrane.Core.Element.ActionHandlerTest do
   end
 
   describe "handling_actions" do
-    @describetag :focus
```
Hajto: what did ":focus" did?

test/membrane/integration/sync_test.exs
```diff
@@ -102,6 +102,26 @@ defmodule Membrane.Integration.SyncTest do
...
+    options = %Testing.Pipeline.Options{
+      module: Membrane.Support.Sync.Pipeline,
+      custom_args: %Membrane.Pipeline.Spec{}
+    }
+
+    {:ok, pipeline} = Testing.Pipeline.start_link(options)
+    :ok = Testing.Pipeline.play(pipeline)
+
+    assert_pipeline_playback_changed(pipeline, :stopped, :prepared)
+    assert_pipeline_playback_changed(pipeline, :prepared, :playing)
```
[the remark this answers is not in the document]
bblaszkow06: Well, it was done to clear up the message queue - I had an error later in test and without these assertions some messages were unhandled

test/membrane/core/element/action_handler_test.exs
bblaszkow06: Nothing special. It was used to run only the tests I wanted via `mix test --only focus`. It comes from ESpec where using `fit` instead of `it` or `fexample` in place of `example` allowed to run these test with `mix espec --focus`

lib/membrane/pipeline.ex
bblaszkow06: I'm not sure what you meant by this comment.
Notice that `handle_playback_state` is a callback. It is invoked when playback state is changing. `suspend..` results in keeping the playback state and waiting for `continue...` call. This is done after all pending pids change its state. The trouble was that callback in Pipeline was called there, but I moved it to another callback - `handle_playback_state_changed`

lib/membrane/core/playback_handler.ex
bblaszkow06: Don't see the point of it. `handler_return_t` is used as a result of a callback so the returned error can have any reason and it doesn't have to be an atom (a tuple may appear for example)

lib/membrane/core/playback_handler.ex
```diff
@@ -125,6 +128,8 @@ defmodule Membrane.Core.PlaybackHandler do
     end
   end
 
+  @spec unlock_target_state(pb) :: {:ok | {:error, :playback_already_unlocked}, pb}
+        when pb: Playbackable.t()
   def unlock_target_state(playbackable) do
```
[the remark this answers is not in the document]
bblaszkow06: This might be done, but in separate PR I think

PR #202, opened 2019-10-21
Links

mat-hek:


lib/membrane/bin.ex
```diff
@@ -58,10 +58,14 @@ defmodule Membrane.Bin do
   """
   @callback membrane_clock? :: boolean()
 
-  @doc """
-  This function defines a term that allows to reference current bin from `Membrane.ParentSpec`
-  """
-  def itself, do: {__MODULE__, :itself}
+  defmodule Itself do
```
kanes115: Eventually, I think this is not a good idea. Making it implementable for Inspect changes very little: "Bin.itself()" -> "bin itself" (both are descriptive enough). At the same time this approach makes the developer think that `%Bin.Itself{}` has some fields or that we should add some fields in there. Unless there is no need to hold any extra information inside the struct I would not use struct here and leave it as it was.

lib/membrane/core/parent/link.ex
```diff
@@ -41,59 +40,55 @@ defmodule Membrane.Core.Parent.Link do
...
-        |> Keyword.keys()
-        |> Bunch.Enum.try_each(fn
-          key when key in @valid_opt_keys -> :ok
-          invalid_key -> {:error, {:invalid_opts_key, invalid_key}}
-        end)
-      else
-        {:error, {:invalid_opts, opts}}
-      end
-    end
+  def from_spec(links) when is_list(links) do
```
kanes115: Lacking spec

lib/membrane/bin.ex
```diff
@@ -255,7 +259,15 @@ defmodule Membrane.Bin do
     |> reply()
   end
 
-  defmacro __using__(_) do
+  defmacro __using__(args) do
+    bring_spec =
```
kanes115: Why not move this part to `Membrane.Parent` which is common for bin and pipeline (instead of duplicating this)?

test/membrane/parent/link_test.exs
```diff
@@ -0,0 +1,123 @@
+defmodule Membrane.Parent.LinkTest do
```
kanes115: I think this lacks a test for what happens when someone passes links in the old form. Do we deprecate that? Do we totally stop supporting this?

kanes115: Also I would like to have a test with branches and two totally separate.

PR #203, opened 2019-10-25
more element tests, fixed found bugs

mat-hek:


test/membrane/core/element_test.exs
```diff
@@ -11,11 +11,225 @@ defmodule Membrane.Core.ElementTest do
...
+               Element.handle_call(msg, nil, get_state())
+    end)
+
+    assert {:reply, {:error, {:cannot_handle_message, {:invalid_message, :abc, _}, _}}, _state} =
+             Element.handle_call(:abc, nil, get_state())
+  end
+
+  test "other message" do
+    state = get_state()
+    assert {:noreply, state} == Element.handle_info(:abc, state)
```
bblaszkow06: Shouldn't that verify that `handle_other` was actually called?

bblaszkow06:
I guess this should be merged to master, not `bins_merge`, right?

test/membrane/core/element_test.exs
```diff
@@ -11,11 +11,225 @@ defmodule Membrane.Core.ElementTest do
...
+  end
+
+  test "demand unit" do
+    assert {:noreply, state} =
+             Element.handle_info(Message.new(:demand_unit, [:bytes, :output]), linked_state())
+
+    assert state.pads.data.output.other_demand_unit == :bytes
+  end
+
+  test "demand or buffer or caps or event when not playing" do
```
kanes115: Again, more of an opinion: ExUnit gives the opportunity to define test case names as strings which is powerful. I don't like trying to shorten this and not creating sentences (add a predicate!). I can see this notion throughout the code base :( Look how horrible [this gets in erlang](https://github.com/esl/MongooseIM/blob/master/big_tests/tests/mam_SUITE.erl#L78) and still this is a sentence :D

test/membrane/core/element_test.exs
```diff
@@ -11,11 +11,225 @@ defmodule Membrane.Core.ElementTest do
...
+      Message.new(:event, %Membrane.Testing.Event{}, for_pad: :input),
+      Message.new(:event, %Membrane.Testing.Event{}, for_pad: :output)
+    ]
+    |> Enum.each(fn msg ->
+      assert {:noreply, state} = Element.handle_info(msg, initial_state)
+      assert {:ok, state} == Element.PlaybackBuffer.store(msg, initial_state)
+    end)
+  end
+
+  test "demand" do
```
kanes115: This test does not test a lot. This only tests the situation in which callback is not implemented. This is fine of course, but I feel this should be reflected in test name as now it sounds like it tests the general happy path.

kanes115: As discussed in private.
I think this callback call should be mocked somehow (preferably there should be a short callback module injected from test).

test/membrane/core/element_test.exs
```diff
@@ -11,11 +11,225 @@ defmodule Membrane.Core.ElementTest do
...
+  end
+
+  test "handle unlink" do
+    assert {:noreply, state} =
+             Element.handle_info(Message.new(:handle_unlink, :input), linked_state())
+
+    refute Map.has_key?(state.pads.data, :input)
+  end
+
+  test "timer tick" do
```
kanes115: Those kind of tests will pass even if Element implements `handle_info` fallback and not do anything. This imo makes those tests hardly useful (I believes we should assert something more).

test/membrane/core/element_test.exs
```diff
@@ -11,11 +11,225 @@ defmodule Membrane.Core.ElementTest do
...
+    ]
+    |> Enum.each(fn msg ->
+      assert {:stop, {:error, {:cannot_handle_message, {:invalid_message, ^msg, _}, _}}, _state} =
+               Element.handle_info(msg, get_state())
+
+      assert {:reply, {:error, {:cannot_handle_message, {:invalid_message, ^msg, _}, _}}, _state} =
+               Element.handle_call(msg, nil, get_state())
+    end)
+
+    assert {:reply, {:error, {:cannot_handle_message, {:invalid_message, :abc, _}, _}}, _state} =
```
kanes115: Why there's no `handle_cast`?

lib/membrane/core/element/message_dispatcher.ex
```diff
@@ -118,7 +118,7 @@ defmodule Membrane.Core.Element.MessageDispatcher do
     {:ok, new_state}
   end
 
-  defp do_handle_message(Message.new(_, _) = message, mode, state) do
+  defp do_handle_message(Message.new(_, _, _) = message, mode, state) do
```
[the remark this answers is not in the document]
mat-hek: ok, will edit the commit message

test/membrane/core/element_test.exs
mat-hek: because we don't use it at all :)

test/membrane/core/element_test.exs
mat-hek: These tests are for the element core, i.e. GenServer behind the element. I think checks for callbacks being called should be in proper controllers or integration tests

PR #204, opened 2019-10-29
Bin and Pipeline commons refactor

mat-hek:


lib/membrane/core/parent/child.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Membrane.Core.Parent.Child do
```
kanes115: This name is confusing. Maybe: `ChildEntry` instead?

PR #205, opened 2019-10-29
Fix bins' issues

kanes115:
This PR fixes things around Bin and dynamic pads' callbacks:
* Creates those callbacks for bin
* Moves `CallbackContext` out of `Element` namespace
* Membrane does not call `handle_pad_added` and `handle_pad_removed` on private pads anymore
* fixes toilet problem (not being created when needed)
* Fixes RC in linking buffer for pads that are being dynamically linked


lib/membrane/child.ex
```diff
@@ -3,6 +3,49 @@ defmodule Membrane.Child do
   Module that keeps track of types used by both elements and bins
   """
   alias Membrane.{Bin, Element}
+  alias Membrane.Element.{Action, CallbackContext}
```
mat-hek: This uses an element's callback context. Actually the pad controller also uses it, which I think is counterintuitive and will cause problems when contexts diverge. For that reason, I'm not sure if we should move pad callbacks to commons.

kanes115: Or duplicate `handle_pad_added/removed` for bins (as they differ in, for instance, what actions can be returned)

lib/membrane/core/child/pad_controller.ex
```diff
@@ -321,6 +322,12 @@ defmodule Membrane.Core.Child.PadController do
   defp handle_pad_added(ref, state) do
     pad_opts = PadModel.get_data!(state, ref, :options)
 
+    action_handler =
+      case state do
+        %Membrane.Core.Bin.State{} -> Membrane.Core.Bin.ActionHandler
+        _ -> ActionHandler
```
mat-hek: - match on element's state
- keeping any ActionHandler aliased makes it ambiguous when reading

lib/membrane/bin.ex
```diff
@@ -66,6 +70,26 @@ defmodule Membrane.Bin do
...
               | {:error, any}
 
+  @doc """
+  Callback that is called when new pad has beed added to bin. Executed
+  ONLY for dynamic pads.
+  """
+  @callback handle_pad_added(
+              pad :: Pad.ref_t(),
+              context :: CallbackContext.PadAdded.t(),
+              state :: state_t()
```
mat-hek: This type refers to bin's internal state

lib/membrane/bin.ex
```diff
@@ -66,6 +70,26 @@ defmodule Membrane.Bin do
...
+            ) :: callback_return_t
+
+  @doc """
+  Callback that is called when some pad of the bin has beed removed. Executed
+  ONLY for dynamic pads.
+  """
+  @callback handle_pad_removed(
+              pad :: Pad.ref_t(),
+              context :: CallbackContext.PadRemoved.t(),
+              state :: state_t()
```
mat-hek: This type refers to bin's internal state

lib/membrane/bin/callback_context/pad_added.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Membrane.Bin.CallbackContext.PadAdded do
+  @moduledoc """
+  Structure representing a context that is passed to the bin when
+  when new pad added is created
```
mat-hek:
```suggestion
  when a new dynamic pad instance is created
```

lib/membrane/bin/callback_context/pad_removed.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Membrane.Bin.CallbackContext.PadRemoved do
+  @moduledoc """
+  Structure representing a context that is passed to the bin
+  when pad is removed
```
mat-hek:
```suggestion
  when a dynamic pad is removed
```

lib/membrane/element/callback_context.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Membrane.Element.CallbackContext do
+defmodule Membrane.CallbackContext do
```
mat-hek: you should also move the file

mat-hek:
also, have a look at `Membrane.Pad.Data` docs and mark fields that are valid for elements only

kanes115:
Problems with `handle_pad_added` being invoked after `handle_demand` should be sorted out now thanks to 561922eed5fde3ce6a1e2be5124b0e4880c6990f and b2e5812a53aadc6b8a18675d66cc60c7f18104a4.

lib/membrane/pad_data.ex
```diff
@@ -7,7 +7,7 @@ defmodule Membrane.Pad.Data do
       See `Membrane.Caps.Matcher` for more information
     - `:availability` - see `Membrane.Pad.availability_t`
     - `:caps` - the most recent `Membrane.Caps` that have been sent (output) or received (input)
-      on the pad. May be `nil` if not yet set
+      on the pad. May be `nil` if not yet set. This field only applies to elements' pads.
     - `:demand` - current demand requested on the pad working in pull mode
```
mat-hek: this also applies to elements only, and some other too

lib/membrane/pad.ex
```diff
@@ -165,6 +165,9 @@ defmodule Membrane.Pad do
   defguard is_availability_dynamic(availability) when availability == :on_request
   defguard is_availability_static(availability) when availability == :always
 
+  def public_name?(name) when is_public_name(name), do: true
```
mat-hek: but why?

lib/membrane/core/bin/linking_buffer.ex
```diff
@@ -26,34 +26,52 @@ defmodule Membrane.Core.Bin.LinkingBuffer do
                    to be sent
   * bin_state - state of the bin
   """
-  @spec store_or_send(t(), Message.t(), Pad.ref_t(), State.t()) :: t()
-  def store_or_send(buf, msg, sender_pad, bin_state) do
+  @spec store_or_send(Message.t(), Pad.ref_t(), State.t()) :: State.t()
+  def store_or_send(msg, sender_pad, bin_state) do
+    buf = bin_state.linking_buffer
+
     case PadModel.get_data(bin_state, sender_pad) do
```
mat-hek: this will forward message to a pad before handle linking finished. you should probably also check in dynamic_currently_linking

lib/membrane/pad.ex
kanes115: So that you don't have to require `Pad` before using `public_name?`/`is_public_name`?

kanes115: I will change it though

lib/membrane/core/bin/linking_buffer.ex
kanes115: Isn't this the point? I am checking if the pad is already linked by fetching its data from state. If it's there I simply send the message to it, otherwise I store.

mat-hek: yes, but you cannot be sure whether the pad is linked on the other side until `handle_linking_finished`

lib/membrane/core/element/demand_controller.ex
```diff
@@ -18,34 +18,43 @@ defmodule Membrane.Core.Element.DemandController do
...
+    end
+  end
 
-      {total_size, state} =
-        state
-        |> PadModel.get_and_update_data!(pad_ref, :demand, fn demand ->
-          (demand + size) ~> {&1, &1}
-        end)
+  @spec ignore?(Pad.ref_t(), State.t()) :: boolean()
+  defp ignore?(pad_ref, state), do: state.pads.data[pad_ref].mode == :push
```
mat-hek: this is overkill

kanes115: I don't think it is. This way you tell explicitly that you IGNORE demands. I don't like lines this long `state.pads.data[pad_ref].mode == :push` in if's clause, to be honest.

PR #209, opened 2019-11-04
Pad names

mat-hek:


test/support/bin/test_bins.ex
```diff
@@ -81,7 +121,25 @@ defmodule Membrane.Support.Bin.TestBins do
...
+      links = [
+        link(:filter2) |> to_bin_output(pad)
+      ]
+
+      {{:ok, spec: %ParentSpec{links: links}}, state}
+    end
+
+    def handle_pad_added(_pad, _ctx, state) do
+      {:ok, state}
+    end
```
bblaszkow06: I think this case should not be handled and should result in an error

lib/membrane/parent.ex
```diff
@@ -95,10 +95,19 @@ defmodule Membrane.Parent do
         end
       end
 
+    bring_pad =
+      if args |> Keyword.get(:bring_pad?, true) do
```
bblaszkow06: This option is not documented!

lib/membrane/element/base.ex
```diff
@@ -293,7 +293,15 @@ defmodule Membrane.Element.Base do
     end
   end
 
-  defmacro __using__(_) do
+  defmacro __using__(args \\ []) do
+    bring_pad =
+      if args |> Keyword.get(:bring_pad?, true) do
```
bblaszkow06: This option is not documented!

lib/membrane/parent_spec.ex
```diff
@@ -57,6 +57,21 @@ defmodule Membrane.ParentSpec do
         link_bin_input() |> to(:filter1) |> to(:filter2) |> to_bin_output(:custom_output)
       ]
 
+  ### Dynamic pads
+
+  In most cases dynamic pads can be linked the same way as static ones, although
+  in the following situations exact pad reference must be passed:
+
+  - When pad id, apart from distinguishing different instances of the same pad,
+    has some specific semantic meaning
```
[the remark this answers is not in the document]
mat-hek: no idea :D

lib/membrane/element/base.ex
```diff
@@ -293,7 +293,23 @@ defmodule Membrane.Element.Base do
...
+  Brings common stuff needed to implement an element. Used by
+  `Membrane.Source.__using__/1`, `Membrane.Filter.__using__/1`
+  and `Membrane.Sink.__using__/1`.
+
+  Options:
+    - `:bring_pad?` - if true (default) requires and aliases `Membrane.Pad`
+  """
+  defmacro __using__(options) do
+    bring_pad =
+      if options |> Keyword.get(:bring_pad?, true) do
```
kanes115: As discussed, I don't like this idea. `use`ing `Membrane.Pad` would be better imo.

lib/membrane/parent_spec.ex
```diff
@@ -148,101 +170,141 @@ defmodule Membrane.ParentSpec do
...
   end
 
-  @spec to_bin_output(LinkBuilder.t(), Pad.name_t(), pad_options_t) :: LinkBuilder.t() | no_return
-  def to_bin_output(builder, pad \\ :output, opts \\ [])
+  @doc """
+  Ends a link with a bin's output.
+
+  See the _links_ section of the moduledoc for more information.
+  """
+  @spec to_bin_output(LinkBuilder.t(), Pad.name_t(), pad_props_t) :: LinkBuilder.t() | no_return
```
kanes115: Shouldn't it be `Pad.ref_t()` instead of `name_t`?

test/membrane/core/parent/child_life_controller/link_handler_test.exs
```diff
@@ -0,0 +1,157 @@
...
+  defp get_state(child_module, availability \\ :always) do
+    %Membrane.Core.Bin.State{
+      module: nil,
+      name: :my_bin,
+      clock_proxy: nil,
+      children: [a: %ChildEntry{module: child_module}, b: %ChildEntry{module: child_module}],
+      pads: %{
+        info:
+          [input: %{availability: availability}, output: %{availability: availability}]
+          |> Enum.flat_map(fn {name, info} -> [{name, info}, {{:private, name}, info}] end)
```
kanes115: I would consider putting it to a different function as it gets messy here.

PR #213, opened 2019-11-14
Enable toilet on message

kanes115:
Toilet has to be enabled after linking as it needs to know if behind bin's edge there is a push mode pad. This is impossible in linking phase as linking is not synchronized between bin's internal nad external links (public/private pads)

### DIFF
--- a/lib/membrane/core/child/pad_controller.ex
+++ b/lib/membrane/core/child/pad_controller.ex
@@ -39,7 +39,7 @@ defmodule Membrane.Core.Child.PadController do
     with :ok <- validate_pad_being_linked!(pad_ref, direction, info, state),
          :ok <- validate_dir_and_mode!({pad_ref, info}, {other_ref, other_info}) do
       props = parse_pad_props!(props, pad_name, state)
-      state = init_pad_data(info, pad_ref, pid, other_ref, other_info, props, state)
+      state = init_pad_data(info, pad_ref, pid, other_ref, props, state)
 
       state =
         case Pad.availability_mode(info.availability) do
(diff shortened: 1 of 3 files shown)


lib/membrane/core/child/pad_controller.ex
```diff
@@ -91,13 +91,30 @@ defmodule Membrane.Core.Child.PadController do
...
 
+  defp send_push_mode_announcments(state) do
+    state.pads.data
+    |> Map.values()
+    |> Enum.filter(&(&1.mode == :push))
+    |> Enum.each(&Message.send(&1.pid, :push_mode_announcment, [], for_pad: &1.other_ref))
+  end
+
+  @spec enable_toilet(Pad.ref_t(), state_t()) :: {:ok, state_t()}
+  def enable_toilet(pad_ref, state) do
```
mat-hek: all this can be done with a single `PadModel.update_pad_data` call

lib/membrane/core/child/pad_controller.ex
```diff
@@ -262,18 +279,15 @@ defmodule Membrane.Core.Child.PadController do
     %{}
   end
 
-  defp init_pad_mode_data(%{mode: :pull, direction: :input} = data, other_info, props, state) do
+  defp init_pad_mode_data(%{mode: :pull, direction: :input} = data, _other_info, props, state) do
```
mat-hek: `other_info` seems to be unneeded in all clauses now

lib/membrane/core/element/demand_controller.ex
```diff
@@ -18,30 +18,34 @@ defmodule Membrane.Core.Element.DemandController do
   @spec handle_demand(Pad.ref_t(), non_neg_integer, State.t()) ::
           State.stateful_try_t()
   def handle_demand(pad_ref, size, state) do
-    PadModel.assert_data(state, pad_ref, %{direction: :output})
+    if state.pads.data[pad_ref].mode == :pull do
```
mat-hek: this if is too large

lib/membrane/core/input_buffer.ex
```diff
@@ -81,43 +83,40 @@ defmodule Membrane.Core.InputBuffer do
...
-
     %__MODULE__{
       name: name,
       q: @qe.new(),
       preferred_size: preferred_size,
       min_demand: min_demand,
       demand: preferred_size,
       metric: metric,
-      toilet: toilet
+      # Toilet can be enabled with `enable_toilet/1`
```
mat-hek: this comment is unneeded

PR #216, opened 2019-11-18
Fix EOS and unlinking in bins problems

kanes115:
This PR fixes:
* Wrong elements' behaviour on duplicated `EndOfStream` events
* Not notifying parents about `EndOfStream` event
* Not generating `EndOfStream` when elements went from `:playing` to `:prepared`

### DIFF
--- a/lib/membrane/bin.ex
+++ b/lib/membrane/bin.ex
@@ -247,6 +247,12 @@ defmodule Membrane.Bin do
   end
 
   @impl GenServer
+  def handle_info(Message.new(:handle_unlink, pad_ref), state) do
+    PadController.handle_pad_removed(pad_ref, state)
+    |> noreply()
+  end
+
+  @impl GenServer
   def handle_info(message, state) do
     Parent.MessageDispatcher.handle_message(message, state)
   end
--- a/lib/membrane/core/element/event_controller.ex
+++ b/lib/membrane/core/element/event_controller.ex
@@ -122,8 +122,11 @@ defmodule Membrane.Core.Element.EventController do
   end
 
   defp handle_special_event(pad_ref, %Event.EndOfStream{}, state) do
-    with %{direction: :input, start_of_stream?: true, end_of_stream?: false} <-
-           PadModel.get_data!(state, pad_ref) do
+    pad_data = PadModel.get_data!(state, pad_ref)
+
+    with %{direction: :input, start_of_stream?: true, end_of_stream?: false} <- pad_data,
+         %{playback: playback} <- state,
+         %{state: :playing} <- playback do
       state
       |> PadModel.set_data!(pad_ref, :end_of_stream?, true)
       ~> {{:ok, :handle}, &1}
@@ -132,7 +135,11 @@ defmodule Membrane.Core.Element.EventController do
         {{:error, {:received_end_of_stream_through_output, pad_ref}}, state}
 
       %{end_of_stream?: true} ->
-        {{:error, {:end_of_stream_already_received, pad_ref}}, state}
+        warn("Ignoring event EndOfStream as it has already come before", state)
+        {{:ok, :ignore}, state}
+
+      %{state: playback_state} ->
+        {{:error, {:received_eos_in_incorrect_state, pad_ref, playback_state}}, state}
 
       %{start_of_stream?: false} ->
         {{:ok, :ignore}, state}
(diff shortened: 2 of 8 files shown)


lib/membrane/bin.ex
```diff
@@ -246,6 +246,12 @@ defmodule Membrane.Bin do
     |> noreply()
   end
 
+  @impl GenServer
+  def handle_info(Message.new(:handle_unlink, pad_ref), state) do
```
mat-hek: bin's children should be removed here

mat-hek:
closes #216

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -56,4 +58,12 @@ defmodule Membrane.Core.Parent.MessageDispatcher do
...
     |> noreply(state)
   end
+
+  defp inform_parent(state, msg, msg_params) do
+    if not pipeline?(state) and state.watcher,
+      do: Message.send(state.watcher, msg, [state.name | msg_params])
+  end
+
+  defp pipeline?(%Pipeline.State{}), do: true
+  defp pipeline?(_), do: false
```
mat-hek: maybe adding a `parent_type` field to the state would be helpful?

lib/membrane/core/element/event_controller.ex
```diff
@@ -132,7 +135,11 @@ defmodule Membrane.Core.Element.EventController do
         {{:error, {:received_end_of_stream_through_output, pad_ref}}, state}
 
       %{end_of_stream?: true} ->
-        {{:error, {:end_of_stream_already_received, pad_ref}}, state}
+        warn("Ignoring event EndOfStream as it has already come before", state)
+        {{:ok, :ignore}, state}
+
+      %{state: playback_state} ->
+        {{:error, {:received_eos_in_incorrect_state, pad_ref, playback_state}}, state}
```
mat-hek: This should never happen, we should raise here

lib/membrane/core/element/event_controller.ex
```diff
@@ -132,7 +135,11 @@ defmodule Membrane.Core.Element.EventController do
         {{:error, {:received_end_of_stream_through_output, pad_ref}}, state}
 
       %{end_of_stream?: true} ->
-        {{:error, {:end_of_stream_already_received, pad_ref}}, state}
+        warn("Ignoring event EndOfStream as it has already come before", state)
```
mat-hek: This should be at most a debug

lib/membrane/core/parent/message_dispatcher.ex
kanes115: `%Pipeline.State{parent_type: :pipeline}` <- redundant and unnecessary imo. Let's leverage pattern matching and elixir structs here.

kanes115:
@mat-hek also tests from `test/membrane/integration/bin_test.exs` cover some of this (especially the communication of EOS to parents). More tests are quite difficult to accomplish as this is a fix for RC.

lib/membrane/core/element/event_controller.ex
```diff
@@ -135,11 +134,11 @@ defmodule Membrane.Core.Element.EventController do
         {{:error, {:received_end_of_stream_through_output, pad_ref}}, state}
 
       %{end_of_stream?: true} ->
-        warn("Ignoring event EndOfStream as it has already come before", state)
+        debug("Ignoring event EndOfStream as it has already come before", state)
         {{:ok, :ignore}, state}
 
       %{state: playback_state} ->
-        {{:error, {:received_eos_in_incorrect_state, pad_ref, playback_state}}, state}
+        raise {:received_eos_in_incorrect_state, pad_ref, playback_state}
```
mat-hek: you cannot raise a tuple

mat-hek:
@kanes115 if they cover, why didn't they catch that? maybe they need a fix? I don't see any RC in the description of the PR

kanes115:
Yea, okey, so there's part of the issue that is RC (receiving duplicate EOS events) and not RC part (not generating EOS in the first place).

test/membrane/core/element/event_controller_test.exs
```diff
@@ -71,16 +73,20 @@ defmodule Membrane.Core.Element.EventControllerTest do
...
-        end)
-
-      state = %{state | pads: pads}
+      state = put_start_of_stream(state, :input)
 
       assert {:ok, state} = EventController.handle_event(:input, %Event.EndOfStream{}, state)
       assert state.pads.data.input.end_of_stream?
     end
+
+    test "end of stream is generated when playback state changes from :playing to :prepared", %{
```
mat-hek: this should rather be in lifecycle controller test, shouldn't it?

PR #221, opened 2019-11-29
Accept stop responses in genserver

kanes115:


lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -22,7 +24,13 @@ defmodule Membrane.Core.Parent.MessageDispatcher do
   end
 
   def handle_message(Message.new(:stop_and_terminate), state) do
-    LifecycleController.handle_stop_and_terminate(state)
+    case LifecycleController.handle_stop_and_terminate(state) do
+      {:ok, :stop} ->
+        {:stop, :normal, state}
+
+      e ->
```
mat-hek: - e suggests it's always an error here
- this can be fit in one line

lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -71,11 +71,11 @@ defmodule Membrane.Core.Parent.LifecycleController do
   end
 
   @spec handle_stop_and_terminate(state_t) ::
-          {:stop, :normal, state_t} | PlaybackHandler.handler_return_t()
+          PlaybackHandler.handler_return_t() | {:ok, :stop}
   def handle_stop_and_terminate(state) do
     case state.playback.state do
       :stopped ->
-        {:stop, :normal, state}
+        {:ok, :stop}
```
mat-hek: I would return the state anyway for consistency

lib/membrane/helper/genserver.ex
```diff
@@ -4,6 +4,11 @@ defmodule Membrane.Helper.GenServer do
   use Membrane.Log, tags: :core
 
   def noreply({:ok, new_state}), do: {:noreply, new_state}
+
+  def noreply({:stop, _reason, _new_state} = res), do: res
```
mat-hek: This is not needed now

lib/membrane/helper/genserver.ex
```diff
@@ -4,6 +4,11 @@ defmodule Membrane.Helper.GenServer do
...
+
+  def noreply({:stop, _reason, _new_state} = res, _old_state), do: res
```
mat-hek: This is not needed now

lib/membrane/core/element/message_dispatcher.ex
```diff
@@ -15,27 +15,38 @@ defmodule Membrane.Core.Element.MessageDispatcher do
...
   use Bunch
 
   @doc """
   Parses message incoming to element and forwards it to proper controller.
   """
   @spec handle_message(Message.t(), :info | :call | :other, State.t()) ::
-          State.stateful_try_t() | State.stateful_try_t(any)
+          State.stateful_try_t()
+          | State.stateful_try_t(any)
+          | Helper.GenServer.genserver_return_t()
```
mat-hek: shouldn't this return only `Helper.GenServer.genserver_return_t()` ?

lib/membrane/core/element/message_dispatcher.ex
```diff
@@ -15,27 +15,38 @@ defmodule Membrane.Core.Element.MessageDispatcher do
...
+      else
+        handle: {_error, {{:error, reason}, state}} ->
+          handle_message_error(message, mode, reason, state)
+
+        demands: {{:error, reason}, state} ->
+          handle_message_error(message, mode, reason, state)
+      end
+
+    case mode do
+      :info -> result |> noreply(state)
```
mat-hek: This makes it impossible to e.g. return stop from handle_info. I think this should be done in`do_handle_message`. And make sure you handle stop properly whenever needed

lib/membrane/core/element/message_dispatcher.ex
kanes115: Well, if it's invoked with mode `:other` we return the result not wrapped in genserver return tuples, so it still might return this stateful try.

lib/membrane/core/element/message_dispatcher.ex
kanes115: Well, it does. Actually nothing changes in this regard. You can just ignore :noreply tuple and return stop whenever needed. The whole design requires changes: the very fact that we pass information about if the message came from handle_cast or call to message dispatcher breaks the idea of layer architecture here. This makes the whole problem exist in the first place.

mat-hek: So if we're trying to fix anything here, let's make it look as it should, instead of complicating it further

kanes115:
Merging due to the fact that there are more serious flaws/refactor to be addressed described in #228.

PR #222, opened 2019-12-19
altering timer interval

### LINKED ISSUE #196: Process.send_after equivalent allowing to use Membrane.Clock
`start_timer` action is not enough when interval between ticks may change. The examples are `:ibm_speech_to_test` and `:gcloud_speech_to_text` that support FLAC files with variable frame size. They are using `Process.send_after` that is based on erlang time and the equivalent using `Membrane.Clock` is needed.

Ideas:
- `send_after` action
- `Membrane.Timer.send_after/5` wrapping `Process.send_after/4` and accepting clock pid

cc @mat-hek

mat-hek:
todo:
- [x] support for changing timer interval (closes #196)
- [x] support for delaying setting the interval
- [x] Docs and specs

### DIFF
--- a/lib/membrane/element/action.ex
+++ b/lib/membrane/element/action.ex
@@ -10,7 +10,7 @@ defmodule Membrane.Element.Action do
   do not return any actions) unless explicitly stated otherwise.
   """
 
-  alias Membrane.{Buffer, Caps, Clock, Event, Notification, Time}
+  alias Membrane.{Buffer, Caps, Clock, Event, Notification}
   alias Membrane.Pad
 
   @typedoc """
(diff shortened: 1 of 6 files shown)


lib/membrane/element/action.ex
```diff
@@ -161,23 +161,46 @@ defmodule Membrane.Element.Action do
...
+
+  Permitted only from `c:Membrane.Element.Base.handle_tick/3`, unless the interval
+  was previously set to `:no_interval`.
+
+  If the `interval` is `:no_interval`, the timer won't issue any ticks until
+  another `t:set_interval_t/0` action. Otherwise, the timer will issue ticks every
+  new `interval`. The next tick after interval change is scheduled at
+  `new_interval + previous_time`, where previous_time is the time of the latest
+  tick or the time of returning `t:start_timer_t/0` action if no tick has been
+  sent yet.
```
bblaszkow06: If I understand correctly this means you can have a burst of ticks trying to catch up with the amount of ticks that would've been sent if the new interval was set exactly at the time of last tick? You can mention it here

PR #223, opened 2020-01-08
Refactor/time

mat-hek:


lib/membrane/time.ex
```diff
@@ -11,67 +11,26 @@ defmodule Membrane.Time do
...
-  }
+
+  @units [
+    %{plural: :days, singular: :day, abbrev: "d", duration: 86_400_000_000_000},
+    %{plural: :hours, singular: :hour, abbrev: "h", duration: 3_600_000_000_000},
+    %{plural: :minutes, singular: :minute, abbrev: "min", duration: 60_000_000_000},
+    %{plural: :seconds, singular: :second, abbrev: "s", duration: 1_000_000_000},
+    %{plural: :milliseconds, singular: :millisecond, abbrev: "ms", duration: 1_000_000},
+    %{plural: :microseconds, singular: :microsecond, abbrev: "us", duration: 1_000},
+    %{plural: :nanoseconds, singular: :nanosecond, abbrev: "ns", duration: 1}
```
kanes115: Would it be possible to express those numbers as multiplications? Like `60 * 60 * 1000 * 1_000 * 1_000` for hours? (to go through each unit)

lib/membrane/time.ex
```diff
@@ -11,67 +11,26 @@ defmodule Membrane.Time do
...
+  ]
 
   @doc """
   Checks whether value is Membrane.Time.t
   """
   defguard is_t(value) when is_integer(value)
```
kanes115: Since it's a public guard, I would go for a full `is_time`. It's short anyway.

lib/membrane/time.ex
```diff
@@ -206,281 +153,105 @@ defmodule Membrane.Time do
...
+    def unquote(as_fun_name)(time) when is_t(time) do
+      Ratio./(time, unquote(unit.duration))
+    end
+  end)
 
   defp best_unit(time) do
-    {unit, divisor} = units() |> Enum.find(fn {_unit, divisor} -> time |> rem(divisor) == 0 end)
-
-    {time |> div(divisor), unit}
+    unit = @units |> Enum.find(&(rem(time, &1.duration) == 0))
```
kanes115: I know it shouldn't happen but what if it's not found? Maybe we should check that in guard `is_t`? That the integer is at least convertible to nanoseconds?

mat-hek: It is always found since the time is represented in nanoseconds (duration of a nanosecond is 1)

kanes115: I know, that's what I meant by `I know it shouldn't happen but...`. I just wonder why we don't check that in `is_time` guard

lib/membrane/time.ex
kanes115: bump @mat-hek

lib/membrane/time.ex
mat-hek: We do. Each integer satisfies that.

lib/membrane/time.ex
mat-hek: why?

PR #237, opened 2020-03-13
Fix shutting down elements not prepared for termination

kanes115:
Fixed problems:
* Children did not inherit the `terminating?` state of their parents. This is resolved by adding `terminating` state
* Make parent always wait for its children to die before it decides to die
* Refactor playback handler and get rid of playbackable concept

Rules of stopping parents and children (from now on - after this PR):
* We move every child to a new `terminating` state which is an indication that a child is ready to be shutdown.
* Children commit commit suicide directly after going into `terminating` state.
* Also when child is removed by action `remove_child`, parent puts a monitor on the child, tells it to go to terminating state, then the child kills itself and parent can safely remove this child from its children list (due to `DOWN` message)


lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -50,42 +50,50 @@ defmodule Membrane.Core.Parent.LifecycleController do
...
-        state = %{state | terminating?: true}
-
-        PlaybackHandler.change_and_lock_playback_state(
-          :stopped,
-          Parent.LifecycleController,
-          state
-        )
-    end
+  def handle_terminate(state) do
+    PlaybackHandler.change_and_lock_playback_state(
```
mat-hek, line 92: change_and_lock_playback_state is now used only for :terminating state, let's handle that automatically in  change_playback_state

lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -128,7 +130,7 @@ defmodule Membrane.Core.Parent.LifecycleController do
           PlaybackHandler.handler_return_t()
   def child_playback_changed(
         _pid,
-        _new_playback_state,
+        # new_playback_state,
```
mat-hek, line 133: ?

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -26,26 +26,17 @@ defmodule Membrane.Core.Parent.MessageDispatcher do
...
 
-  def handle_message(Message.new(:stop_and_terminate), state) do
-    case LifecycleController.handle_stop_and_terminate(state) do
-      {{:ok, :stop}, state} ->
-        {:stop, :normal, state}
-
-      result ->
-        result |> noreply(state)
-    end
+  def handle_message(Message.new(:terminate), state) do
```
mat-hek, line 29: Since we have 'terminating' playback state, let's use regular playback change messages for that

lib/membrane/core/playback_handler.ex
```diff
@@ -72,10 +72,12 @@ defmodule Membrane.Core.PlaybackHandler do
           | :handle_prepared_to_playing
           | :handle_prepared_to_stopped
           | :handle_stopped_to_prepared
+          | :handle_stopped_to_terminating
   def state_change_callback(:stopped, :prepared), do: :handle_stopped_to_prepared
   def state_change_callback(:playing, :prepared), do: :handle_playing_to_prepared
   def state_change_callback(:prepared, :playing), do: :handle_prepared_to_playing
   def state_change_callback(:prepared, :stopped), do: :handle_prepared_to_stopped
+  def state_change_callback(:stopped, :terminating), do: :handle_stopped_to_terminating
```
mat-hek, line 80: we already have handle_shutdown, so this seems redundant

mat-hek:
> Consider refactoring playback handler so that it does not use callbacks(?) (I feel like this should be in a separate PR)

Let's move that to a separate PR then.

lib/membrane/core/element/lifecycle_controller.ex
```diff
@@ -128,32 +128,23 @@ defmodule Membrane.Core.Element.LifecycleController do
...
-      PlaybackHandler.change_and_lock_playback_state(:stopped, __MODULE__, state)
-    end
+  @spec terminate(State.t()) :: State.stateful_try_t()
+  def terminate(state, opts \\ []) do
+    kill_after? = Keyword.get(opts, :kill_after?, false)
+
+    result = PlaybackHandler.change_and_lock_playback_state(:terminating, __MODULE__, state)
+
+    if kill_after?,
+      do: {{:ok, {:stop, :normal, state}}, state},
```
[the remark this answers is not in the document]
kanes115: I was about to change it after I check if it’s really possible. Sorry for not making it draft, but the PR is not ready yet. Good point though.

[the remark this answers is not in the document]
mat-hek: no problem, but next time just don't add reviewers till the PR is ready for review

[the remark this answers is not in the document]
kanes115: Yea, reviewers were added before the last review, before we decided on refactoring playback. Sure!

lib/membrane/core/parent/message_dispatcher.ex
kanes115: We do use them! This terminate is only for pipeline (root of the tree, command to initiate terminating). It replaces old message `stop_and_terminate`. I can understand why it's confusing - that would be nice to have a clear message API of pipeline/bin/element/parent/child, separated. This might be achieved after we get rid of message dispatchers.

mat-hek: IMHO we don't need this message even for pipeline

lib/membrane/core/playback_handler.ex
kanes115: Yea, but won't it be unintuitive to not have this callback? It will also introduce some ugly ifs in core code.

kanes115:
Waits for #247 before merging.

mat-hek:
https://github.com/membraneframework/membrane-core/pull/237/files#diff-28bbc0033337291363deb68ea4f25caaL73 this should never happen now, we can raise there

lib/membrane/core/element/message_dispatcher.ex
```diff
@@ -65,8 +74,18 @@ defmodule Membrane.Core.Element.MessageDispatcher do
...
+  defp do_handle_message(
+         Message.new(:change_playback_state, new_playback_state),
+         :info,
+         state
+       ) do
+    with {:stop, _reason, _state} = stop_tuple <-
+           PlaybackHandler.change_playback_state(new_playback_state, LifecycleController, state) do
+      # TODO get rid of this hack when we resign from
+      # Bunch.stateful_try_with and family. This hack unwraps
+      # gensever :stop tuple.
```
bblaszkow06, line 86: I don't think leaving just TODOs in code is enough if we're going to merge them. I would suggest creating an issue with more detailed description and link it here.
Applies to all TODOs in this PR

lib/membrane/core/playback_handler.ex
bblaszkow06: > won't it be unintuitive to not have this callback

I've just taken a quick look at this PR and it seems that `:terminate` playback state is just an implementation detail, not a valid state that the user should know of. The only difference I see between `:handle_shutdown` and this callback (from user perspective ofc) is context. I would unify them and rather use a different name than other playback state change callbacks.

mat-hek: There is one more difference: stopped_to_terminating is now called only upon request to change the playback state, while handle_shutdown - within GenServer's terminate callback, and imho the latter is better. We may rename `handle_shutdown` to `handle_terminate` and add the context, but it's rather out of this PR

lib/membrane/core/playback_handler.ex
```diff
@@ -84,112 +87,100 @@ defmodule Membrane.Core.PlaybackHandler do
...
+
+    handler_res =
+      handler.handle_playback_state_changed(
+        old_playback.state,
+        old_playback.pending_state,
+        state
+      )
+
+    case handler_res do
+      {:stop, _reason, state} = stop_tuple ->
```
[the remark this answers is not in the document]
kanes115: Why do you think it's a hack? 😨 It is simply a possibility of returning stop tuple in playback callback.

lib/membrane/core/element/message_dispatcher.ex
kanes115: I'm all for the idea of creating issues and referencing them in TODO comment

PR #244, opened 2020-03-31
avoid creating unnecessary log messages

mat-hek:


lib/membrane/log.ex
```diff
@@ -111,7 +111,12 @@ defmodule Membrane.Log do
         end
 
       true ->
-        quote location: :keep, bind_quoted: [message: message, tags: tags] do
+        quote location: :keep do
+          if false do
+            _ = unquote(message)
+            _ = unquote(tags)
+          end
```
bblaszkow06: Clever, but hacky. I think it deserves a comment as I'm afraid it might break in the future, i.e. start generating warnings

PR #247, opened 2020-04-16
Replace Playbackable calls

kanes115:


lib/membrane/core/state.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Membrane.Core.State do
```
mat-hek, line 1: moduledoc false

mat-hek: and you can merge ;)

PR #248, opened 2020-04-20
Add NTP timestamp conversion to Membrane.Time

bblaszkow06:
NTP timestamps are used by RTCP Sender report


lib/membrane/time.ex
```diff
@@ -179,6 +182,41 @@ defmodule Membrane.Time do
...
+  NTP timestamp uses fixed point representation with the integer part in the first 32 bits
+  and the fractional part in the last 32 bits.
+  """
+  @spec from_ntp_timestamp(ntp_time :: <<_::64>>) :: t()
+  def from_ntp_timestamp(<<ntp_seconds::32, ntp_fraction::32>>) do
+    fractional =
+      ntp_fraction
+      |> Ratio.new(Ratio.pow(2, 32))
+      |> Ratio.mult(second())
+      |> Ratio.trunc()
```
bblaszkow06, line 198: I have figured out another way to do this:
More precise and possibly faster,  but hacky. Do you think it's too hacky to include @mat-hek?
```suggestion
    <<fraction_ieee::float-size(64)>> = <<0::1, 1023::11, ntp_fraction::32, 0::20>>
    fractional = trunc((fraction_ieee - 1) * second())

```

lib/membrane/time.ex
```diff
@@ -179,6 +184,41 @@ defmodule Membrane.Time do
...
+  NTP timestamp uses fixed point representation with the integer part in the first 32 bits
+  and the fractional part in the last 32 bits.
+  """
+  @spec from_ntp_timestamp(ntp_time :: <<_::64>>) :: t()
+  def from_ntp_timestamp(<<ntp_seconds::32, ntp_fraction::32>>) do
+    fractional =
+      ntp_fraction
+      |> Ratio.new(@two_to_pow_32)
+      |> Ratio.mult(second())
+      |> Ratio.trunc()
```
bblaszkow06, line 200: I have figured out another way to do this:
More precise and possibly faster,  but hacky. Do you think it's too hacky to include @mat-hek?
```suggestion
    <<fraction_ieee::float-size(64)>> = <<0::1, 1023::11, ntp_fraction::32, 0::20>>
    fractional = trunc((fraction_ieee - 1) * second())
```

mat-hek: Imho seems not worth it, but I don't insist. However, isn't `div(ntp_fraction*second, @two_to_pow_32)` equivalent?

lib/membrane/time.ex
```diff
@@ -179,6 +184,41 @@ defmodule Membrane.Time do
...
+  Converts the timestamp into NTP timestamp. May introduce small rounding errors.
+  """
+  @spec to_ntp_timestamp(timestamp :: t()) :: <<_::64>>
+  def to_ntp_timestamp(timestamp) do
+    ts_as_ratio = timestamp |> as_seconds()
+    seconds = ts_as_ratio |> Ratio.trunc()
+    ntp_seconds = seconds + @ntp_unix_epoch_diff
+
+    fraction = ts_as_ratio |> Ratio.sub(seconds)
+    fixed_point_fraction = fraction |> Ratio.mult(@two_to_pow_32) |> Ratio.trunc()
```
mat-hek, line 217: I believe getting rid of ratio here would make it both more readable and faster.

PR #249, opened 2020-04-22
Log

mat-hek:
- [x] switch to elixir logger
- [x] add verbose debugs
- [x] add log_metadata action

should be merged after #237


lib/membrane/core/element/action_handler.ex
```diff
@@ -304,13 +299,10 @@ defmodule Membrane.Core.Element.ActionHandler do
...
   @spec send_caps(Pad.ref_t(), Caps.t(), State.t()) :: State.stateful_try_t()
   def send_caps(pad_ref, caps, state) do
-    debug(
-      """
-      Sending caps through pad #{inspect(pad_ref)}
-      Caps: #{inspect(caps)}
-      """,
-      state
-    )
+    Logger.debug("""
```
bblaszkow06, line 302: Why it's not verbose like others?

lib/membrane/core/element/playback_buffer.ex
```diff
@@ -151,13 +150,10 @@ defmodule Membrane.Core.Element.PlaybackBuffer do
...
     PadModel.assert_data!(state, pad_ref, %{direction: :input})
 
-    debug(
-      """
-      Received caps on pad #{inspect(pad_ref)}
-      Caps: #{inspect(caps)}
-      """,
-      state
-    )
+    Logger.debug("""
```
bblaszkow06, line 153: Why it's not verbose like others?

lib/membrane/core/element/timer_controller.ex
```diff
@@ -66,11 +65,8 @@ defmodule Membrane.Core.Element.TimerController do
       |> Bunch.Access.update_in([:synchronization, :timers, timer_id], &Timer.tick/1)
       ~> {:ok, &1}
     else
-      {{:error, reason}, state} ->
-        warn_error("Error while handling tick", reason, state)
-
-      false ->
-        {:ok, state}
+      {{:error, reason}, state} -> {{:error, reason}, state}
```
bblaszkow06, line 68: How about
```suggestion
      {{:error, _reason}, _state} = err -> err
```

lib/membrane/core/parent/child_life_controller/startup_handler.ex
```diff
@@ -123,9 +124,9 @@ defmodule Membrane.Core.Parent.ChildLifeController.StartupHandler do
     end
   end
 
-  defp start_child(child, parent_clock, syncs) do
+  defp start_child(child, parent_clock, syncs, log_metadata) do
     %ChildEntry{name: name, module: module, options: options} = child
-    debug("Starting child: name: #{inspect(name)}, module: #{inspect(module)}")
+    Logger.debug("Starting child: name: #{inspect(name)}, module: #{inspect(module)}")
```
bblaszkow06, line 129: I think that one log when starting an element could get `info` level. It should be either this or the one showing `Element start_link` (the latter includes element options)

lib/membrane/log.ex
```diff
@@ -1,4 +1,5 @@
 defmodule Membrane.Log do
+  @moduledoc deprecated: "Use Elixir `Logger` or `Membrane.Logger` instead"
```
bblaszkow06, line 2: If we hide `Membrane.Logger` this will have to be changed

lib/membrane/pipeline.ex
```diff
@@ -176,6 +165,7 @@ defmodule Membrane.Pipeline do
   end
 
   def init({module, pipeline_options}) do
+    Logger.metadata(mb_name: :pipeline)
```
bblaszkow06, line 168: Meh, I would either use module name or "pipeline #{pid}"

spec/membrane/core/input_buffer_spec.exs
```diff
@@ -14,7 +14,7 @@ defmodule Membrane.Core.InputBufferSpec do
   end
 
   describe ".init/6" do
-    let :name, do: :name
+    let :log_tag, do: "test"
```
bblaszkow06, line 17: I don't see a test for the `raise` you have added, so we never tested the toilet overflow. We may fix it now :wink:

lib/membrane/core/element/playback_buffer.ex
```diff
@@ -74,7 +76,7 @@ defmodule Membrane.Core.Element.PlaybackBuffer do
   """
   @spec eval(State.t()) :: State.stateful_try_t()
   def eval(%State{playback: %Playback{state: :playing}} = state) do
-    debug("evaluating playback buffer", state)
+    Logger.debug("Evaluating playback buffer")
```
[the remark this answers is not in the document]
mat-hek: I believe deciding which logs are needed is far beyond this PR. Also, `debug_verbose` is intended for logs that are usually displayed too often, not anything that may be useful in some cases

lib/membrane/core/element/action_handler.ex
mat-hek: Caps shouldn't be sent too often, and if they are it's important to see that

lib/membrane/core/element/playback_buffer.ex
mat-hek: as above

lib/membrane/core/parent/child_life_controller/startup_handler.ex
mat-hek: Why?

lib/membrane/core/element/action_handler.ex
```diff
@@ -3,29 +3,22 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+  require Logger
+  require Membrane.Core.Child.PadModel
+  require Membrane.Core.Message
+  require Membrane.Logger
+
+  alias Membrane.{ActionError, Buffer, Caps, CallbackError, Event, Notification, Pad}
+  alias Membrane.Core.Element.{DemandHandler, LifecycleController, State, TimerController}
+  alias Membrane.Core.{Message, PlaybackHandler}
+  alias Membrane.Core.Child.PadModel
+  alias Membrane.Element.Action
```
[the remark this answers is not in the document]
mat-hek: it's this convention: https://github.com/christopheradams/elixir_style_guide#module-attribute-ordering and it seems credo can do it https://github.com/rrrene/credo-proposals/issues/22 but haven't used that

lib/membrane/logger.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  Debugs done via this macro are displayed only when verbosity is set to `:verbose`
+  via `Mix.Config`:
+
+      use Mix.Config
+      config :membrane_core, :logger, verbosity: :verbose
+
+  while the default verbosity is `:normal`.
+
+  A `mb_verbosity: :verbose` metadata entry is added to each debug.
```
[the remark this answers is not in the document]
mat-hek: I wanted to design it in an extensible way. `verbose_logging: true` is not really extensible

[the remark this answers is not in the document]
mat-hek: I also can imagine a verbose `warn`, for example

lib/membrane/logger.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Membrane.Logger do
+  @moduledoc """
+  Logging utils.
+  """
```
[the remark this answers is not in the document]
mat-hek: > Do you intend to allow users to use this?
yes, from within callbacks like `handle_demand` or `handle_process`

lib/membrane/bin.ex
```diff
@@ -174,7 +172,9 @@ defmodule Membrane.Bin do
   end
 
   @impl GenServer
-  def init({my_name, module, bin_options}) do
+  def init({name, module, bin_options, log_metadata}) do
+    Logger.metadata([mb_name: inspect(name)] ++ log_metadata)
```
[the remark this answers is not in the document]
mat-hek: I believe implicitly changing the configuration of the logger, especially of a particular backend is really counterintuitive.

lib/membrane/core/child/pad_controller.ex
```diff
@@ -279,17 +278,17 @@ defmodule Membrane.Core.Child.PadController do
...
 
     Message.send(pid, :demand_unit, [demand_unit, other_ref])
 
     input_buf =
       InputBuffer.init(
-        state.name,
         demand_unit,
         pid,
         other_ref,
+        inspect(ref),
```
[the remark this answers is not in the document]
mat-hek: That's why I thought of prepending log messages with the element name. That, however, would require sticking to our wrapper

lib/membrane/logger.ex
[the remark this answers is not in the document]
bblaszkow06: I don't think that extensibility is needed here. Verbose `warn` could be achieved with this flag as well - the only difference would be that you could not enable debug on Elixir logger and verbose logging on warn only.
However, I don't think such option (verbose warn) should be ever implemented. It's not normal to have 2 different settings for logger level. This config above is rather a workaround for missing log level `silly` or similar. If something is considered a warning, it should not be muted (unless you set log level to `:error` but I can't imagine why anyone would do that) and if it's spamming too much then something is wrong with that logger call in the first place: either it should not be warn or the code should be changed to limit the number of messages (e.g. we would like to show it only once and hide repeating entries)

lib/membrane/bin.ex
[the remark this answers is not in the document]
bblaszkow06: IMHO having mostly useless logs by default is much bigger problem. I would like to emphasize that we would only change the default behaviour (i.e. when nothing was set for that key). That way the users just poking around with our framework will see something meaningful and if after modifying configs our metadata will disappear, they will start looking on how to configure logging in Membrane. With current situation they will most likely never see the that part of docs and will have no idea that something can be logged.
If that's completely out of option for you, then you need to come up with some other solution because such user experience is not acceptable.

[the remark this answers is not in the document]
mat-hek: > we would only change the default behaviour (i.e. when nothing was set for that key)
- it's going to stop working once somebody configures metadata for a possibly unrelated thing
- it may result in RC when another library goes the same way
- there's no public API to check for that afaik

What about checking if `mb_name` is present in the all backends' configurations, and if it's not, prepending it to the message? That would
- still require using private API (but since it's `Application.get_env(:logger, :backends)` and `Application.get_env(:logger, backend, :metadata)` it's rather not going to change)
- still require having own wrapper, but even if one doesn't use the wrapper, names are going to be displayed if metadata is configured properly

Another approach would be just to `warn` if the check doesn't pass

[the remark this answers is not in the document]
bblaszkow06: >it's going to stop working once somebody configures metadata for a possibly unrelated thing

Already addressed that:
>if after modifying configs our metadata will disappear, they will start looking on how to configure logging in Membrane

And as I mentioned - it's a problem, but still better than not useless logs & having no clue it can be changed

> there's no public API to check for that afaik

Of course there is, even you used it few lines below (`Application.get_env(:logger, :console)` will return `nil` or whatever you pass as 3rd arg if there's no such config, so `Application.get_env(:logger, :console, []) |> Keyword.get(:metadata) != nil` will say if something was set in config). Config keys are not private API, they are [documented](https://hexdocs.pm/logger/Logger.html#module-console-backend) and any change there would be a breaking change of the API

Anyway, the RC is a problem and I agree it's hacky. It seems that getting back to wrapper might be the only reasonable solution.
Regarding your idea:
 * there's no guarantee any backend besides `:console` has `:metadata` config
* If we have the wrapper I don't see the point of tinkering with `:logger` config and would provide our own configuration of shown (i.e. prepended to the message) keys. The metadata would still have to contain them to provide filtering via `:logger`

lib/membrane/core/parent/child_life_controller/startup_handler.ex
bblaszkow06: Because
* it doesn't appear too often,
* provides info about what's going on in the pipeline(s) even if we're not debugging
* we don't have any `info` logs.

I won't insist on having it here, but in general, providing more granularity of log levels would be better

lib/membrane/logger.ex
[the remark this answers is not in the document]
bblaszkow06: OK, however mentioning the configuration in a separate page (and/or README) could be useful.

lib/membrane/bin.ex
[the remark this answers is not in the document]
mat-hek: > Of course there is, even you used it few lines below

The fact that configuration related functions are public doesn't mean that getting that configuration will give you the actual configuration of the logger. There exist `Logger.configure` and `Logger.configure_backend` and they don't state they alter the 'public' Logger configuration (but I checked and they do that)

> there's no guarantee any backend besides :console has :metadata config

yes, but
> It is recommended that backends support at least the following configuration options:
> :level - the logging level for that backend
> :format - the logging format for that backend
> :metadata - the metadata to include in that backend

> If we have the wrapper I don't see the point of tinkering with :logger config and would provide our own configuration of shown (i.e. prepended to the message) keys

The only benefit would be that when one wants to display that metadata, it won't be displayed twice. But I agree configuring that manually may be a better option.

> The metadata would still have to contain them to provide filtering via :logger

I don't see any option to filter logs by metadata other than in the compile-time, so it won't work. However one may want to display the metadata when not using our wrapper.

lib/membrane/core/parent/child_life_controller/startup_handler.ex
mat-hek: > it doesn't appear too often
Don't agree. It can appear very often especially if you have multiple pipelines
> provides info about what's going on in the pipeline
Maybe we can `info` about pipeline playback changes like gstreamer does?

lib/membrane/bin.ex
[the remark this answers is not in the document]
bblaszkow06: https://hexdocs.pm/logger/Logger.html#configure/1 says:
> See the "Runtime Configuration" section in the Logger module documentation for the available options. **The changes done here are automatically persisted to the :logger application environment**.

>I don't see any option to filter logs by metadata other than in the compile-time, so it won't work. However one may want to display the metadata when not using our wrapper.
You're right, the filtering would have to be done on our side, that's another argument for using a wrapper.

lib/membrane/core/element/playback_buffer.ex
```diff
@@ -122,12 +123,10 @@ defmodule Membrane.Core.Element.PlaybackBuffer do
...
-    debug(
-      ["
-      Received buffers on pad #{inspect(pad_ref)}
-      Buffers: ", Buffer.print(buffers)],
-      state
-    )
+    Membrane.Logger.debug_verbose("""
+    Received buffers on pad #{inspect(pad_ref)}
+    Buffers: #{inspect(buffers)}
+    """)
```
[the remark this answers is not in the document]
mat-hek: it won't help, as each buffer would have to be inspected anyway

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
+        :ok
+      end
+    end
+  end
+
+  Enum.map([:debug, :info, :warn, :error], fn method ->
+    @doc """
+    Wrapper around `Logger.#{method}/2` that adds Membrane prefix.
+
+    For details, see the 'prefixes' section of the moduledoc.
```
bblaszkow06, line 85:
```suggestion
    For details, see the ['prefixes'](#module-prefixes) section of the moduledoc.
```

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
+        require Logger
+        Logger.unquote(method)(unquote(prepend_prefix(message)), unquote(metadata))
+      end
     end
+  end)
+
+  @doc """
+  Wrapper around `Logger.log/3` that adds Membrane prefix.
+
+  For details, see the 'prefixes' section of the moduledoc.
```
bblaszkow06, line 100:
```suggestion
  For details, see the ['prefixes'](#module-prefixes) section of the moduledoc.
```

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
+          case message.() do
+            {content, metadata} -> {[unquote(@prefix_getter), message], metadata}
+            content -> [unquote(@prefix_getter), content]
+          end
+        end
+      else
+        [unquote(@prefix_getter), message]
+      end
+    end
+  end
```
bblaszkow06, line 55: Why is this a macro? Couldn't that be the body of the last `prepend_prefix` function clause?

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
+                  end)
+
+  defmacrop prefix_prepender(message) do
+    quote do
+      message = unquote(message)
+
+      if is_function(message, 0) do
+        fn ->
+          case message.() do
+            {content, metadata} -> {[unquote(@prefix_getter), message], metadata}
```
bblaszkow06, line 47: `unquote(@prefix_getter)` is used in every possible branch, it should be outside the `if`

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
+  """
 
-  while the default verbosity is `:normal`.
+  # TODO: compile_env should be always used when we require Elixir 1.10
+  @config if function_exported?(Application, :compile_env, 3),
+            do: Application.compile_env(:membrane_core, :logger, []),
+            else: Application.get_env(:membrane_core, :logger, [])
+
+  @prefix_getter (quote do
+                    Process.get(:membrane_logger_prefix, "")
```
bblaszkow06, line 37: What is the point of using our custom Process dict entry? Can't we use Logger metadata?

lib/membrane/logger.ex
```diff
@@ -1,46 +1,178 @@
...
-  Macro for verbose debug logs that should be silenced by default.
+  By default, this wrapper prepends each log with a prefix containig the context
+  of the log, such as element name. This can be turned off via configuration:
+
+      use Mix.Config
+      config :membrane_core, :logger, prefix: false
 
-  Debugs done via this macro are displayed only when verbosity is set to `:verbose`
-  via `Mix.Config`:
+  Regardless of the config, the prefix is passed to `Logger` metadata under `:mb` key.
```
bblaszkow06, line 13: `:mb` would be OK if we stored a map with different Membrane metadata keys. If it's only prefix I think it should be `:mb_prefix`

test/membrane/core/input_buffer_test.exs
```diff
@@ -0,0 +1,37 @@
...
+        |> InputBuffer.store(bufs(20))
+      end
+    ]
+    |> Enum.map(fn f -> assert_raise RuntimeError, ~r/Toilet overflow.*/, f end)
+  end
+
+  test "no toilet overflow" do
+    toilet(10) |> InputBuffer.store(bufs(10))
+    toilet(10) |> InputBuffer.store(bufs(7)) |> InputBuffer.store(bufs(3))
+    assert true
```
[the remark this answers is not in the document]
mat-hek: it's here to indicate that the test passes when nothing raises

lib/membrane/logger.ex
mat-hek: IMHO it's better for our stuff not to depend on the logger metadata

lib/membrane/logger.ex
mat-hek: wanted to keep it as short as possible in case it was printed

bblaszkow06: But it's not going to be printed by accident, right? Either someone explicitly states it should be printed or uses `metadata: :all` and will get a lot of entries among which it which a few letters more won't make a big difference when it comes to overall length but might improve interpretation of the printed metadata

lib/membrane/logger.ex
```diff
@@ -0,0 +1,178 @@
...
+            content -> [prefix, content]
+          end
+        end
+      else
+        [prefix, message]
+      end
+    end
+  end
+
+  defmacrop dynamically_prepend_prefix_macro(message), do: dynamically_prepend_prefix(message)
```
bblaszkow06, line 55: These macro and AST returning function + `prepend_prefix` function is still a hard to understand.
* I'd rename `prepend_prefix` to `prepended_prefix_ast` or similar
* `dynamically_prepend_prefix` -> `runtime_prepend_prefix_ast`
* `dynamically_prepend_prefix_macro` -> `runtime_prepend_prefix` or `prepend_prefix`
You may adjust the names, but I think you get the idea - let's not name the functions returning AST as they would be normal functions.

lib/membrane/logger.ex
mat-hek: For example, we may want to differentiate the metadata entry from the prepended prefix. Currently, the prepended prefix is wrapped in square braces and space is appended to it.

bblaszkow06: OK, it doesn't make much sense right now (I'm not really convinced that 4 additional characters justify another entry in process dict), however interpreting this entry as a cache for string representation of Membrane metadata does (e.g. if we wanted to store a list of prefixes in Logger.metadata and just a string here)

PR #252, opened 2020-04-24
Fix RC in parent termination

kanes115:


lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -149,6 +135,16 @@ defmodule Membrane.Core.Parent.LifecycleController do
     {:ok, %{state | children: new_children}}
   end
 
+  defp finish_pids_transition(state, pid) do
+    {{:ok, child}, state} = Parent.ChildrenModel.pop_child_by(state, :pid, pid)
```
mat-hek, line 139: why do you pop child when it goes to any playback state? shouldn't you just remove the flag?

lib/membrane/core/parent/child_entry.ex
```diff
@@ -18,10 +19,11 @@ defmodule Membrane.Core.Parent.ChildEntry do
           options: struct | nil,
           pid: pid,
           clock: Clock.t(),
-          sync: Sync.t()
+          sync: Sync.t(),
+          in_playback_transition?: boolean()
```
mat-hek, line 23: don't believe it's a good name... children being dynamically removed are also in playback transition, and they are not flagged unless the pipeline changes its playback state

lib/membrane/core/parent/children_model.ex
```diff
@@ -34,6 +34,19 @@ defmodule Membrane.Core.Parent.ChildrenModel do
     end
   end
 
+  @spec pop_child_by(State.t(), atom(), any()) ::
+          {{:ok, child_data :: map() | nil}, State.t()}
+  def pop_child_by(%{children: children} = state, key, val) do
```
mat-hek, line 39: imho it's too specific

lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -109,35 +108,22 @@ defmodule Membrane.Core.Parent.LifecycleController do
...
 
+  defp no_child_in_transition?(state),
+    do: Parent.ChildrenModel.all?(state, &(not &1.in_playback_transition?))
+
+  defp transition_finished?(pending_state, new_state)
+  defp transition_finished?(pb_state, pb_state), do: true
+  defp transition_finished?(_, _), do: false
+
   # Child was removed
   def handle_child_death(pid, :normal, state) do
```
mat-hek, line 128: we discussed that child should be unflagged here and playback change continuation conditions should be checked

lib/membrane/core/parent/children_model.ex
kanes115: imo ChildrenModel module is not a very good approach. I believe it is good to be able to pattern match on all structures instead of having these object-oriented-like "getters". Anyway, I agree this function can be removed.

lib/membrane/core/parent/lifecycle_controller.ex
kanes115: I agree continue conditions should be checked but don't we also want to remove the child?

PR #254, opened 2020-06-02
Corrected documentation for `Membrane.Testing.Source`.

Hajto:
I corrected errors in documentation for Membrane Testing Source and expanded a bit.

### DIFF
--- a/lib/membrane/testing/source.ex
+++ b/lib/membrane/testing/source.ex
@@ -1,25 +1,31 @@
 defmodule Membrane.Testing.Source do
   @moduledoc """
-  Testing Element for supplying data based on generator function passed through
-  options.
+  Testing Element for supplying data based on generator function or payloads passed
+  through options.
 
   ## Example usage
 
   As mentioned earlier you can use this element in one of two ways, providing
   either a generator function or an `Enumerable.t`.
 
+  If you provide an `Enumerable.t` with payloads, then each of those payloads will
+  be wrapped in a `Membrane.Buffer` and sent through `:output` pad.
   ```
-  %Sink{output: [0xA1, 0xB2, 0xC3, 0xD4]}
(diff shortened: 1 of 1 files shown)


lib/membrane/testing/source.ex
```diff
@@ -1,25 +1,31 @@
...
-  %Sink{output: [0xA1, 0xB2, 0xC3, 0xD4]}
+  %Source{output: [0xA1, 0xB2, 0xC3, 0xD4]}
   ```
 
-  In order to specify `Membrane.Testing.Sink` with generator function you need
-  to provide initial state and function that matches `t:generator/0` type.
+  In order to specify `Membrane.Testing.Source` with generator function you need
+  to provide initial state and function that matches `t:generator/0` type. This
+  function should take state and demand size as its' arguments and return
+  a tuple consisting of actions that element will return during the
```
Hajto, line 20: I added this part to emphasize to possible users that they can do whatever they want that is possible via `handle_demand/5` callback.

lib/membrane/testing/source.ex
```diff
@@ -1,25 +1,31 @@
...
   ```
-  %Sink{output: [0xA1, 0xB2, 0xC3, 0xD4]}
+  %Source{output: [0xA1, 0xB2, 0xC3, 0xD4]}
   ```
 
-  In order to specify `Membrane.Testing.Sink` with generator function you need
-  to provide initial state and function that matches `t:generator/0` type.
+  In order to specify `Membrane.Testing.Source` with generator function you need
+  to provide initial state and function that matches `t:generator/0` type. This
+  function should take state and demand size as its' arguments and return
```
mat-hek, line 19:
```suggestion
  function should take state and demand size as its arguments and return
```

PR #255, opened 2020-06-04
mention ERL-1273

mkaput:


mkaput:
Erlang devs confirmed the issue and have some clues on how to fix, maybe let's wait with this message for a day or two as they may promptly release 23.0.3 with fixes

mat-hek:
btw, I'd mention that it only applies to elements that have native components

mkaput:
OTP PR has been merged so I expect the new release to be soon available

PR #265, opened 2020-07-28
Credo linting

bblaszkow06:
Circle now uses 2 parallel jobs, one builds in dev env, the other in test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - run: mix deps.get
       - run: mix format --check-formatted
       - run: mix compile --force --warnings-as-errors
-      - run: mix credo
+      - run: MIX_ENV=dev mix credo
       - run: mix test --include long_running
       - run: mix espec
       - run: MIX_ENV=dev mix docs && ! mix docs 2>&1 | grep -q "warning:"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - run: mix deps.get
       - run: mix format --check-formatted
       - run: mix compile --force --warnings-as-errors
-      - run: mix credo -C no-test
+      - run: mix credo
       - run: mix test --include long_running
       - run: mix espec
       - run: MIX_ENV=dev mix docs && ! mix docs 2>&1 | grep -q "warning:"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.0
 jobs:
-  build:
+  test:
     docker:
       - image: membrane/membrane:latest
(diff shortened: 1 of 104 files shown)


.circleci/config.yml
```diff
@@ -11,8 +11,29 @@ jobs:
...
+      - run: mix compile
+      - run: mix credo
+      - run: mix docs && ! mix docs 2>&1 | grep -q "warning:"
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - test
+      - lint
```
bblaszkow06, line 39: The difference is ~ 1min:
https://app.circleci.com/pipelines/github/membraneframework/membrane_core/735/workflows/7f53c7d5-cffc-4384-b5d8-7d11ce255c81
https://app.circleci.com/pipelines/github/membraneframework/membrane_core/734/workflows/9f9c0dd1-75d0-4fd7-9092-44de6f873342

bblaszkow06:
The GitHub checks are stuck because of settings: task `build` is required, but only `test` and `lint` are now run. After approval these settings can be changed

lib/membrane/time.ex
```diff
@@ -250,12 +250,14 @@ defmodule Membrane.Time do
     Returns one #{unit.singular} in `#{inspect(__MODULE__)}` units.
     """
     @spec unquote(unit.singular)() :: t
+    # credo:disable-for-next-line Credo.Check.Readability.Specs
```
mat-hek, line 253: isn't the spec present?

test/membrane/element_test.exs
```diff
@@ -14,8 +14,10 @@ defmodule Membrane.ElementTest do
 
     def_options target: [type: :pid]
 
+    # credo:disable-for-next-line Credo.Check.Readability.Specs
     def assert_callback_called(name), do: assert_receive({:callback_called, ^name})
```
mat-hek, line 18: does this need to be public?

lib/membrane/time.ex
bblaszkow06: Credo does see that :man_shrugging:

test/membrane/element_test.exs
bblaszkow06: If placed in this module - yes because it's used in ElementTest

mat-hek: is there any reason why it's placed there?

bblaszkow06: Yep - cuz it's specific for this element (matches atom sent in callbacks) and you can call it like this: `TestFilter.assert_callback_called(:handle_end_of_stream)`

mat-hek: So maybe just add specs for that

PR #269, opened 2020-07-29
Feat/telemetry

Qizot:
Introduction to telemetry events generated by core. Starting with events emitted only by InputBuffer. Additionally created helper module for uniquely identifying element within pipeline via path of following element names, required for telemetry and maybe logging.


lib/membrane/telemetry/telemetry.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Membrane.Telemetry do
+  @moduledoc """
+  Defines basic telemetry event types used by Membrane.Core that can be consumed by attaching to [Telemetry Package](https://hex.pm/packages/telemetry).
```
bblaszkow06, line 3: If you really want to mention module then it should be wrapped with backticks (\`\`). However I think we can be more general here and say about the whole package:
```suggestion
  Defines basic telemetry event types used by Membrane's Core that can be consumed by attaching to [Telemetry Package](https://hex.pm/packages/telemetry).
```

lib/membrane/telemetry/telemetry.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  It is user's responsibility to use some sort of metrics reporter
+  that will be attached to mentioned library to consume and process generated metrics. Membrane.Core does not
```
bblaszkow06, line 6: Like above

lib/membrane/telemetry/telemetry.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Membrane.Telemetry do
+  @moduledoc """
+  Defines basic telemetry event types used by Membrane.Core that can be consumed by attaching to [Telemetry Package](https://hex.pm/packages/telemetry).
+
+  It is user's responsibility to use some sort of metrics reporter
```
bblaszkow06, line 5: If user has to handle metrics, he needs to know what they are and it should be visible in docs. You can even call the function from this module :wink: E.g. Phoenix has the list here: https://hexdocs.pm/phoenix/Phoenix.Logger.html

lib/membrane/helper/path_locator.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Membrane.Helper.PathLocator do
```
bblaszkow06, line 1: 1. `Helper` here is redundant. 
2. `PathLocator` would make sense if functions would really located anything (e.g. traversed up on process tree building the path). My suggestion is to name it `ProcessPath` (I'm missing a word describing pipelines, bins and elements. Any ideas @mat-hek ?)
3. Since module contains Path it's not needed in every single function as postfix. I think `ProcessPath.get()` etc. will be descriptive enough.

Qizot: Where should I put the `ProcessPath` then? Directly in `lib/`?

bblaszkow06: To be clear I meant `Membrane.ProcessPath` and yes, then it would land in `lib/membrane`

mat-hek: I thought about 'component'

PR #273, opened 2020-08-06
Bin & pipeline timer

### LINKED ISSUE #264: Add timers to bins
Timers can be started by elements only, I see no reason for not enabling them in bins

mat-hek:
closes #264


lib/membrane/core/callback_handler.ex
```diff
@@ -129,8 +129,7 @@ defmodule Membrane.Core.CallbackHandler do
   end
 
   @spec handle_callback_result(
-          {{:ok, [Membrane.Element.Action.t() | Membrane.Parent.Action.t()]},
-           internal_state :: any},
+          {{:ok, actions :: Keyword.t()}, internal_state :: any},
```
bblaszkow06, line 132: Isn't that spec too wide?

lib/membrane/core/child/pad_controller.ex
```diff
@@ -4,17 +4,17 @@ defmodule Membrane.Core.Child.PadController do
...
 
   use Bunch
-
   alias Bunch.Type
   alias Membrane.{Event, LinkError, Pad, ParentSpec}
   alias Membrane.Core
-  alias Membrane.Core.{CallbackHandler, Child, InputBuffer, Message}
+  alias Membrane.Core.{CallbackHandler, InputBuffer, Message}
   alias Membrane.Core.Bin.LinkingBuffer
+  alias Membrane.Core.{CallbackHandler, Component, Message, InputBuffer}
```
bblaszkow06, line 12: aliases here are duplicated

lib/membrane/core/element.ex
```diff
@@ -89,6 +89,14 @@ defmodule Membrane.Core.Element do
     :ok
   end
 
+  @doc """
+  Checks whether module is an element.
+  """
+  @spec bin?(module) :: boolean
+  def bin?(module) do
```
bblaszkow06, line 96:
```suggestion
  @spec element?(module) :: boolean
  def element?(module) do
```

bblaszkow06: BTW this is the duplicate of `Membrane.Element.element?/1`. Remove one.

spec/membrane/core/pipeline_spec.exs
```diff
@@ -0,0 +1,189 @@
+defmodule Membrane.Core.PipelineSpec do
```
bblaszkow06, line 1: Since you've already touched that - would it be possible to rewrite it to ExUnit?

lib/membrane/pipeline/action.ex
```diff
@@ -0,0 +1,100 @@
...
+  or stopping it via `t:stop_timer_t/0`.
+
+  If `interval` is set to `:no_interval`, the timer won't issue any ticks until
+  the interval is set with `t:timer_interval_t/0` action.
+
+  If no `clock` is passed, pipeline clock is chosen.
+
+  Timers use `Process.send_after/3` under the hood.
+  """
+  @type start_timer_t ::
```
bblaszkow06, line 55: The PR's name should include pipeline as well

lib/membrane/element/action.ex
```diff
@@ -168,7 +168,7 @@ defmodule Membrane.Element.Action do
   If `interval` is set to `:no_interval`, the timer won't issue any ticks until
   the interval is set with `t:timer_interval_t/0` action.
 
-  If no `clock` is passed, pipeline clock is chosen.
+  If no `clock` is passed, parent clock is chosen.
```
bblaszkow06, line 171:
```suggestion
  If no `clock` is passed, parent's clock is chosen.
```

lib/membrane/bin/action.ex
```diff
@@ -0,0 +1,99 @@
...
+  callback and can be used for changing its interval via `t:timer_interval_t/0`
+  or stopping it via `t:stop_timer_t/0`.
+
+  If `interval` is set to `:no_interval`, the timer won't issue any ticks until
+  the interval is set with `t:timer_interval_t/0` action.
+
+  If no `clock` is passed, parent clock is chosen.
+
+  Timers use `Process.send_after/3` under the hood.
+  """
```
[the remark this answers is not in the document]
mat-hek: There are several problems with that approach:
- user cannot just browse actions with their documentation
- actions get split into various modules: common for parents, children and all components
- the only thing that actions residing in `Parent.Action` would have in common is that they are common for Pipeline and Bin, what is not really valuable information for a user, or at least less valuable than having all the Bin actions, for example, in one place
- the docs are not exactly the same, they differ by references to the callbacks

If I was to change anything here, I would move all the actions to a single module, order them in a reasonable way (at least in code, as in docs they would be ordered alphabetically anyway) and reference in `Pipeline.action_t` etc

lib/membrane/core/callback_handler.ex
mat-hek: I'd like to keep callback handler independent from specific components. I even thought of releasing it as a separate library :P

lib/membrane/bin/action.ex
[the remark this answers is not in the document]
bblaszkow06: OK, so maybe we could just put the docs in one place programmatically? I mean doing something similar to get_pad_docs: https://github.com/membraneframework/membrane_core/blob/8060c4595edde3b90c18f396c487bdd40d4b8236/lib/membrane/core/child/pads_specs.ex#L35

PR #276, opened 2020-08-07
Feat/telemetry link event

Qizot:
Added new telemetry event indicating initialization of new link between elements.


lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -128,11 +128,36 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
     end
   end
 
+  defp get_pad_name(pad) do
```
mat-hek, line 131: it actually doesn't get pad name, but it converts private pad ref to the corresponding public one

mat-hek:
- please rebase/merge the master
- don't we need a corresponding event for link removal?

Issue #220, opened 2019-11-28
StartOfStream and EndOfStream events should be made private


mat-hek:
... as the API is now to return `:end_of_stream` action and start of stream is sent automatically with the first buffer

mat-hek:
closed as of #280

PR #280, opened 2020-08-24
fix docs modules groups

mat-hek:


mix.exs
```diff
@@ -84,7 +84,9 @@ defmodule Membrane.Mixfile do
           ~r/^Membrane.Sync$/,
           ~r/^Membrane.Time.*/,
           ~r/^Membrane.Pad.*/,
-          ~r/^Membrane.PlaybackState.*/
+          ~r/^Membrane.PlaybackState.*/,
+          ~r/^Membrane.Telemetry.*/,
+          ~r/^Membrane.ComponentPath.*/
```
bblaszkow06, line 89: After checking out the docs, there are a couple more things to fix:
- `PipelineError` & `ParentError` should rather be in `Exceptions` group (the dot in all regexes above mean any character as it's not escaped
- `Membrane.Pad` under utils is a bit "meh". Maybe `Communication` would be a better group?
- Deprecated group should most likely be at the bottom, after exceptions

bblaszkow06 (changes requested):

mat-hek:
closes #220

bblaszkow06 (approved):

PR #283, opened 2020-10-07
improve generating options docs, generating bins moduledocs

mat-hek:
should be merged after #282

### DIFF
--- a/lib/membrane/core/child.ex
+++ b/lib/membrane/core/child.ex
@@ -10,7 +10,7 @@ defmodule Membrane.Core.Child do
   ]
 
   @spec generate_moduledoc(module, :element | :bin) :: Macro.t()
-  def generate_moduledoc(module, child) do
+  def generate_moduledoc(module, child_type) do
     membrane_pads_moduledoc =
       Module.get_attribute(module, :membrane_pads)
       |> __MODULE__.PadsSpecs.generate_docs_from_pads_specs()
@@ -26,9 +26,7 @@ defmodule Membrane.Core.Child do
     if moduledoc != false do
       moduledoc =
         if String.trim(moduledoc) == "" do
-          quote do
-            unquote(moduledoc) <> "Membrane #{unquote(child)}.\n\n"
-          end
+          "Membrane #{child_type}.\n\n"
         else
           moduledoc
         end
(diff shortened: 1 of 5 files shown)


lib/membrane/core/child.ex
```diff
@@ -2,4 +2,50 @@ defmodule Membrane.Core.Child do
...
   @type state_t :: Membrane.Core.Element.State.t() | Membrane.Core.Bin.State.t()
+
+  @docs_order [
+    :membrane_options_moduledoc,
+    :membrane_pads_moduledoc,
+    :membrane_clock_moduledoc
+  ]
+
+  @spec generate_moduledoc(module, :element | :bin) :: Macro.t()
+  def generate_moduledoc(module, child) do
```
bblaszkow06, line 13: Maybe `child_type` would be more descriptive? Or something similar?

lib/membrane/core/child.ex
```diff
@@ -2,4 +2,50 @@ defmodule Membrane.Core.Child do
...
+      case Module.get_attribute(module, :moduledoc, "") do
+        {_line, doc} -> doc
+        doc -> doc
+      end
+
+    if moduledoc != false do
+      moduledoc =
+        if String.trim(moduledoc) == "" do
+          quote do
+            unquote(moduledoc) <> "Membrane #{unquote(child)}.\n\n"
```
bblaszkow06, line 30: What is the purpose of concatenating whitespaces (`String.trim(moduledoc) == ""`) to that line?

lib/membrane/core/child/pads_specs.ex
```diff
@@ -222,7 +222,7 @@ defmodule Membrane.Core.Child.PadsSpecs do
       config
       |> Enum.map(fn {k, v} ->
         {
-          k |> to_string() |> String.replace("_", "&nbsp;") |> String.capitalize(),
+          k |> to_string() |> String.replace("_", " ") |> String.capitalize(),
```
bblaszkow06, line 225: I'm pretty sure it was there for a reason, regular space may allow line break. Are you sure this change doesn't break anything?

mat-hek: `&nbsp;` just doesn't work there anymore, it's just displayed as is, like any html

bblaszkow06 (approved):

PR #284, opened 2020-10-14
fix bugs and refactor children playback handling in parent

### LINKED ISSUE #104: RC on playback state change when dynamically spawning children
Consider the following scenario:
1. pipeline changes playback state to playing
2. new child `X` is spawned
3. pipeline starts switching to prepared
4. all children receive request to go to prepared, besides `X`, who's changing to playing and does not receive message yet
5. `X` goes from stopped to prepared and notifies the pipeline, all other children go from playing to prepared and also notify the pipeline
6. pipeline finishes the change to prepared
7. `X` switches to playing, because its target state has not been updated yet
8. `X` finally receives  the request that others received at 4.
9. `X` goes to prepared

In points 7. and 8. `X` is in playing, and the pipeline is in prepared, which is unwanted. Although this does not seem to be a serious bug, it may cause problems. Informimg pipeline of both new and target playback state might be a fix.

### LINKED ISSUE #214: `handle_linking_finished` always called on all elements
When only partial spec action is returned from callback (only concerning a subset of children), `handle_linking_finished` is called on every child element. This is due to taking children from parent's state and not from resolved links.

mat-hek:
Fixes #104 and another bug: a child spawned during parent playback change remained in the playback state the parent was at the moment of spawning, instead of continuing the change with the parent

Fixes #214

### DIFF
--- a/lib/membrane/child_entry.ex
+++ b/lib/membrane/child_entry.ex
@@ -11,6 +11,8 @@ defmodule Membrane.ChildEntry do
   Other fields in the struct ARE NOT PART OF THE PUBLIC API and should not be
   accessed or relied on.
   """
+  use Bunch.Access
+
   @type t :: %__MODULE__{
           name: Membrane.Child.name_t(),
           module: module,
@@ -19,8 +21,17 @@ defmodule Membrane.ChildEntry do
           pid: pid,
           clock: Membrane.Clock.t(),
           sync: Membrane.Sync.t(),
-          pending?: boolean()
+          playback_synced?: boolean()
         }
 
-  defstruct [:name, :module, :options, :component_type, :pid, :clock, :sync, :pending?]
+  defstruct [
+    :name,
+    :module,
+    :options,
+    :component_type,
+    :pid,
+    :clock,
+    :sync,
+    playback_synced?: false
+  ]
 end
--- a/lib/membrane/child_entry.ex
+++ b/lib/membrane/child_entry.ex
@@ -21,7 +21,8 @@ defmodule Membrane.ChildEntry do
           pid: pid,
           clock: Membrane.Clock.t(),
           sync: Membrane.Sync.t(),
-          playback_synced?: boolean()
+          playback_synced?: boolean(),
+          terminating?: boolean()
         }
 
   defstruct [
@@ -32,6 +33,7 @@ defmodule Membrane.ChildEntry do
     :pid,
     :clock,
     :sync,
-    playback_synced?: false
+    playback_synced?: false,
+    terminating?: false
   ]
 end
--- a/lib/membrane/core/parent/child_life_controller/link_handler.ex
+++ b/lib/membrane/core/parent/child_life_controller/link_handler.ex
@@ -194,8 +194,8 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
     endpoints =
       links
       |> Enum.flat_map(&[&1.from, &1.to])
-      |> Enum.uniq()
       |> Enum.reject(&(&1.child == {Membrane.Bin, :itself}))
+      |> Enum.uniq()
 
     Bunch.Enum.try_each(endpoints, &Message.call(&1.pid, :linking_finished))
   end
--- a/lib/membrane/core/parent/child_life_controller/link_handler.ex
+++ b/lib/membrane/core/parent/child_life_controller/link_handler.ex
@@ -191,12 +191,10 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
   end
 
   defp send_linking_finished(links) do
-    endpoints =
-      links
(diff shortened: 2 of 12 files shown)


lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -195,4 +189,14 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
     ])
     ~> {&1, state}
   end
+
+  defp send_linking_finished(links) do
+    endpoints =
+      links
+      |> Enum.flat_map(&[&1.from, &1.to])
+      |> Enum.uniq()
+      |> Enum.reject(&(&1.child == {Membrane.Bin, :itself}))
```
bblaszkow06, line 198: Probably negligible performance impact, but `uniq` is probably `O(n*log(n))`:
```suggestion
      links
      |> Enum.flat_map(&[&1.from, &1.to])
      |> Enum.reject(&(&1.child == {Membrane.Bin, :itself}))
      |> Enum.uniq()
```

lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -195,4 +189,14 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
...
+
+    Bunch.Enum.try_each(endpoints, &Message.call(&1.pid, :linking_finished))
```
bblaszkow06, line 200: Why it isn't part of the pipe above?

lib/membrane/core/parent/children_model.ex
```diff
@@ -35,28 +35,34 @@ defmodule Membrane.Core.Parent.ChildrenModel do
...
-  end
+  @spec update_children(Parent.state_t(), [Child.name_t()], (ChildEntry.t() -> ChildEntry.t())) ::
+          {:ok, Parent.state_t()} | {:error, {:unknown_child, Child.name_t()}}
+  def update_children(state, children_names, mapper) do
+    result =
+      Bunch.Enum.try_reduce(children_names, state.children, fn name, children ->
+        case Map.fetch(children, name) do
+          {:ok, child} -> {:ok, %{children | name => mapper.(child)}}
+          :error -> {:error, {:unknown_child, name}}
+        end
```
bblaszkow06, line 46:
```suggestion
        if Map.has_key?(children, name) do
          {:ok, Map.update!(children, name, mapper)}
        else
          {:error, {:unknown_child, name}}
        end
```

lib/membrane/core/playback_handler.ex
```diff
@@ -165,10 +165,15 @@ defmodule Membrane.Core.PlaybackHandler do
...
+           controlling_pid: controlling_pid,
+           playback: %{state: playback_state, target_state: playback_state}
+         }
+       )
        when not is_nil(controlling_pid),
-       do:
-         handler.notify_controller(:playback_changed, state.playback.state, state.controlling_pid)
+       do: handler.notify_controller(:playback_changed, playback_state, controlling_pid)
 
   defp maybe_notify_controller(_handler, _state), do: :ok
```
bblaszkow06, line 179: I don't see the tests confirming the bugs described in PR are fixed :wink:

lib/membrane/child_entry.ex
```diff
@@ -11,6 +11,8 @@ defmodule Membrane.ChildEntry do
   Other fields in the struct ARE NOT PART OF THE PUBLIC API and should not be
   accessed or relied on.
   """
+  use Bunch.Access
+
```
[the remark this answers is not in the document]
mat-hek: It is, for use with `put_in` and friends.

lib/membrane/core/parent/child_life_controller/link_handler.ex
mat-hek: It's more readable this way IMO

lib/membrane/core/playback_handler.ex
mat-hek: Added a test, it doesn't check for the bugs that are race conditions and would take too much work to test

lib/membrane/core/parent/child_life_controller/link_handler.ex
bblaszkow06: I really don't see any benefit besides naming endpoints, but you can to this in pipe
```suggestion
      links
      |> Enum.flat_map(&[&1.from, &1.to])
      |> Enum.uniq()
      |> Enum.reject(&(&1.child == {Membrane.Bin, :itself}))
      |> Bunch.Enum.try_each(fn endpoint -> Message.call(endpoint.pid, :linking_finished) end)
```
(commiting above from GH won't work)
Anyway, it's a matter of personal preference, so do as you wish

bblaszkow06 (approved):
Just one nitpick left. Merry Christmas!

bblaszkow06 (approved):

PR #287, opened 2021-02-18
Develop

mat-hek:


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -119,11 +119,12 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+  def maybe_handle_child_death(pid, reason, state) do
     withl find: {:ok, child_name} <- child_by_pid(pid, state),
+          assert: :normal = reason,
           handle: state = Bunch.Access.delete_in(state, [:children, child_name]),
           handle: {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
-          {{:ok, :child}, state}
+      {{:ok, :child}, state}
     else
       find: :error -> {{:ok, :not_child}, state}
       handle: error -> error
```
AndrzejPodo: `:assert` error is not handled in `else`

AndrzejPodo (changes requested):

mat-hek:
@AndrzejPodo I refactored the linking part, hope it's more intuitive now. Also added docs and specs. Only espec tests don't pass, though they're fixed in one of your PRs afaik.

lib/membrane/core/child/pad_controller.ex
```diff
@@ -27,30 +27,38 @@ defmodule Membrane.Core.Child.PadController do
...
+    name = this.pad_ref |> Pad.name_by_ref()
+    info = state.pads.info[name]
+
+    {:ok, other_info} =
+      if other_info do
+        {:ok, other_info}
+      else
+        other_direction = Pad.opposite_direction(direction)
+        Message.call(other.pid, :handle_link, [other_direction, other, this, info])
+      end
```
AndrzejPodo, line 46: I'm wondering if the `direction` argument is necessary. It seems to be used only for validation purpose. Is there not enough information in `info`?

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -117,18 +117,24 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+  def maybe_handle_child_death(pid, reason, state) do
+    withl find: {:ok, child_name} <- child_by_pid(pid, state),
+          assert: :normal = reason,
+          handle: state = Bunch.Access.delete_in(state, [:children, child_name]),
+          handle: {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
+      {{:ok, :child}, state}
+    else
+      find: :error -> {{:ok, :not_child}, state}
+      handle: error -> error
+    end
```
AndrzejPodo, line 131: It would be nice if there was a short docs here. Especially `LifecycleController.maybe_finish_playback_transition(state)` was not very clear to me at the first glance.

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -117,18 +117,24 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
   end
```
AndrzejPodo, line 132: It would be nice if there was a short docs here. Especially `LifecycleController.maybe_finish_playback_transition(state)` was not very clear to me at the first glance.

lib/membrane/core/child/pad_controller.ex
```diff
@@ -27,30 +27,38 @@ defmodule Membrane.Core.Child.PadController do
...
 
-    with :ok <- validate_pad_being_linked!(pad_ref, direction, info, state),
-         :ok <- validate_dir_and_mode!({pad_ref, info}, {other_ref, other_info}) do
-      props = parse_pad_props!(props, pad_name, state)
-      state = init_pad_data(info, pad_ref, pid, other_ref, props, state)
+    with :ok <- validate_pad_being_linked!(this.pad_ref, direction, info, state),
+         :ok <- validate_dir_and_mode!({this.pad_ref, info}, {other.pad_ref, other_info}) do
+      props = parse_pad_props!(this.pad_props, name, state)
+
+      state =
```
AndrzejPodo, line 52: I'm wondering if the `direction` argument is necessary. It seems to be used only for validation purpose. Is there not enough information in info?

lib/membrane/core/parent/link.ex
```diff
@@ -53,10 +53,8 @@ defmodule Membrane.Core.Parent.Link do
       %ParentSpec.LinkBuilder{links: [%{from: from} | _]} ->
         raise ParentError,
               "Invalid link specification: link from #{inspect(from)} lacks its destination."
-
-      links ->
-        from_spec(links)
     end)
+    |> Enum.reverse()
```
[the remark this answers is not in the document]
AndrzejPodo: ok, good note.

lib/membrane/core/child/pad_controller.ex
mat-hek: The direction argument is the direction determined from links specification, while the direction in `info` is the one specified in the element's pads specification. We need to ensure they match so that nobody links e.g. input as an output. You can see that a few lines below:
```
      info.direction != direction ->
        raise LinkError, """
        Invalid pad direction:
          expected: #{inspect(info.direction)},
          actual: #{inspect(direction)}
        """
```

AndrzejPodo (dismissed):

andpodob (approved):

PR #288, opened 2021-03-12
Handle delay supply fix

andpodob:
- Removes necessity of calling `DemandHandler.handle_delayed_demands()` after receiving a message.
- Replaces `delayed_demands Map` with `delayed_demands Set` after getting rid of sync/async delayed demand types

### DIFF
--- a/lib/membrane/core/element/demand_handler.ex
+++ b/lib/membrane/core/element/demand_handler.ex
@@ -129,7 +129,18 @@ defmodule Membrane.Core.Element.DemandHandler do
           Pad.ref_t(),
           State.t()
         ) :: State.stateful_try_t()
+  def supply_demand(pad_ref, %State{supplying_demand?: true} = state) do
+    {:ok, delay_supply(pad_ref, :async, state)}
+  end
+
   def supply_demand(pad_ref, state) do
+    with {:ok, state} <- do_supply_demand(pad_ref, %State{state | supplying_demand?: true}) do
+      {:ok, state} = handle_delayed_demands(state)
+      {:ok, %State{state | supplying_demand?: false}}
+    end
+  end
+
+  defp do_supply_demand(pad_ref, state) do
     pad_data = state |> PadModel.get_data!(pad_ref)
 
     {{_buffer_status, data}, new_input_buf} =
@@ -174,10 +185,10 @@ defmodule Membrane.Core.Element.DemandHandler do
           State.t()
         ) :: State.stateful_try_t()
   defp do_handle_input_buf_output(pad_ref, {:event, e}, state),
-    do: EventController.exec_handle_event(pad_ref, e, %{supplying_demand?: true}, state)
+    do: EventController.exec_handle_event(pad_ref, e, state)
 
   defp do_handle_input_buf_output(pad_ref, {:caps, c}, state),
-    do: CapsController.exec_handle_caps(pad_ref, c, %{supplying_demand?: true}, state)
+    do: CapsController.exec_handle_caps(pad_ref, c, state)
 
   defp do_handle_input_buf_output(
          pad_ref,
@@ -186,6 +197,6 @@ defmodule Membrane.Core.Element.DemandHandler do
        ) do
     state = PadModel.update_data!(state, pad_ref, :demand, &(&1 - size))
 
-    BufferController.exec_buffer_handler(pad_ref, buffers, %{supplying_demand?: true}, state)
+    BufferController.exec_buffer_handler(pad_ref, buffers, state)
   end
 end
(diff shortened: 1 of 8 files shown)


lib/membrane/core/element/demand_handler.ex
```diff
@@ -67,15 +69,10 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-    |> Bunch.Struct.put_in([:delayed_demands, {pad_ref, :supply}], :async)
-  end
-
-  def delay_supply(pad_ref, :sync, state) do
+  @spec delay_supply(Pad.ref_t(), State.t()) :: State.t()
+  def delay_supply(pad_ref, state) do
     state
-    |> Map.update!(:delayed_demands, &Map.put_new(&1, {pad_ref, :supply}, :sync))
+    |> Map.update!(:delayed_demands, &MapSet.put(&1, {pad_ref, :supply}))
   end
```
mat-hek, line 76: this is now only called by supply_demand, let's move the code there

lib/membrane/core/element/demand_handler.ex
```diff
@@ -87,11 +84,11 @@ defmodule Membrane.Core.Element.DemandHandler do
   @spec delay_redemand(Pad.ref_t(), State.t()) :: State.t()
   def delay_redemand(pad_ref, state) do
```
mat-hek, line 85: this is now only called by handle_redemand, let's move the code there

lib/membrane/core/element/demand_handler.ex
```diff
@@ -100,25 +97,33 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-          supply_demand(pad_ref, state)
+    case action do
+      :supply ->
+        supply_demand(pad_ref, state)
 
-        {:supply, :async} ->
-          Message.self(:invoke_supply_demand, pad_ref)
-          {:ok, state}
+      :redemand ->
+        DemandController.handle_demand(pad_ref, 0, state)
```
mat-hek, line 108: calling `handle_redemand` would look better

lib/membrane/core/element/demand_handler.ex
```diff
@@ -129,7 +134,18 @@ defmodule Membrane.Core.Element.DemandHandler do
...
           State.t()
         ) :: State.stateful_try_t()
+  def supply_demand(pad_ref, %State{supplying_demand?: true} = state) do
+    {:ok, delay_supply(pad_ref, state)}
+  end
+
   def supply_demand(pad_ref, state) do
+    with {:ok, state} <- do_supply_demand(pad_ref, %State{state | supplying_demand?: true}) do
+      {:ok, state} = handle_delayed_demands(%State{state | supplying_demand?: false})
+      {:ok, %State{state | supplying_demand?: false}}
```
mat-hek, line 144: seems redundant

mat-hek (changes requested):

lib/membrane/core/element/demand_handler.ex
AndrzejPodo: Which part do you mean?

lib/membrane/core/element/demand_handler.ex
```diff
@@ -100,53 +79,64 @@ defmodule Membrane.Core.Element.DemandHandler do
...
+    case action do
+      :supply ->
+        supply_demand(pad_ref, state)
 
-        :redemand ->
-          DemandController.handle_demand(pad_ref, 0, state)
-      end
-
-    with {:ok, state} <- res do
-      handle_delayed_demands(state)
```
mat-hek, line 120: why remove that? seems still needed

mat-hek (changes requested):

lib/membrane/core/element/demand_handler.ex
mat-hek: image

AndrzejPodo: hm, but `handle_delayed_demands` calls `supply_demand` which calls `do_supply_demand` with `supplying_demand?: true`. I'm quite unsure if `supplying_demand?` is always `false` after all this calls.

lib/membrane/core/element/demand_handler.ex
AndrzejPodo: if i remember well this recurrence call has been moved to `supply_demand`

lib/membrane/core/element/demand_handler.ex
mat-hek: If it was to be set to true after `handle_delayed_demands`, something would have to change it there. However, the only place where it changes is in `supply_demand` and it's changed back immediately, so I see no possibility that it remains true.

AndrzejPodo: Ok, I see, you're right.

lib/membrane/core/element/demand_handler.ex
mat-hek: Ok, I see... though it won't be sufficient if supply_demand results in more than one delayed demands

AndrzejPodo: I see, good point.

mat-hek (approved):

PR #295, opened 2021-04-14
add handle_shutdown callback for bins

Qizot:


lib/membrane/bin.ex
```diff
@@ -47,6 +47,15 @@ defmodule Membrane.Bin do
...
   @callback handle_init(options :: options_t) :: callback_return_t()
 
+  @doc """
+  Callback invoked when bin is shutting down.
+  Internally called in `c:GenServer.terminate/2` callback.
+
+  Useful for any cleanup required.
+  """
+  @callback handle_shutdown(reason, state :: state_t) :: :ok
+            when reason: :normal | :shutdown | {:shutdown, any}
```
mat-hek, line 57: https://hexdocs.pm/elixir/GenServer.html#c:terminate/2 says `when reason: :normal | :shutdown | {:shutdown, term()} | term()`. Please fix in pipeline and element too.

mat-hek (approved):

lib/membrane/bin.ex
Qizot: Done

PR #299, opened 2021-04-26
Links

### LINKED ISSUE #159: A new link spec
Passing links as a map doesn't allow to link an element with dynamic output pad more than once, because keys in a map cannot be duplicated.

The idea is to create API similar to ESpec's assertions:
```elixir
links = [
 link(:source) |> to(:filter),
 link(:filter, :out_pad) |> to(:filter_b, :in_pad),
 link(:filter_b, :out_pad, pad: [opt: true]) |> to(:filter_b, :in_pad, buffer: [preffered_size: 42]]),
]
```

mat-hek:
closes #159 
Usage example: https://github.com/membraneframework/membrane_webrtc_plugin/pull/16

### DIFF
--- a/lib/membrane/core/parent/child_entry_parser.ex
+++ b/lib/membrane/core/parent/child_entry_parser.ex
@@ -10,8 +10,10 @@ defmodule Membrane.Core.Parent.ChildEntryParser do
           component_type: :element | :bin
         }
 
-  @spec from_spec(ParentSpec.children_spec_t() | any) :: [raw_child_entry_t] | no_return
-  def from_spec(children_spec) when is_map(children_spec) or is_list(children_spec) do
+  @spec parse_children_entries(ParentSpec.children_spec_t() | any) ::
+          [raw_child_entry_t] | no_return
+  def parse_children_entries(children_spec)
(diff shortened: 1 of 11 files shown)


lib/membrane/core/parent/child_entry_parser.ex
```diff
@@ -10,8 +10,10 @@ defmodule Membrane.Core.Parent.ChildEntryParser do
           component_type: :element | :bin
         }
 
-  @spec from_spec(ParentSpec.children_spec_t() | any) :: [raw_child_entry_t] | no_return
-  def from_spec(children_spec) when is_map(children_spec) or is_list(children_spec) do
+  @spec parse_children_entries(ParentSpec.children_spec_t() | any) ::
+          [raw_child_entry_t] | no_return
+  def parse_children_entries(children_spec)
```
mickel8, line 15: I would consider renaming it to `parse` as we have child entry in the Module's name but don't insist

mickel8 (approved):

andpodob (approved):

mickel8 (approved):

PR #300, opened 2021-04-27
pass messages not from child to handle_other

andpodob:
Fix the issue of handling `:DOWN` messages from processes that are not children of pipeline. This PR makes `develop` to work properly when crash groups are not used. Crash groups are yet to be debugged.

### DIFF
--- a/lib/membrane/core/parent/child_life_controller.ex
+++ b/lib/membrane/core/parent/child_life_controller.ex
@@ -144,23 +144,23 @@ defmodule Membrane.Core.Parent.ChildLifeController do
   - if in playback transition, checks if it can be finished (in case the dead child
     was the last one we were waiting for to change the playback state)
 
-  If a pid turns out not to be a pid of any child, it is a NOOP.
+  If a pid turns out not to be a pid of any child error is raised.
   """
   @spec handle_child_death(child_pid :: pid(), reason :: atom(), state :: Parent.state_t()) ::
           {:ok | {:error, :not_child}, Parent.state_t()}
   def handle_child_death(pid, :normal, state) do
     with {:ok, child_name} <- child_by_pid(pid, state),
          state = Bunch.Access.delete_in(state, [:children, child_name]),
-         state = remove_child_links(child_name, state),
-         {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
-      {:ok, state}
+         state = remove_child_links(child_name, state) do
+      LifecycleController.maybe_finish_playback_transition(state)
     else
-      {:error, :not_child} -> {{:error, :not_child}, state}
-      error -> error
+      {:error, :not_child} ->
+        raise Membrane.PipelineError,
+              "Tried to handle child death that wasn't a child of that pipeline."
     end
   end
 
-  def handle_child_death(pid, {:shutdown, :group_kill}, state) do
+  def handle_child_death(pid, {:shutdown, :membrane_crash_group_kill}, state) do
     with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state) do
       state =
         state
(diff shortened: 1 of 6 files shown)


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -146,44 +146,43 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
-          handle: {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
-      {{:ok, :child}, state}
+  @spec handle_child_death(child_pid :: pid(), reason :: atom(), state :: Parent.state_t()) ::
+          {:ok | {:error, :not_child}, Parent.state_t()}
+  def handle_child_death(pid, :normal, state) do
+    with {:ok, child_name} <- child_by_pid(pid, state),
+         state = Bunch.Access.delete_in(state, [:children, child_name]),
+         state = remove_child_links(child_name, state),
+         {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
+      {:ok, state}
```
mat-hek, line 156:
```suggestion
    with {:ok, child_name} <- child_by_pid(pid, state) do
       state = Bunch.Access.delete_in(state, [:children, child_name])
       state = remove_child_links(child_name, state)
       LifecycleController.maybe_finish_playback_transition(state)
```

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
andpodob: `maybe_finish_playback_transition(state)` is having specs for return value like this: `{:ok | {:error, any}, Parent.state_t()}`, is that safe to just return result of `LifecycleController.maybe_finish_playback_transition(state)`?

mat-hek: Specs (and possibly error handling) for `maybe_finish_playback_transition` should be fixed, but the code I suggested works exactly the same way as yours, unless I'm missing something

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -144,23 +144,23 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
-         {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
-      {:ok, state}
+         state = remove_child_links(child_name, state) do
+      LifecycleController.maybe_finish_playback_transition(state)
     else
-      {:error, :not_child} -> {{:error, :not_child}, state}
-      error -> error
+      {:error, :not_child} ->
+        raise Membrane.PipelineError,
+              "Tried to handle child death that wasn't a child of that pipeline."
```
mat-hek, line 159: fix english :P

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -144,23 +144,23 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
   """
   @spec handle_child_death(child_pid :: pid(), reason :: atom(), state :: Parent.state_t()) ::
           {:ok | {:error, :not_child}, Parent.state_t()}
   def handle_child_death(pid, :normal, state) do
     with {:ok, child_name} <- child_by_pid(pid, state),
          state = Bunch.Access.delete_in(state, [:children, child_name]),
-         state = remove_child_links(child_name, state),
-         {:ok, state} <- LifecycleController.maybe_finish_playback_transition(state) do
-      {:ok, state}
+         state = remove_child_links(child_name, state) do
```
mat-hek: assignments can be moved out of with

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -169,7 +169,9 @@ defmodule Membrane.Core.Parent.ChildLifeController do
 
       {:ok, state}
     else
-      {:error, :not_member} = error -> {error, state}
+      {:error, :not_member} ->
+        raise Membrane.PipelineError,
+              "Child  that was not a member of a group killed with :membrane_crash_group_kill."
```
mat-hek, line 174:
```suggestion
              "Child that was not a member of a crash group killed with :membrane_crash_group_kill."
```

mat-hek (changes requested):

mat-hek (approved):

PR #302, opened 2021-04-30
Pipeline handle group down

andpodob:
* add callback `handle_crash_group_down/3`
* add wrapper in `Membrane.Testing.Pipeline` for that callback
* updated integration tests to verify if callback is being called
* changed syntax from `x |> f(y)` to `f(x, y)` where it was looking bad imo

### DIFF
--- a/lib/membrane/core/parent/child_life_controller.ex
+++ b/lib/membrane/core/parent/child_life_controller.ex
@@ -180,18 +180,17 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         raise Membrane.PipelineError,
               "Tried to handle death of process that wasn't a child of that pipeline."
 
-      {:error, :not_member} ->
-        if reason == {:shutdown, :membrane_crash_group_kill} do
-          raise Membrane.PipelineError,
-                "Child that was not a member of any crash group killed with :membrane_crash_group_kill."
-        else
-          Membrane.Logger.debug("""
-          Pipeline child crashed but was not a member of any crash group.
-          Terminating.
-          """)
+      {:error, :not_member} when reason == {:shutdown, :membrane_crash_group_kill} ->
+        raise Membrane.PipelineError,
+              "Child that was not a member of any crash group killed with :membrane_crash_group_kill."
 
-          propagate_child_crash(state)
-        end
+      {:error, :not_member} ->
+        Membrane.Logger.debug("""
+        Pipeline child crashed but was not a member of any crash group.
+        Terminating.
+        """)
+
+        propagate_child_crash(state)
     end
   end
 
--- a/lib/membrane/core/parent/child_life_controller.ex
+++ b/lib/membrane/core/parent/child_life_controller.ex
@@ -4,7 +4,7 @@ defmodule Membrane.Core.Parent.ChildLifeController do
 
   alias __MODULE__.{StartupHandler, LinkHandler, CrashGroupHandler}
   alias Membrane.ParentSpec
-  alias Membrane.Core.{Parent, PlaybackHandler, CallbackHandler}
+  alias Membrane.Core.{Parent, PlaybackHandler, CallbackHandler, Component}
 
   alias Membrane.Core.Parent.{
     ChildEntryParser,
@@ -18,6 +18,7 @@ defmodule Membrane.Core.Parent.ChildLifeController do
   require Membrane.Logger
   require Membrane.Bin
   require Membrane.Element
+  require Membrane.Core.Component
 
   @spec handle_spec(ParentSpec.t(), Parent.state_t()) ::
           {{:ok, [Membrane.Child.name_t()]}, Parent.state_t()} | no_return
@@ -181,13 +182,18 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         |> CrashGroupHandler.remove_member_of_crash_group(group.name, pid)
         |> CrashGroupHandler.remove_crash_group_if_empty(group.name)
 
-      CallbackHandler.exec_and_handle_callback(
-          :handle_crash_group_down,
-          Membrane.Core.Pipeline.ActionHandler,
-          %{context: fn _state -> group.members end},
-          [group.name],
-          state
+      context =
+        Component.callback_context_generator(:parent, CrashGroupDown, state,
+          members: group.members
         )
+
+      CallbackHandler.exec_and_handle_callback(
+        :handle_crash_group_down,
+        Membrane.Core.Pipeline.ActionHandler,
+        %{context: context},
+        [group.name],
+        state
+      )
 
       {:ok, state}
     else
--- a/lib/membrane/core/parent/child_life_controller.ex
+++ b/lib/membrane/core/parent/child_life_controller.ex
@@ -41,18 +41,19 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         state.children_log_metadata
       )
 
+    children_names = children |> Enum.map(& &1.name)
+    children_pids = children |> Enum.map(& &1.pid)
+
     # monitoring children
-    :ok = children |> Enum.each(&Process.monitor(&1.pid))
+    :ok = Enum.each(children_pids, &Process.monitor(&1))
 
     :ok = StartupHandler.maybe_activate_syncs(syncs, state)
     {:ok, state} = StartupHandler.add_children(children, state)
-    children_names = children |> Enum.map(& &1.name)
 
     # adding crash group to state
     {:ok, state} =
       if spec.crash_group do
-        children_pids = children |> Enum.map(& &1.pid)
-        CrashGroupHandler.add_crash_group(spec.crash_group, children_pids, state)
+        CrashGroupHandler.add_crash_group(spec.crash_group, children_names, children_pids, state)
       else
         {:ok, state}
       end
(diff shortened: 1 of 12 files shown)


mat-hek (changes requested):
btw let's move CrashGroupHandler to membrane/core/parent

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -207,6 +198,25 @@ defmodule Membrane.Core.Parent.ChildLifeController do
     end
   end
 
+  defp exec_handle_crash_group_down_callback(group, state) do
+    members_names =
+      Enum.map(group.members, fn member_pid ->
```
mat-hek, line 203: won't this be empty when we execute the callback?

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
andpodob: oops, I think you're right

lib/membrane/core/parent/crash_group.ex
```diff
@@ -13,9 +13,10 @@ defmodule Membrane.Core.Parent.CrashGroup do
   @type t :: %__MODULE__{
           name: name_t(),
           mode: :temporary,
-          members: [pid()]
+          members: [pid()],
+          dead_members: [pid()]
```
mat-hek, line 17: couldn't we just keep a list of element names here?

andpodob: Ofc we can but we use this list to i.a. send messages to members. Keeping pids let us avoid necessity to change name to pid every time we send message.

mat-hek: But using names makes it consistent with all the core and allows to do anything else without converting pid to name. However, crash groups are quite specific, because apart from sending messages, they get the monitor down signals, which are also identified by pids. That's why I'd stick to
- `members`, containing names of all the children in the group, no matter if alive or dead
- `alive_pids`, containing pids of members that are alive

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,33 +163,49 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
 
       {:ok, state}
     else
+      {:error, :not_child} ->
+        raise Membrane.PipelineError,
+              "Tried to handle death of process that wasn't a child of that pipeline."
+
       {:error, :not_member} ->
         Membrane.Logger.debug("""
         Pipeline child crashed but was not a member of any crash group.
```
mat-hek: These clauses differ only by error handling, let's merge them

lib/membrane/core/parent/crash_group.ex
```diff
@@ -13,10 +13,11 @@ defmodule Membrane.Core.Parent.CrashGroup do
...
-          members: [pid()],
-          dead_members: [pid()]
+          members: [Membrane.Child.name_t()],
+          alive_members: [pid()],
+          triggered?: boolean()
         }
 
   @enforce_keys [:name, :mode]
-  defstruct @enforce_keys ++ [members: []] ++ [dead_members: []]
+  defstruct @enforce_keys ++ [members: [], alive_members: [], triggered?: false]
```
mat-hek, line 22:
```suggestion
  defstruct @enforce_keys ++ [members: [], alive_members_pids: [], triggered?: false]
```

lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -40,7 +40,12 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
 
   @spec unlink_crash_group(CrashGroup.t(), Parent.state_t()) :: Parent.state_t()
   def unlink_crash_group(crash_group, state) do
-    %CrashGroup{members: members_pids} = crash_group
+    %CrashGroup{members: members_names} = crash_group
+
+    members_pids =
+      Enum.map(state.children, &elem(&1, 1))
+      |> Enum.filter(&(&1.name in members_names))
+      |> Enum.map(& &1.pid)
```
mat-hek, line 48: we can just operate on names
```suggestion
```

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,33 +163,49 @@ defmodule Membrane.Core.Parent.ChildLifeController do
   end
 
   def handle_child_death(pid, {:shutdown, :membrane_crash_group_kill}, state) do
-    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state) do
+    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state),
+         {:ok, child_name} <- child_by_pid(pid, state) do
       {result, state} =
         state
+        |> Bunch.Access.delete_in([:children, child_name])
```
mat-hek, line 170: we should do that when all the children from the crash group are dead

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,33 +163,49 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
     end
   end
 
   def handle_child_death(pid, _reason, state) do
-    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state) do
+    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state),
+         {:ok, child_name} <- child_by_pid(pid, state) do
       {result, state} =
         crash_all_group_members(group, state)
+        |> Bunch.Access.delete_in([:children, child_name])
```
mat-hek, line 195: as above

lib/membrane/core/parent/child_life_controller/crash_group_handler.ex
```diff
@@ -7,22 +7,45 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupHandler do
...
+      | crash_groups:
+          Map.update(
+            state.crash_groups,
+            group_name,
+            crash_group,
+            fn %CrashGroup{members: current_children_names, alive_members: current_alive_members} = existing_group ->
+              %{existing_group | members: current_children_names ++ children_names, alive_members: current_alive_members ++ children_pids}
+            end
+          )
+    }
```
mat-hek, line 38: use `update_in`

lib/membrane/core/parent/child_life_controller/crash_group_handler.ex
```diff
@@ -51,13 +73,26 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupHandler do
...
 
     case crash_group do
       %CrashGroup{} -> {:ok, crash_group}
       nil -> {:error, :not_member}
     end
   end
+
+  @spec set_triggered(Pipeline.State.t(), CrashGroup.name_t(), boolean()) :: Pipeline.State.t()
+  def set_triggered(state, group_name, value \\ true) do
+    if group_name in Map.keys(state.crash_groups) do
```
mat-hek, line 88: why do we allow passing a non-existing crash group?

mat-hek (changes requested):

mat-hek:
and fix CI

lib/membrane/core/parent/child_life_controller.ex
andpodob: Why? Doesn't it make some possibility for race condition then?

lib/membrane/core/parent/child_life_controller/crash_group_handler.ex
andpodob: sorry, debugging leftover

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,57 +162,36 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+        if reason == {:shutdown, :membrane_crash_group_kill} do
+          raise Membrane.PipelineError,
+                "Child that was not a member of any crash group killed with :membrane_crash_group_kill."
+        else
+          Membrane.Logger.debug("""
+          Pipeline child crashed but was not a member of any crash group.
+          Terminating.
+          """)
+
+          propagate_child_crash(state)
```
mat-hek, line 193:
```suggestion
        {:error, :not_member} when reason == {:shutdown, :membrane_crash_group_kill} ->
          raise Membrane.PipelineError,
                "Child that was not a member of any crash group killed with :membrane_crash_group_kill."
        {:error, :not_member} ->
          Membrane.Logger.debug("""
          Pipeline child crashed but was not a member of any crash group.
          Terminating.
          """)

          propagate_child_crash(state)
```

lib/membrane/core/parent/child_life_controller/crash_group_handler.ex
```diff
@@ -17,35 +17,38 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupHandler do
...
-      members: children_names,
-      alive_members: children_pids
-    }
+    state =
+      Bunch.Access.update_in(state, [:crash_groups, group_name], fn maybe_group ->
+        if maybe_group do
+          %CrashGroup{
+            members: current_children_names,
+            alive_members_pids: current_alive_members
+          } = maybe_group
```
mat-hek, line 26:
```suggestion
      Bunch.Access.update_in(state, [:crash_groups, group_name], fn 
          %CrashGroup{
            members: current_children_names,
            alive_members_pids: current_alive_members
          } = group ->
```

lib/membrane/core/parent/child_life_controller/crash_group_handler.ex
```diff
@@ -17,35 +17,38 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupHandler do
...
-              %{existing_group | members: current_children_names ++ children_names, alive_members: current_alive_members ++ children_pids}
-            end
-          )
-    }
+          %CrashGroup{
+            maybe_group
+            | members: current_children_names ++ children_names,
+              alive_members_pids: current_alive_members ++ children_pids
+          }
+        else
```
mat-hek, line 33:
```suggestion
        nil ->
```

mat-hek (approved):

Issue #297, opened 2021-04-16
Add more telemetry events


mat-hek:
For example:
- [x] mailbox size
- [x] bitrate
- [x] buffers per second
- [x] processed events (in total)
- [x] processed caps (in total)

sheldak:
Processed events and processed caps added in [More telemetry events #308](https://github.com/membraneframework/membrane_core/pull/308)

PR #308, opened 2021-06-01
More telemetry events

sheldak:
Work in progress.

Current changes:
- `method` is now called `metric`
- `input_buffer_size_event` is now called `metric_event`
- added reporting `caps`
- added reporting `event`
- added possibility to disable reporting metrics

Adds two metrics from [Add more telemetry events #297](https://github.com/membraneframework/membrane_core/issues/297)

### DIFF
--- a/lib/membrane/core/element/caps_controller.ex
+++ b/lib/membrane/core/element/caps_controller.ex
@@ -6,15 +6,14 @@ defmodule Membrane.Core.Element.CapsController do
   use Bunch
 
   alias Membrane.{Caps, Pad}
-  alias Membrane.ComponentPath
   alias Membrane.Core.{CallbackHandler, InputBuffer}
   alias Membrane.Core.Child.PadModel
   alias Membrane.Core.Element.{ActionHandler, State}
   alias Membrane.Element.CallbackContext
-  alias Membrane.Telemetry
 
   require Membrane.Core.Child.PadModel
   require Membrane.Logger
+  require Membrane.Telemetry
 
   @doc """
   Handles incoming caps: either stores them in InputBuffer, or executes element callback.
@@ -67,16 +66,6 @@ defmodule Membrane.Core.Element.CapsController do
   end
 
   defp report_caps(log_tag) do
-    :telemetry.execute(
-      Telemetry.metric_event_name(),
-      %{
-        element_path:
-          ComponentPath.get_formatted() <>
-            "/" <> (log_tag || ""),
-        metric: "caps",
-        value: 1
-      },
-      %{}
-    )
+    Membrane.Telemetry.report_metric("caps", 1, log_tag)
   end
 end
--- a/lib/membrane/core/element/caps_controller.ex
+++ b/lib/membrane/core/element/caps_controller.ex
@@ -6,10 +6,12 @@ defmodule Membrane.Core.Element.CapsController do
   use Bunch
 
   alias Membrane.{Caps, Pad}
+  alias Membrane.ComponentPath
   alias Membrane.Core.{CallbackHandler, InputBuffer}
   alias Membrane.Core.Child.PadModel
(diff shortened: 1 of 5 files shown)


lib/membrane/core/element/caps_controller.ex
```diff
@@ -61,4 +65,18 @@ defmodule Membrane.Core.Element.CapsController do
         {{:error, reason}, state}
     end
   end
+
+  defp report_caps(log_tag) do
```
Qizot, line 69: Can you create a separate module for that to keep the telemery functions in one place? Would be easier to track all events and their reporting functions.

lib/membrane/core/element/event_controller.ex
```diff
@@ -170,4 +174,18 @@ defmodule Membrane.Core.Element.EventController do
 
   defp stream_event_to_callback(%Events.StartOfStream{}), do: :handle_start_of_stream
   defp stream_event_to_callback(%Events.EndOfStream{}), do: :handle_end_of_stream
+
+  defp report_event(log_tag) do
```
Qizot, line 178: Same as we report caps.

Qizot (changes requested):

lib/membrane/core/element/caps_controller.ex
sheldak: Maybe in `Membrane.Telemetry`? What do you think about it?

lib/membrane/telemetry.ex
```diff
@@ -11,32 +11,39 @@ defmodule Membrane.Telemetry do
...
 
+  @enable_metrics Application.compile_env(:membrane_core, :enable_metrics, false)
+
   @type event_name_t :: [atom(), ...]
 
   @typedoc """
   * element_path - element's process path with pad's name that input buffer is attached to
-  * method - input buffer's function call name
-  * value - current buffer's size
+  * metric - metric name
```
Qizot, line 35:
```suggestion
  * metric - metric's name
```

lib/membrane/core/element/caps_controller.ex
```diff
@@ -13,12 +13,15 @@ defmodule Membrane.Core.Element.CapsController do
...
   require Membrane.Core.Child.PadModel
   require Membrane.Logger
+  require Membrane.Telemetry
 
   @doc """
   Handles incoming caps: either stores them in InputBuffer, or executes element callback.
   """
   @spec handle_caps(Pad.ref_t(), Caps.t(), State.t()) :: State.stateful_try_t()
   def handle_caps(pad_ref, caps, state) do
+    report_caps(inspect(pad_ref))
```
Qizot, line 23: Do you need to wrap the `Membrane.Telemetry.report_metric` into a function? With this approach you are calling unnecessarily a function that will just return `:ok`. When using directly there won't event be code to run.

And I think if you just use it in a single place you don't have to wrap that call and it will look more descriptive.

lib/membrane/core/element/event_controller.ex
Qizot: This thread is old but do the same thing as with caps, you probably don't need this function.

Qizot (changes requested):

lib/membrane/telemetry.ex
```diff
@@ -55,4 +62,88 @@ defmodule Membrane.Telemetry do
...
+  """
+  defmacro report_metric(metric, value, log_tag) do
+    if @enable_metrics do
+      quote do
+        alias Membrane.ComponentPath
+
+        path = ComponentPath.get_formatted() <> "/" <> (unquote(log_tag) || "")
+
+        :telemetry.execute(
+          unquote(metric_event_name()),
```
Qizot, line 79: I guess you can put directly the array here instead of having it in function call.

Qizot (approved):
Change this last think and otherwise LGTM

PR #310, opened 2021-06-10
Release membrane_core v0.7.0

mickel8:
### Release "Simply Pro Cub"

| Project | New Version |
|---|---|
| membrane_core | `0.7.0` |

<sup>proudly automated with <a href="https://github.com/membraneframework/sebex">Sebex</a></sup>


lib/membrane/parent_spec.ex
```diff
@@ -160,6 +160,8 @@ defmodule Membrane.ParentSpec do
     end
   end
 
+  @opaque link_builder_t :: link_builder_t()
```
mat-hek, line 163: shouldn't that be

```suggestion
  @opaque link_builder_t :: LinkBuilder.t()
```

please check if dialyzer passes

mat-hek (changes requested):

mat-hek (approved):
If dialyzer passes, we can merge. Btw we could add it to CI

mat-hek (dismissed):

lib/membrane/telemetry.ex
```diff
@@ -53,6 +57,87 @@ defmodule Membrane.Telemetry do
...
+      end
+    end
+  end
+
+  @doc """
+  Macro for reporting new links.
+
+  Links are reported only when metrics reporting is enabled in the application using Membrane Core.
+  """
+  defmacro report_new_link(from, to) do
```
mat-hek, line 101: I would instead provide a generic `report_metric_lazy(metric, function, log_tag)` that would accept a function calculating value instead of a value itself. Reporting links could use that instead of having a separate macro for that purpose.

lib/membrane/telemetry.ex
```diff
@@ -53,6 +57,87 @@ defmodule Membrane.Telemetry do
...
+    if @enable_metrics do
+      quote do
+        alias Membrane.ComponentPath
+
+        path = ComponentPath.get_formatted() <> "/" <> (unquote(log_tag) || "")
+
+        :telemetry.execute(
+          [:membrane, :metric, :value],
+          %{
+            element_path: path,
```
mat-hek, line 75:
```suggestion
            component_path: path,
```

mat-hek (changes requested):

lib/membrane/telemetry.ex
sheldak: I can change that. It also requires changes in Reporter and Dashboard (due to change in DB I assume is necessary for consistence in naming - there is entire table called `element_paths` now, so I will change it to `component_paths`). 

Now, I think, is the best moment for any other naming changes if they are necessary, because I am already working on Reporter and Dashboard.

lib/membrane/telemetry.ex
```diff
@@ -94,50 +91,65 @@ defmodule Membrane.Telemetry do
...
+      calculate_measurement
+    )
+  end
 
-        require Membrane.Pad
+  @doc """
+  Reports new link connection being initialized in pipeline.
+  """
+  @spec report_new_link(Endpoint.t(), Endpoint.t()) :: :ok
+  def report_new_link(from, to) do
```
mat-hek, line 110: The point of making `report_metric` accept function was to make this private, as it's only used inside the core. So let's move that to `Membrane.Core.Parent.Link`.

mat-hek (changes requested):

lib/membrane/telemetry.ex
sheldak: Currently `report_measurement` accepts function. I wanted to separate metrics (events, caps, input buffer's sizes) from all measurements (both metrics and new links) so I have changed the name.

Okay, I can make `report_new_link` private and move to `Membrane.Core.Parent.Link`, but what about `report_metric` which is called from 3 places?

mat-hek: report_metric is generic, doesn't require awareness of core internals, and can be possibly used from outside of the core - so it can be public. Btw you could make two versions of report_metric - one accepting just a value, and the other accepting a function.

mat-hek (dismissed):

mat-hek (approved):

PR #312, opened 2021-06-15
Clarify callback_return_t and add typedoc

sax:
We traced through some code to try to describe the consequences of return specific values from callbacks, and we think this is correct.

`t:CallbackHandler.callback_return_t()` is in a `@moduledoc false` module, so does not end up linkable. We thought that making that module show up in documentation might have unintended consequences, so switched `c:handle_init/1` to use the local type. We also weren't sure how to describe `CallbackHandler` or put it into the proper place in the doc list.

Alternatively, if there was a good moduledoc that could be put into `Membrane.Core.CallbackHandler`, that could be where these types all get directed to and document.


mickel8:
I have merged your changes to our release branch :slightly_smiling_face:

PR #315, opened 2021-06-24
add crash group docs

andpodob:
Added documentation for Crash Groups.

### DIFF
--- a/lib/membrane/parent_spec.ex
+++ b/lib/membrane/parent_spec.ex
@@ -151,8 +151,7 @@ defmodule Membrane.ParentSpec do
   ```elixir
   @impl true
   def handle_crash_group_down(crash_group_id, ctx, state) do
-    Membrane.Logger.info("Crash group: #{inspect(crash_group_id)} is down.")
-    # do some stuff
+    # do some stuff in reaction to crash of group with id crash_group_id
   end
   ```
 
--- a/lib/membrane/parent_spec.ex
+++ b/lib/membrane/parent_spec.ex
@@ -123,30 +123,37 @@ defmodule Membrane.ParentSpec do
   clock. The pipeline clock is the default clock used by elements' timers.
   For more information see `Membrane.Element.Base.def_clock/1`.
 
-  # Crash Groups
-  Crash group is a logical entity that prevents the whole pipeline from crashing when one of its children crashes.
-  Internally crash groups utilizes process monitoring mechanism to detect child crash and thanks to that Membrane
-  doesn't influence value of `:trap_exit` flag.
+  ## Crash Groups
+  The crash group is a logical entity that prevents the whole pipeline from crashing when one of
+  its children crash.
 
   ## Adding children to the Crash Group
 
   ```elixir
   children = %{
-    {:endpoint_bin, endpoint_id} => %EndpointBin{
+    :some_element_1 => %SomeElement{
+      # ...
+    },
+    :some_element_2 => %SomeElement{
       # ...
     }
   }
 
-  spec = %ParentSpec{children: children, crash_group: {endpoint_id, :temporary}}
+  spec = %ParentSpec{children: children, crash_group: {group_id, :temporary}}
   ```
 
-  Crash group is defined by two element tuple, first element is group id which is of type `Membrane.Crashgroup.name_t()`, and the second is a mode. Currently we support only
-  `:temporary` which means that membrane will not make any attempts to restart crashed child.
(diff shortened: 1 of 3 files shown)


lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
   clock. The pipeline clock is the default clock used by elements' timers.
   For more information see `Membrane.Element.Base.def_clock/1`.
 
+  # Crash Groups
```
mat-hek, line 126: `#` shouldn't be used in docs, start with `##`. Generate the docs and check if they look good

lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
...
+  Crash group is a logical entity that prevents the whole pipeline from crashing when one of its children crashes.
+  Internally crash groups utilizes process monitoring mechanism to detect child crash and thanks to that Membrane
+  doesn't influence value of `:trap_exit` flag.
+
+  ## Adding children to the Crash Group
+
+  ```elixir
+  children = %{
+    {:endpoint_bin, endpoint_id} => %EndpointBin{
```
mat-hek, line 135: endpoint bin may be a bit confusing, use `some_element` or so

lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
   clock. The pipeline clock is the default clock used by elements' timers.
   For more information see `Membrane.Element.Base.def_clock/1`.
 
+  # Crash Groups
+  Crash group is a logical entity that prevents the whole pipeline from crashing when one of its children crashes.
+  Internally crash groups utilizes process monitoring mechanism to detect child crash and thanks to that Membrane
+  doesn't influence value of `:trap_exit` flag.
```
mat-hek, line 129: I wouldn't mention that, especially since we're going to change that

lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
...
+    {:endpoint_bin, endpoint_id} => %EndpointBin{
+      # ...
+    }
+  }
+
+  spec = %ParentSpec{children: children, crash_group: {endpoint_id, :temporary}}
+  ```
+
+  Crash group is defined by two element tuple, first element is group id which is of type `Membrane.Crashgroup.name_t()`, and the second is a mode. Currently we support only
+  `:temporary` which means that membrane will not make any attempts to restart crashed child.
```
mat-hek, line 144: I'd move that before the example

lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
...
+
+  In above snippet we create a new children - EndpointBin and we add it to crash group with id `endpoint_id`. When EndpointBin crashes the whole pipeline will still be alive.
```
mat-hek, line 146: use a tool like grammarly to check the spelling for you

```suggestion
  In the above snippet, we create new children - EndpointBin and we add it to the crash group with id `endpoint_id`. When EndpointBin crashes, the rest of the pipeline will still be alive.
```

lib/membrane/parent_spec.ex
```diff
@@ -123,6 +123,40 @@ defmodule Membrane.ParentSpec do
...
+
+  ```elixir
+  @impl true
+  def handle_crash_group_down(crash_group_id, ctx, state) do
+    # do some stuff in reaction to crash of group with id crash_group_id
+  end
+  ```
+
+  ## Limitations
+  At this moment Crash Groups are only useful for elements with dynamic pads.
```
mat-hek, line 159: mention that they're currently not supported within bins

mat-hek (changes requested):
There are some uncathegorized modules in docs, can you assign them also?

image

lib/membrane/parent_spec.ex
andpodob: are we?

andpodob: aren't we going to add dummy ancestor and stay with monitors?

mat-hek: IIRC the 'dummy ancestor' was to use links and trap exits

mat-hek (approved):

PR #317, opened 2021-07-12
Added init/termiante telemetry events

Qizot:
Besides reporting events mentioned in title also added `buffer`, `queue_len` and `bitrate` events, all of which are reported when the processing element sends buffers to its descendant.

### DIFF
--- a/lib/membrane/component_path.ex
+++ b/lib/membrane/component_path.ex
@@ -10,6 +10,24 @@ defmodule Membrane.ComponentPath do
   @key :membrane_path
 
   @doc """
+  Prepends a nonce string to given name and initializes the current path.
+
+  Nonce is used to guarantee path uniqueness between different application runs.
+  Prepended nonce is of format '([A-Za-z0-9+]) '.
+  """
+  @spec init_with_nonce(String.t()) :: :ok
+  def init_with_nonce(name) do
+    nonce =
+      :crypto.strong_rand_bytes(4)
+      |> Base.encode64()
+      |> String.trim_trailing("==")
+      |> String.replace("/", "O")
+
+    Process.put(@key, ["(#{nonce}) #{name}"])
+    :ok
+  end
+
+  @doc """
   Appends given name to the current path.
 
   If path has not been previously set then creates new one with given name.
--- a/lib/membrane/component_path.ex
+++ b/lib/membrane/component_path.ex
@@ -10,24 +10,6 @@ defmodule Membrane.ComponentPath do
   @key :membrane_path
 
   @doc """
-  Prepends a nonce string to given name and initializes the current path.
-
-  Nonce is used to guarantee path uniqueness between different application runs.
-  Prepended nonce is of format '([A-Za-z0-9+]) '.
-  """
-  @spec init_with_nonce(String.t()) :: :ok
-  def init_with_nonce(name) do
-    nonce =
-      :crypto.strong_rand_bytes(4)
-      |> Base.encode64()
-      |> String.trim_trailing("==")
-      |> String.replace("/", "O")
-
-    Process.put(@key, ["(#{nonce}) #{name}"])
-    :ok
-  end
-
-  @doc """
   Appends given name to the current path.
 
   If path has not been previously set then creates new one with given name.
--- a/lib/membrane/core/element.ex
+++ b/lib/membrane/core/element.ex
@@ -100,7 +100,7 @@ defmodule Membrane.Core.Element do
 
     :ok = ComponentPath.set_and_append(options.log_metadata[:parent_path] || [], name_str)
 
-    Telemetry.report_init(:element, ComponentPath.get())
+    Telemetry.report_init(:element)
 
     state =
       options
@@ -117,7 +117,7 @@ defmodule Membrane.Core.Element do
 
   @impl GenServer
   def terminate(reason, state) do
-    Telemetry.report_terminate(:element, ComponentPath.get())
+    Telemetry.report_terminate(:element)
 
     {:ok, _state} = LifecycleController.handle_shutdown(reason, state)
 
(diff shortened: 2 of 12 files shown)


lib/membrane/telemetry.ex
```diff
@@ -58,5 +74,30 @@ defmodule Membrane.Telemetry do
...
+  @compile {:inline, pipeline_terminate_event_name: 0}
+  def pipeline_terminate_event_name(), do: [:membrane, :pipeline, :terminate]
+
+  @spec bin_init_event_name() :: event_name_t()
+  @compile {:inline, bin_init_event_name: 0}
+  def bin_init_event_name(), do: [:membrane, :bin, :init]
+
+  @spec bin_terminate_event_name() :: event_name_t()
+  @compile {:inline, bin_terminate_event_name: 0}
+  def bin_terminate_event_name(), do: [:membrane, :_terminate, :terminate]
```
mat-hek, line 94:
```suggestion
  def bin_terminate_event_name(), do: [:membrane, :bin, :terminate]
```

lib/membrane/telemetry.ex
```diff
@@ -58,5 +74,30 @@ defmodule Membrane.Telemetry do
...
+
+  @spec element_init_event_name() :: event_name_t()
+  @compile {:inline, element_init_event_name: 0}
+  def element_init_event_name(), do: [:membrane, :element, :init]
+
+  @spec element_terminate_event_name() :: event_name_t()
+  @compile {:inline, element_terminate_event_name: 0}
+  def element_terminate_event_name(), do: [:membrane, :element, :terminate]
```
mat-hek, line 102: do we have to have all these functions? IMHO it's enough that these names are documented

mat-hek (changes requested):

lib/membrane/component_path.ex
```diff
@@ -9,6 +9,24 @@ defmodule Membrane.ComponentPath do
...
+    nonce =
+      :crypto.strong_rand_bytes(4)
+      |> Base.encode64()
+      |> String.trim_trailing("==")
+      |> String.replace("/", "O")
+
+    Process.put(@key, ["(#{nonce}) #{name}"])
+    :ok
+  end
+
```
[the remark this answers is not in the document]
Qizot: I can see that it has been closed and not merged.  I can apply that change here.

lib/membrane/telemetry.ex
Qizot: Sure ;)

lib/membrane/component_path.ex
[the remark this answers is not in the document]
Qizot: Nevermind, I see that the `System.pid()` gets added in timescale reporter.

lib/membrane/core/element.ex
```diff
@@ -212,6 +213,11 @@ defmodule Membrane.Core.Element do
 
   @impl GenServer
   def handle_info(message, state) do
+    Telemetry.report_metric(
+      :queue_len,
+      :erlang.process_info(self(), :message_queue_len) |> elem(1)
+    )
```
mat-hek, line 219: This should rather happen each time handle_info is called. I'd convert existing clauses to `do_handle_info` not to repeat that every clause

mat-hek (changes requested):

lib/membrane/core/element.ex
Qizot: I don't think that adding another function call for reporting the `queue_len` is worth it.

mat-hek: if you add `@compile :inline` to `do_handle_info` there will be no difference

Qizot: Ok, makes sense as the compile inline will work just like pattern matching on functions but it will internally create a `case` so we will end up with the same amount of pattern matching without any additional overhead.

mat-hek:
Can we also somehow solve the problem of bitrate and buffers at sinks? That metrics won't be reported for them or for any element that only receives data. Maybe we'd have `incoming_bitrate`, `outgoing_bitrate`, `incoming_buffers` and `outgoing_buffers`?

Qizot:
> Can we also somehow solve the problem of bitrate and buffers at sinks? That metrics won't be reported for them or for any element that only receives data. Maybe we'd have `incoming_bitrate`, `outgoing_bitrate`, `incoming_buffers` and `outgoing_buffers`?

We can either:
-  perform the measuring bitrate and number of buffers when buffers arrive on input pads. We will support a more vast range of elements (including elements that just consume data and does not produce, sinks, ect.) but again we will lose information about sources producing such buffers (from what I researched there is no an easy way to measure those 2 metrics for sources without duplicating the information for other types of elements)
- measure metrics for sinks in buffer controlelrs as I did and ignore the elements that does not produce data (I think that those elements should be sinks in the end)

mat-hek (changes requested):

lib/membrane/core/telemetry.ex
```diff
@@ -38,6 +38,7 @@ defmodule Membrane.Core.Telemetry do
   @doc """
   Reports metrics such as input buffer's size inside functions, incoming events and received caps.
   """
+  @compile {:inline, report_metric: 3}
```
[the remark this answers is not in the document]
Qizot: Changed to macros :)

lib/membrane/core/telemetry.ex
```diff
@@ -2,114 +2,146 @@ defmodule Membrane.Core.Telemetry do
...
+                0,
+                &(Membrane.Payload.size(&1.payload) + &2)
+              )
+        }
+      end
+
+    report_event(event, value)
+  end
+
+  def get_public_pad_name(pad) do
```
mat-hek, line 60: This should be private. Generate the docs and check if nothing's missing

mat-hek (changes requested):

lib/membrane/core/telemetry.ex
Qizot: The problem is that the macros can't access that when it is private.

mat-hek (approved):

mat-hek (approved):

PR #319, opened 2021-09-06
Allow to specify node when spawning bin or element.

ostinelli:
Adds the ability to distribute elements across a cluster. It will use node in `%ParentSpec{}` in `ChildLifeController`.

### DIFF
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -33,28 +33,40 @@ defmodule Membrane.Core.Bin do
   Returns the same values as `GenServer.start_link/3`.
   """
   @spec start_link(options_t, GenServer.options()) :: GenServer.on_start()
-  def start_link(options, process_options \\ []) do
-    do_start(:start_link, options, process_options)
-  end
+  def start_link(options, process_options \\ []),
+    do: start_link(node(), options, process_options)
+
+  @spec start_link(node(), options_t, GenServer.options()) :: GenServer.on_start()
+  def start_link(node, options, process_options),
+    do: do_start(node, :start_link, options, process_options)
 
   @doc """
   Works similarly to `start_link/2`, but does not link to the current process.
   """
   @spec start(options_t(), GenServer.options()) :: GenServer.on_start()
-  def start(options, process_options \\ []) do
-    do_start(:start, options, process_options)
-  end
+  def start(options, process_options \\ []),
+    do: start(node(), options, process_options)
 
-  defp do_start(method, options, process_options) do
+  @spec start(node(), options_t(), GenServer.options()) :: GenServer.on_start()
+  def start(node, options, process_options),
+    do: do_start(node, :start, options, process_options)
+
+  defp do_start(node, method, options, process_options) do
     if options.module |> Membrane.Bin.bin?() do
       Membrane.Logger.debug("""
-      Bin start link: name: #{inspect(options.name)}
+      Bin #{method}: #{inspect(options.name)}
+      node: #{node},
       module: #{inspect(options.module)},
       bin options: #{inspect(options.user_options)},
       process options: #{inspect(process_options)}
       """)
 
-      apply(GenServer, method, [Membrane.Core.Bin, options, process_options])
+      # rpc if necessary
+      if node == node() do
+        apply(GenServer, method, [Membrane.Core.Bin, options, process_options])
+      else
+        :rpc.call(node, GenServer, method, [Membrane.Core.Bin, options, process_options])
+      end
     else
       raise """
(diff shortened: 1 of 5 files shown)


lib/membrane/core/element.ex
```diff
@@ -47,27 +47,41 @@ defmodule Membrane.Core.Element do
...
       Element #{method}: #{inspect(name)}
+      node: #{node},
       module: #{inspect(module)},
       element options: #{inspect(user_options)},
       process options: #{inspect(process_options)}
       """)
 
-      apply(GenServer, method, [__MODULE__, options, process_options])
+      # rpc if node set
+      if node == nil do
```
Qizot, line 80: same case as above with `Membrane.Core.Bin`

lib/membrane/core/bin.ex
```diff
@@ -33,28 +33,40 @@ defmodule Membrane.Core.Bin do
...
+      Bin #{method}: #{inspect(options.name)}
+      node: #{node},
       module: #{inspect(options.module)},
       bin options: #{inspect(options.user_options)},
       process options: #{inspect(process_options)}
       """)
 
-      apply(GenServer, method, [Membrane.Core.Bin, options, process_options])
+      # rpc if node set
+      if node == nil do
```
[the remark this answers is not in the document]
Qizot: EDIT: I know that `nil` is also an atom and the question is if we want to treat `nil` and `node()` (current node) the same way.

lib/membrane/core/bin.ex
```diff
@@ -33,28 +33,40 @@ defmodule Membrane.Core.Bin do
...
 
   @doc """
   Works similarly to `start_link/2`, but does not link to the current process.
   """
   @spec start(options_t(), GenServer.options()) :: GenServer.on_start()
-  def start(options, process_options \\ []) do
-    do_start(:start, options, process_options)
-  end
+  def start(options, process_options \\ []),
+    do: start(node(), options, process_options)
```
Qizot, line 48: I guess that it should be a `nil` to avoid spawning it via rpc call?

Qizot (changes requested):

lib/membrane/core/bin.ex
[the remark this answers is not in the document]
ostinelli: > Can it ever be `nil`? If so then the function spec should account for that or we should probably raise.

It can be `nil` because `ChildLifeController` calls `StartupHandler.start_children/5` with the `node` parameter coming from the optional `node` element in `%ParentSpec{}` from [here](https://github.com/geometerio/membrane_core/commit/3741c61ff8ef15e527eed1646a47e479b0e17551#diff-9fc8a1f82166b1f51259297d76ef5b53ffce74fc633eb00060190baf97edd403R41).

> I think we should also check if the node equals current node and start it without the `rpc` as you do in `nil` case.

I conformed to Erlang's `rpc` module that doesn't do this check, but I'm totally fine in adding this.

lib/membrane/core/bin.ex
ostinelli: I'm adding the suggested check so this can stay as is.

Qizot (approved):

lib/membrane/parent_spec.ex
```diff
@@ -238,7 +238,8 @@ defmodule Membrane.ParentSpec do
           links: links_spec_t,
           crash_group: crash_group_spec_t(),
           stream_sync: :sinks | [[Child.name_t()]],
-          clock_provider: Child.name_t()
+          clock_provider: Child.name_t(),
+          node: node()
```
mat-hek, line 242:
```suggestion
          clock_provider: Child.name_t(),
          node: node() | nil
```

ostinelli: Indeed! Thanks for catching it.

mat-hek: Actually clock_provider can be `nil` either, maybe let's fix them both ;)

mat-hek: @ostinelli bump ;)

ostinelli: I wasn't aware you were waiting for me! Will update asap.

mat-hek (approved):

mat-hek:
@sax @ostinelli can you please rebase?

ostinelli:
> @sax @ostinelli can you please rebase?

Do you mean squash the commits?

mat-hek:
@ostinelli I mean update to the membrane master, which changed since the PR was created. I don't expect any conflicts, but I cannot rebase since it's geometer's repo and I don't have rights

image

mat-hek (approved):

PR #321, opened 2021-09-07
Handle removing elements that are crashgroup members

daniel-jodlos:


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -234,6 +235,18 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(child_pid, state) do
+      {_result, state} =
+        CrashGroupHandler.remove_member_of_crash_group(state, group.name, child_pid)
+        |> CrashGroupHandler.remove_crash_group_if_empty(group.name)
+
+      state
+    else
+      {:error, :not_member} -> state
+    end
+  end
```
mat-hek, line 248: can we refactor that so it could be used in both clauses of `handle_child_death`?

mat-hek (approved):

PR #322, opened 2021-09-15
Reduce warnings when elements are shut down

sax:
Known shutdown reasons still warn, but without including the entire element state.

### DIFF
--- a/lib/membrane/core/element/lifecycle_controller.ex
+++ b/lib/membrane/core/element/lifecycle_controller.ex
@@ -20,6 +20,7 @@ defmodule Membrane.Core.Element.LifecycleController do
 
   @safe_shutdown_reasons [
     {:shutdown, :child_crash},
+    {:shutdown, :membrane_crash_group_kill},
     {:shutdown, :parent_crash}
   ]
 
--- a/lib/membrane/core/element/lifecycle_controller.ex
+++ b/lib/membrane/core/element/lifecycle_controller.ex
@@ -77,7 +77,7 @@ defmodule Membrane.Core.Element.LifecycleController do
         Membrane.Logger.debug("Terminating element, reason: #{inspect(reason)}")
 
       reason in @safe_shutdown_reasons ->
-        Membrane.Logger.warn("""
+        Membrane.Logger.debug("""
         Terminating element possibly not prepared for termination as it was in state #{inspect(playback_state)}.
         Reason: #{inspect(reason)}"
         """)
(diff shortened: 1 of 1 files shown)


jasonwc (dismissed):
LGTM!

lib/membrane/core/element/lifecycle_controller.ex
```diff
@@ -18,6 +18,12 @@ defmodule Membrane.Core.Element.LifecycleController do
   require Membrane.Core.Playback
   require Membrane.Logger
 
+  @safe_shutdown_reasons [
+    {:shutdown, :child_crash},
```
mat-hek, line 22: This shouldn't be the case in element, as it doesn't have children

lib/membrane/core/element/lifecycle_controller.ex
```diff
@@ -66,14 +72,22 @@ defmodule Membrane.Core.Element.LifecycleController do
...
-      """)
+    cond do
+      playback_state == :terminating ->
+        Membrane.Logger.debug("Terminating element, reason: #{inspect(reason)}")
+
+      reason in @safe_shutdown_reasons ->
+        Membrane.Logger.warn("""
+        Terminating element possibly not prepared for termination as it was in state #{inspect(playback_state)}.
+        Reason: #{inspect(reason)}"
+        """)
```
mat-hek, line 83: I'd make this a debug too

```suggestion
        Membrane.Logger.debug("""
        Terminating element possibly not prepared for termination as it was in state #{inspect(playback_state)}.
        Reason: #{inspect(reason)}"
        """)
```

lib/membrane/core/element/lifecycle_controller.ex
sax: I see these errors in my logs when a `Membrane.RTP.Parser` crashes in the rtp plugin:

```
[warn] [{:srtcp_encryptor, #Reference<0.4160045054.241172482.20287>}] Terminating element possibly not prepared for termination as it was in state :playing.
Reason: {:shutdown, :child_crash}
State: .....
```

This was why I originally included this reason.

mat-hek: ok, that seems weird, let me investigate this ;)

mat-hek: Ok, I have realized that it's probably when a sibling crashes. Seems reasonable 👍

PR #323, opened 2021-09-17
CrashGroupDown now holds information about child that crashed first

agga1:


lib/membrane/bin/callback_context/crash_group_down.ex
```diff
@@ -4,5 +4,6 @@ defmodule Membrane.Bin.CallbackContext.CrashGroupDown do
   when a crash group is down.
   """
   use Membrane.Core.Bin.CallbackContext,
-    members: [Membrane.Child.name_t()]
+    members: [Membrane.Child.name_t()],
+    first_crashed_child: Membrane.Child.name_t()
```
mat-hek, line 8: I'd call this in a way indicating that the crash was actually triggered by the given child. Some ideas that come to my mind are crash_initiator, failed_child or guilty child :P

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,14 +162,22 @@ defmodule Membrane.Core.Parent.ChildLifeController do
   end
 
   def handle_child_death(pid, reason, state) do
-    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state) do
+    with {:ok, group} <- CrashGroupHandler.get_group_by_member_pid(pid, state),
+         {:ok, child_name} <- child_by_pid(pid, state) do
       {result, state} =
         crash_all_group_members(group, state)
+        |> set_first_crashed_child_if_needed(group, child_name)
```
mat-hek, line 169: The child_name could be an argument to crash_all_group_members, and then it could be passed to CrashGroupHandler.set_trigerred. I think it would make this a little bit less verbose

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
agga1: ok, good idea.
However [set_triggered](https://github.com/membraneframework/membrane_core/blob/master/lib/membrane/core/parent/child_life_controller/crash_group_handler.ex#L85)
 currently accepts value to be set (with "true" as default), so theoretically it can be called to set "triggered" as false -
can I remove this argument, as it isn't used anywhere, or should I make separate clauses for "false" and "true" scenarios?

mat-hek: I think it can be removed

mat-hek (approved):

PR #338, opened 2021-10-15
Reduce garbage, reductions when pulling buffers

sax:
My measurements are imprecise, but with a video room of 4 participants I think I'm seeing 25-100% CPU reduction with these changes. This should reduce the number of reductions, as well as some garbage generation.


lib/membrane/core/child/pad_model.ex
```diff
@@ -91,29 +86,57 @@ defmodule Membrane.Core.Child.PadModel do
...
+
+  @spec get_data!(Child.state_t(), Pad.ref_t()) :: Pad.Data.t() | any
+  def get_data!(state, pad_ref) do
+    {:ok, pad_data} = get_data(state, pad_ref)
+    pad_data
   end
 
   @spec get_data!(Child.state_t(), Pad.ref_t(), keys :: atom | [atom]) :: Pad.Data.t() | any
-  def get_data!(state, pad_ref, keys \\ []) do
+  def get_data!(state, pad_ref, keys) do
```
mat-hek, line 124: why?

lib/membrane/core/child/pad_model.ex
```diff
@@ -91,29 +86,57 @@ defmodule Membrane.Core.Child.PadModel do
...
           Type.stateful_t(:ok | unknown_pad_error_t, Child.state_t())
   def set_data(state, pad_ref, keys \\ [], value) do
-    with {:ok, state} <- {assert_instance(state, pad_ref), state} do
-      state
-      |> Bunch.Access.put_in(data_keys(pad_ref, keys), value)
-      ~> {:ok, &1}
+    case assert_instance(state, pad_ref) do
+      :ok ->
+        state
+        |> Bunch.Access.put_in(data_keys(pad_ref, keys), value)
```
mat-hek, line 135:
```suggestion
        put_in(data_keys(pad_ref, keys), value)
```
Bunch.Access seems not to help there and it's an additional function call

lib/membrane/core/child/pad_model.ex
```diff
@@ -217,13 +268,39 @@ defmodule Membrane.Core.Child.PadModel do
...
     constraints |> Enum.all?(fn {k, v} -> data[k] === v end)
   end
 
+  @spec data_keys(Pad.ref_t()) :: [atom]
+  defp data_keys(pad_ref), do: [:pads, :data, pad_ref]
+
   @spec data_keys(Pad.ref_t(), keys :: atom | [atom]) :: [atom]
-  defp data_keys(pad_ref, keys \\ []) do
-    [:pads, :data, pad_ref | Bunch.listify(keys)]
+  defp data_keys(pad_ref, keys)
```
mat-hek, line 297:
```suggestion
  @compile {:inline, data_keys: 2}
  defp data_keys(pad_ref, keys)
```

lib/membrane/core/child/pad_model.ex
```diff
@@ -217,13 +268,39 @@ defmodule Membrane.Core.Child.PadModel do
     state
   end
 
+  @spec constraints_met?(Pad.Data.t(), [
```
mat-hek, line 271: invalid spec

mat-hek (changes requested):

lib/membrane/core/callback_handler.ex
```diff
@@ -109,25 +108,36 @@ defmodule Membrane.Core.CallbackHandler do
...
+    |> parse_callback_result(module, :handle_init, handler_params)
+  end
 
-    args =
-      if handler_params.state do
-        args ++ [state |> Map.fetch!(:internal_state)]
-      else
-        args
-      end
+  defp exec_callback(
```
mat-hek, line 117: We should consider making it a macro

lib/membrane/core/callback_handler.ex
```diff
@@ -173,7 +184,11 @@ defmodule Membrane.Core.CallbackHandler do
     {{:ok, actions}, new_internal_state}
   end
 
-  defp parse_callback_result({{:error, reason}, new_internal_state}, module, cb, %{state: true}) do
+  defp parse_callback_result({:error, reason}, module, cb, %{state: false}) do
```
mat-hek, line 187: Here we rely on the handler params, while in `exec_callback` we check whether the callback is handle_init

mat-hek (changes requested):

lib/membrane/core/child/pad_model.ex
sax: When I ran eprof on a video call for 10 seconds, I noticed that functions wrapping and unwrapping lists were among the most called functions. I realized that this default was only added so that it could be thrown away, and that fixing the callers to call the right function removes not only the need for wrapping/unwrapping, but also the need for this default.

lib/membrane/core/child/pad_model.ex
```diff
@@ -91,29 +86,57 @@ defmodule Membrane.Core.Child.PadModel do
     |> Map.new()
   end
 
+  @spec get_data(Child.state_t(), Pad.ref_t()) :: {:ok, Pad.Data.t() | any} | unknown_pad_error_t
+  def get_data(%{pads: %{data: data}}, pad_ref)
+      when is_map_key(data, pad_ref) do
+    data
+    |> Map.get(pad_ref)
```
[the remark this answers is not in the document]
sax: The simple fix (just pushed) uses `Map.fetch/2` to remove one of the accesses in this function. The idea of replacing this with macros is interesting, but I don't think I can wrap my head around the implementation at the moment.

I was also thinking that moving pad data into ETS could be a solution that make access and updates cheaper. I think it would be good to measure that before committing to it, though.

lib/membrane/core/callback_handler.ex
sax: Yeah, I wasn't sure about this. The main reason I added this was because the callers who care about it seem to all put `%{state: false}` into `exec_and_handle_callback/5` when they care about it. I see now that only happens for `:handle_init`, so...

I see what you're saying. I'll push a commit in a sec that catches `:handle_init` directly. This removes the need for `%{state: false}` everywhere.

sax:
@mat-hek I *think* these changes cover all your suggestions. Please take another look when you get a chance.

I have not had a chance to look at rebasing master, so this might need some updating related to what you've been working on.

sax:
Actually, since I realized it makes more sense to squash commit this all back onto master when ready, I just updated this branch with a merge commit.

lib/membrane/core/child/pad_model.ex
mat-hek: Oh I haven't noticed you added get_data!/2

lib/membrane/core/child/pad_model.ex
```diff
@@ -133,13 +153,12 @@ defmodule Membrane.Core.Child.PadModel do
...
-    with {:ok, state} <- {assert_instance(state, pad_ref), state},
-         {:ok, state} <-
-           state
-           |> Bunch.Access.get_and_update_in(data_keys(pad_ref, keys), f) do
-      {:ok, state}
-    else
-      {{:error, reason}, state} -> {{:error, reason}, state}
+    case assert_instance(state, pad_ref) do
+      :ok ->
+        state |> Bunch.Access.get_and_update_in(data_keys(pad_ref, keys), f)
```
mat-hek, line 158: Let's remove `Bunch.Access` from wherever possible

mat-hek:
Looks great! I'd believe removing `Bunch.Access` calls should be fairly simple and save us some overhead

sax:
This last commit removes the explicit calls to `Bunch.Access`, though it still depends on `use Bunch.Access` being added to the structs. My guess is that this isn't going to help too much, since the compiler might be smart enough to compile out the intermediary calls. It doesn't hurt though.

It seems like removing `Bunch.Access` itself will be a bigger refactor, and since we'll be modifying nested structs/maps, might not give us too much of a gain until operations can be batched. 

Could that be a done as a separate issue? I was thinking that moving these to different data structures might be more beneficial. I know that records were not a huge improvement (and I don't see a great alternative in erlang), but maybe flattening out the data and using ETS could improve update times with acceptable impact to reads. Alternatively, a small C-based mutable data structure using a NIF could speed things up? Anyways, both of those seem larger than this pull request.

mat-hek:
@sax I meant just to remove the `Bunch.Access.*_in` calls, just like you did ;) this will save some overhead, as it will remove the intermediary call before `Kernel.*_in`. Not using elixir's Access is a bigger refactor indeed. We did some research and it seems like using elixir's pattern matching and map put `%{map | key: value}` syntax performs great while providing immutability, so the plan is to create some macros generating that. Of course, that's beyond this PR

mat-hek (dismissed):

sax:
Yay!

mat-hek:
If you could just rebase to master, we updated credo there so the checks should start passing

sax:
Since I started doing merges, I decided to continue with another. Hopefully this can be squashed without too much trouble.

mat-hek (approved):

sax:
Thank you!

PR #344, opened 2021-11-02
bump version to 0.8.0

andpodob:


mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Membrane.Mixfile do
   use Mix.Project
 
-  @version "0.7.0"
+  @version "0.8.0"
```
bblaszkow06, line 4: Update README as well

bblaszkow06 (changes requested):

bblaszkow06 (approved):

PR #352, opened 2021-11-18
fix unknown node error on distribution changes

### LINKED ISSUE #346: Error when starting pipeline that includes an element using C Nodes
Before the introduction of API allowing to start membrane elements on other node starting the pipeline that was spawning elements utilizing C Nodes worked just fine, if it was started with non-distributed version it got upgraded to distributed one.

Right now there is a race condition when application is being started with non-distributed erlang and some new element using C Nodes gets spawned. Application starts with node `:nonode@nohost` and the pipeline will try to spawn
a child on that node as well, but once the element with C Node gets started the distribution will dynamically change and the node `:nonode@nohost` will no longer be available. At some point it will case the pipeline to crash as there will be some nodes variables competing `:nonode@nohost` vs newly generated node name.

The quick solution to that is to always start the appliation in distributed environment when we know there is an element using C Node API but it is a great inconvenience for other users using the framework as they probably should not be aware of such tiny details. 

Such case happened for the first time when using `membrane_ice_plugin` which can operate in two modes:
- using C Nodes
- using NIFs

where C Nodes version is a default one, which means a person using the plugin will by default always encounter the error unless he intentionally uses the distribution.

mat-hek:
closes #346 
cc @ostinelli

### DIFF
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -17,6 +17,7 @@ defmodule Membrane.Core.Bin do
   @type options_t :: %{
           name: atom,
           module: module,
+          node: node | nil,
           parent: pid,
           user_options: Membrane.Bin.options_t(),
           parent_clock: Membrane.Clock.t(),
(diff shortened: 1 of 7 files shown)


mickel8 (approved):
As we disscussed it might be worth to change bundlex behaviour so that it doesn't change node to distributed one dinamically

lib/membrane/core/bin.ex
```diff
@@ -33,40 +34,33 @@ defmodule Membrane.Core.Bin do
...
+    do: do_start(:start, options)
 
-  @spec start(node(), options_t(), GenServer.options()) :: GenServer.on_start()
-  def start(node, options, process_options),
-    do: do_start(node, :start, options, process_options)
+  defp do_start(method, options) do
+    %{module: module, name: name, node: node, user_options: user_options} = options
 
-  defp do_start(node, method, options, process_options) do
     if options.module |> Membrane.Bin.bin?() do
```
andpodob, line 51:
```suggestion
    if module |> Membrane.Bin.bin?() do
```

andpodob (approved):

PR #357, opened 2021-11-26
Fix failing tests.

marcinwasowicz:
Fixes tests failing due to changes introduced in original branch.

### DIFF
--- a/test/support/bin/test_bins.ex
+++ b/test/support/bin/test_bins.ex
@@ -221,7 +221,12 @@ defmodule Membrane.Support.Bin.TestBins do
     end
 
     @impl true
-    def handle_notification(notification, _element, _ctx, state) do
+    def handle_notification(
+          %Membrane.Testing.Notification{payload: _payload} = notification,
+          _element,
+          _ctx,
+          state
+        ) do
       {{:ok, notify: notification}, state}
     end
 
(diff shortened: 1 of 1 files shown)


test/support/bin/test_bins.ex
```diff
@@ -220,6 +220,16 @@ defmodule Membrane.Support.Bin.TestBins do
       {{:ok, spec: spec}, state}
     end
 
+    @impl true
+    def handle_notification(
+          %Membrane.Testing.Notification{payload: _payload} = notification,
```
FelonEkonom, line 225: What is the reason of pattern matching, if struct ` %Membrane.Testing.Notification{}` contains field `payload`, if it is enforced key?

test/support/bin/test_bins.ex
```diff
@@ -220,6 +220,16 @@ defmodule Membrane.Support.Bin.TestBins do
...
+          _element,
+          _ctx,
+          state
+        ) do
+      {{:ok, notify: notification}, state}
+    end
+
```
FelonEkonom, line 232: If stopping forwarding notifications in Bins causes fail of some tests, that are not explicitly related to forwarding notifications, I would rather consider the correctness of these tests or maybe unobvious and undesirable behaviors of Bins caused by this change, then just override the changed callback in implementation used in tests

test/support/bin/test_bins.ex
```diff
@@ -220,6 +220,20 @@ defmodule Membrane.Support.Bin.TestBins do
...
+    def handle_notification(_notification, _element, _ctx, state) do
+      {:ok, state}
+    end
```
andpodob, line 235: As we discussed on today's check-in, let's make `test_bin` to forward all notifications.

test/support/bin/test_bins.ex
```diff
@@ -257,6 +271,20 @@ defmodule Membrane.Support.Bin.TestBins do
...
+
```
andpodob, line 287: Same here.

andpodob (changes requested):

PR #363, opened 2021-12-18
Send notification to parent when end of stream or start of stream encountered in child element.

andpodob:


lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -65,9 +78,16 @@ defmodule Membrane.Core.Parent.MessageDispatcher do
     |> noreply(state)
   end
 
-  defp inform_parent(state, msg, msg_params) do
-    if not pipeline?(state) and state.parent_pid,
-      do: Message.send(state.parent_pid, msg, [state.name | msg_params])
+  defp maybe_notify_parent(event, path_to_element, pad_ref, state) do
```
mat-hek, line 81: as we discussed, this should be removed

mat-hek (changes requested):

lib/membrane/core/parent/message_dispatcher.ex
andpodob: I think we agreed to temporarily replace propagating the stream management event with the notification.

mat-hek (approved):

PR #365, opened 2021-12-21
More fine-grained control over emitted metrics

Qizot:
This PR introduces more fine-grained control over which metrics reporting gets compiled along with the core instead of turning them on and off all at once. This change allows having e.g. elements `init/terminate turned on all the time without any performance penalty or in case if we simply don't care about specific metrics (which again can be quite heavy).


andpodob (approved):
LGTM!

PR #376, opened 2022-01-03
Added log metadata when reporting init

Qizot:
The purpose of this PR is to add logger's metadata when reporting events such as `init/terminate` so that the element being created/terminated can be more easily distinguished beside having the already existing component's path, or so that it will carry arbitrary information necessary helpful for further processing.

### DIFF
--- a/lib/membrane/core/pipeline.ex
+++ b/lib/membrane/core/pipeline.ex
@@ -17,6 +17,13 @@ defmodule Membrane.Core.Pipeline do
     :ok = Membrane.ComponentPath.set([pipeline_name])
     :ok = Membrane.Logger.set_prefix(pipeline_name)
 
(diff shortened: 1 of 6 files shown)


lib/membrane/core/pipeline.ex
```diff
@@ -17,6 +17,13 @@ defmodule Membrane.Core.Pipeline do
     :ok = Membrane.ComponentPath.set([pipeline_name])
     :ok = Membrane.Logger.set_prefix(pipeline_name)
 
+    # NOTE: this would be super helpful to get the metadata inside the `report_init`
```
Qizot, line 20: I am not sure about this one

andpodob (approved):
LGTM!

PR #377, opened 2022-01-12
Bypass invalid table generation in pad docs

### LINKED ISSUE #356: Caps description doesn't render correctly in hex docs
It seems that some HTML - specific characters are being replaced with their HTML codes, including HTML tags
Zrzut ekranu 2021-11-22 o 16 53 15

bblaszkow06:
Since `ex_doc` currently doesn't handle HTML inside a table, we can work around that problem by generating HTML for the table manually

Closes #356


mickel8:
@andpodob said he will do CR as I don't feel like a suitable person to review changes in core

andpodob (approved):

PR #380, opened 2022-01-14
Update value returned by handle_init/1 in Membrane.Element.Base

FelonEkonom:
Old value returned by callback caused unnecessary Dialyzer warnings


andpodob (approved):
LGTM!

andpodob (approved):

PR #382, opened 2022-01-18
Endpoint

FelonEkonom:
Implementation of new type of Membrane element - `Membrane.Endpoint`


lib/membrane/core/element/demand_handler.ex
```diff
@@ -30,7 +30,7 @@ defmodule Membrane.Core.Element.DemandHandler do
       `handle_delayed_demands`.
     * If element isn't supplying demand at the moment `handle_demand` is invoked right away, and it will
       invoke handle_demand callback, which will probably return :redemand and :buffers actions and in
```
mat-hek, line 32: This was poorly written already, let's fix it

```suggestion
    * If the element isn't supplying demand at the moment and there's some unsupplied demand on the given 
      output, `handle_demand` is invoked right away, so that the demand can be synchronously supplied.
```

lib/membrane/core/element/demand_handler.ex
```diff
@@ -30,7 +30,7 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-      that way source will synchronously supply demand.
+      that way source or endpoint will synchronously supply demand.
```
mat-hek, line 33:
```suggestion
```

mat-hek (approved):

mat-hek (approved):

PR #385, opened 2022-02-01
add try-catch to the Message.call/5 implementation

andpodob:
This PR refactors `Message.call` to return an error instead of exiting when the receiver process is not alive.

### DIFF
--- a/lib/membrane/core/message.ex
+++ b/lib/membrane/core/message.ex
@@ -40,6 +40,11 @@ defmodule Membrane.Core.Message do
     end
   end
 
+  @spec call!(GenServer.server(), type_t, args_t, opts_t, timeout()) :: term()
+  def call!(pid, type, args \\ [], opts \\ [], timeout \\ 5000) do
+    GenServer.call(pid, message(type: type, args: args, opts: opts), timeout)
+  end
+
   @spec for_pad(t()) :: Pad.ref_t()
   def for_pad(message(opts: opts)), do: Keyword.get(opts, :for_pad)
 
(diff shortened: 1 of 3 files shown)


lib/membrane/core/message.ex
```diff
@@ -32,7 +32,12 @@ defmodule Membrane.Core.Message do
 
   @spec call(GenServer.server(), type_t, args_t, opts_t, timeout()) :: term()
   def call(pid, type, args \\ [], opts \\ [], timeout \\ 5000) do
-    GenServer.call(pid, message(type: type, args: args, opts: opts), timeout)
+    try do
+      GenServer.call(pid, message(type: type, args: args, opts: opts), timeout)
```
mat-hek, line 36:
```suggestion
      result = GenServer.call(pid, message(type: type, args: args, opts: opts), timeout)
      {:ok, result}
```

And we should handle that properly wherever it's used. We could possibly keep the old version of `call` named `call!` too.

mat-hek (changes requested):

lib/membrane/core/message.ex
andpodob: Currently call is not wrapping the result in the tuple `{:ok, result}`. Why do you want to change that? Also as we noticed during the last discussion on that, we are using `Message.call` only in 4 places in the whole codebase, I don't see a clear benefit of using `call!` over matching against the expected result. 
This code would end up very weird with this wrapping approach:
https://github.com/membraneframework/membrane_core/blob/master/lib/membrane/core/parent/child_life_controller/link_handler.ex#L270
The only place that requires an additional handling is here:
https://github.com/membraneframework/membrane_core/blob/master/lib/membrane/core/bin/pad_controller.ex#L180

lib/membrane/core/message.ex
```diff
@@ -32,6 +32,16 @@ defmodule Membrane.Core.Message do
 
   @spec call(GenServer.server(), type_t, args_t, opts_t, timeout()) :: term()
```
mat-hek, line 33: I'd list the error explicitly in the return type

mat-hek (approved):

PR #388, opened 2022-02-09
Update link_handler.ex

andpodob:
It seems that `%Link{}` has only `from` and `to` fields according to the type spec.
https://github.com/membraneframework/membrane_core/blob/master/lib/membrane/core/parent/link.ex


mat-hek:
Now it does, but more fields may be added, so I'd leave it as is because the current version is more future proof

PR #389, opened 2022-02-10
raise instead of warn when static pads are unlinked

andpodob:
As we discussed, it's better to follow the **fail-fast** principle and raise instead of just printing a warning in that case.


lib/membrane/core/child/pad_controller.ex
```diff
@@ -89,10 +89,10 @@ defmodule Membrane.Core.Child.PadController do
       )
 
     unless Enum.empty?(static_unlinked_pads) do
-      Membrane.Logger.warn("""
+      raise """
```
mat-hek, line 92: Seems like we could raise the `Membrane.LinkError` here

mat-hek (approved):

PR #392, opened 2022-02-14
Meaningful error when demand is not supported by pad

bblaszkow06:


lib/membrane/core/element/demand_handler.ex
```diff
@@ -94,6 +94,17 @@ defmodule Membrane.Core.Element.DemandHandler do
 
     pad_data = state |> PadModel.get_data!(pad_ref)
 
+    if pad_data.input_queue == nil do
```
mat-hek, line 97: I'd do this in action handler upon handling `demand` and `redemand` actions and match on `%{mode: :pull, demand_mode: :auto}` than check if `input_queue` is `nil`

mat-hek (changes requested):

PR #395, opened 2022-02-15
don't report toilet overflow multiple times

mat-hek:


lib/membrane/core/element/demand_handler.ex
```diff
@@ -132,12 +132,20 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-    Toilet.fill(toilet, buf_size)
-    state
+
+    case Toilet.fill(toilet, buf_size) do
+      :ok ->
+        state
+
+      :overflow ->
+        # if the toilet has overflowed, we remove it so it didn't overflow again
+        # and let the parent handle that situation by unlinking this output pad or crashing
```
bblaszkow06, line 146: Something is off with the tenses, didn't refer to the past
```suggestion
        # if the toilet has overflowed, we remove it so it won't overflow again
        # and let the parent handle that situation by unlinking this output pad or crashing
```

bblaszkow06 (approved):

PR #400, opened 2022-02-25
Reliable unlinking/case 3

andpodob:
When the spec that is being started contains bins the linking involves asynchronous operations. If at the same time the playback state change is requested it might happen that some pads remain unlinked (especially static pads because those are required to be linked when transitioning to state :playing).

A possible solution can be to check if some specs are being linked in the playback_handler while it is handling a request for a playback state change. If the list of :pending_specs is not empty the request for playback state change should be delayed and resumed after linking all :pending_specs is finished.


lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -88,7 +88,25 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
...
+      {:ok, state} =
+        if state.delayed_playback_change do
+          Membrane.Core.PlaybackHandler.change_playback_state(
+            state.delayed_playback_change,
+            Membrane.Core.Parent.LifecycleController,
+            %{state | delayed_playback_change: nil}
+          )
+        else
+          {:ok, state}
+        end
```
mat-hek, line 104: Some other specs may not be linked yet. Maybe we could just check if there are any pending specs?

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller/link_handler.ex
andpodob: Added check to verify if the `pending_specs` map is empty.

lib/membrane/core/playback_handler.ex
```diff
@@ -91,18 +91,26 @@ defmodule Membrane.Core.PlaybackHandler do
...
+      else
+        {:ok, %{state | playback: playback}}
+      end
     else
-      {:ok, %{state | playback: playback}}
+      if Component.is_parent?(state) do
+        {:ok, %{state | delayed_playback_change: new_playback_state}}
+      else
+        {:ok, state}
+      end
```
mat-hek, line 113: It's quite an ifology here, can't we refactor it somehow? `cond` perhaps?

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -88,7 +88,25 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
   def proceed_spec_linking(spec_ref, state) do
     spec_data = Map.fetch!(state.pending_specs, spec_ref)
     {spec_data, state} = do_proceed_spec_linking(spec_ref, spec_data, state)
-    put_in(state, [:pending_specs, spec_ref], spec_data)
+
+    if spec_data.status == :linked do
+      state = put_in(state, [:pending_specs], Map.delete(state.pending_specs, spec_ref))
```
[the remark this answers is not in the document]
andpodob: Done.

lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -88,7 +88,25 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
...
+          Membrane.Core.PlaybackHandler.change_playback_state(
+            state.delayed_playback_change,
+            Membrane.Core.Parent.LifecycleController,
+            %{state | delayed_playback_change: nil}
+          )
+        else
+          {:ok, state}
+        end
+
+      state
```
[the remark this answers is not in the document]
andpodob: Done.

lib/membrane/core/parent/child_life_controller/link_handler.ex
```diff
@@ -88,7 +88,25 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkHandler do
...
+    if spec_data.status == :linked do
+      {_spec_data, state} = pop_in(state, [:pending_specs, spec_ref])
+
+      if state.delayed_playback_change != nil and Enum.empty?(state.pending_specs) do
+        {:ok, state} =
+          Membrane.Core.PlaybackHandler.change_playback_state(
+            state.delayed_playback_change,
+            Membrane.Core.Parent.LifecycleController,
+            %{state | delayed_playback_change: nil}
+          )
```
mat-hek, line 101:
```suggestion
      if state.delayed_playback_change != nil do
        {:ok, state} = LifecycleController.change_playback_state(
          state.delayed_playback_change,
          %{state | delayed_playback_change: nil}
        )
```

mat-hek (approved):

lib/membrane/core/parent/child_life_controller/link_handler.ex
andpodob: Done.

mat-hek (approved):

PR #404, opened 2022-03-08
Raise exception when caps are not send before the first buffer

### LINKED ISSUE #341: Disallow sending buffers without sending caps first
Sending a buffer without caps is now possible even though many elements won't work without them (e.g. H264 encoder and decoder). Since there's no scenario, where not sending them is required, we should raise with a clear error when a buffer is sent and there are no caps set on the output pad

varsill:
Add ActionError exception raising when the `:buffer` action is made on pad, which does not have caps set. Add unit test and integration test checking that functionality. Change the existing tests in a`ction_handler_test.exs` and `auto_demands_test.exs` so that they are now working with pads that have the caps set before sending the first buffer.
closes #341


lib/membrane/core/element/action_handler.ex
```diff
@@ -269,12 +269,16 @@ defmodule Membrane.Core.Element.ActionHandler do
           eos: %{end_of_stream?: false} <- pad_data do
       %{pid: pid, other_ref: other_ref} = pad_data
 
-      state =
-        DemandHandler.handle_outgoing_buffers(pad_ref, pad_data, buffers, state)
-        |> PadModel.set_data!(pad_ref, :start_of_stream?, true)
-
-      Message.send(pid, :buffer, buffers, for_pad: other_ref)
-      {:ok, state}
+      if state.pads.data[pad_ref].caps == nil do
```
mat-hek, line 272: - this can be an additional clause in the `withl` above
- you can use the `pad_data` variable

lib/membrane/exceptions.ex
```diff
@@ -117,6 +117,10 @@ defmodule Membrane.ActionError do
     "Invalid event: #{inspect(event, pretty: true)}"
   end
 
+  defp format_reason(:caps_not_sent_before_first_buffer) do
+    "Caps were not sent on this pad before the first buffer was"
```
mat-hek, line 121:
```suggestion
    "Tried to send a buffer, while caps have not been sent on this pad"
```

test/membrane/core/element/action_handler_test.exs
```diff
@@ -155,7 +155,7 @@ defmodule Membrane.Core.Element.ActionHandlerTest do
     end
 
     test "when element is moving to playing", %{state: state} do
-      state = state |> set_playback_state(:prepared)
+      state = state |> set_playback_state(:prepared) |> PadModel.set_data!(:output, :caps, :any)
```
mat-hek, line 158:
```suggestion
      state = state |> set_playback_state(:prepared) |> PadModel.set_data!(:output, :caps, :some_caps)
```

mat-hek (changes requested):

lib/membrane/exceptions.ex
varsill: done

test/membrane/core/element/action_handler_test.exs
varsill: I have found that in this test file there is `@mock_caps` defined and I have used it there.

varsill: done

lib/membrane/core/element/action_handler.ex
varsill: done

mat-hek (approved):

PR #409, opened 2022-03-17
Fast map

mat-hek:
Using macros to speed up access to nested maps and structs. Doesn't compile on Elixir 1.13 to 1.13.3 due to https://github.com/elixir-lang/elixir/issues/11658

TODO:
- [ ] fix the weird 'incompatible types' warning happening in `PadController`
- [ ] separate `FastMap` to its own library
- [ ] add docs and tests for `FastMap` itself
- [ ] check the errors raised when passing invalid map keys
- [ ] verify structs' keys
- [ ] wait for the next Elixir release that will include the fix for https://github.com/elixir-lang/elixir/issues/11658

Performance of the new `FastMap` helper:

```elixir
Mix.install([:benchee, {:membrane_core, path: "."}])
defmodule Benchmark do
  alias Membrane.Core.Helper.FastMap

  require FastMap

  def run() do
    range = 1..10_000
    map = %{
      list: [],
      counter: 0,
      pad: %{
        :input => %{cnt: 1},
        {:input, 0} => %{cnt: 1}
      },
      and_some: :atom
    }
    not_nested = %{
      fast_map: fn ->
        Enum.reduce(range, map, fn i, map ->
          FastMap.update_in(map, [:counter], & &1 + i)
        end)
      end,
      map: fn ->
        Enum.reduce(range, map, fn i, map ->
          Map.update!(map, :counter, & &1 + i)
        end)
      end
    }
    Benchee.run(not_nested)
    nested = %{
      fast_map: fn ->
        Enum.reduce(range, map, fn i, map ->
          FastMap.update_in(map, [:pad, :input, :cnt], & &1 + i)
        end)
      end,
      update_in: fn ->
        Enum.reduce(range, map, fn _i, map ->
          update_in(map[:pad][:input][:cnt], & &1 + 1)
        end)
      end
    }
    Benchee.run(nested)
  end
end

Benchmark.run()
```

```
CPU Information: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.14.0-dev
Erlang 24.3

Name               ips        average  deviation         median         99th %
fast_map        3.37 K      297.03 μs    ±18.08%         284 μs         513 μs
map             1.47 K      681.08 μs    ±25.72%         644 μs     1119.77 μs

Comparison:
fast_map        3.37 K
map             1.47 K - 2.29x slower +384.06 μs

Name                ips        average  deviation         median         99th %
fast_map         1.32 K        0.76 ms   ±123.00%        0.66 ms        1.74 ms
update_in        0.43 K        2.34 ms    ±76.97%        2.17 ms        4.11 ms

Comparison:
fast_map         1.32 K
update_in        0.43 K - 3.10x slower +1.59 ms
```

Performance of the core measured with https://github.com/membraneframework/membrane_performance_test by @varsill 
```
Fast-map: PUSH: 57391.27 PULL: 41546.39 AUTODEMAND: 53491.85 [msg/s]
v0.9.0: PUSH: 46394.90 PULL: 27361.77 AUTODEMAND: 34696.72 [msg/s]
```


lib/membrane/core/helper/fast_map.ex
```diff
@@ -0,0 +1,150 @@
...
+
+        match =
+          quote do
+            %{^unquote(key) => unquote(var)} = unquote(map)
+          end
+
+        {{match, var}, {var, i + 1}}
+      end)
+
+    Enum.unzip(matches_and_vars)
```
bblaszkow06, line 137: Similar to the suggestion above:
```suggestion
    keys
    |> Enum.with_index()
    |> Enum.scan({nil, map_var}, fn {key, i}, {_match, upper_level_map} ->
        var = Macro.unique_var(:"nested_var#{i}", Elixir)

        match =
          quote do
            %{^unquote(key) => unquote(var)} = unquote(upper_level_map)
          end

        {match, var}
      end)
    |> Enum.unzip()
```

lib/membrane/core/helper/fast_map.ex
```diff
@@ -0,0 +1,150 @@
+defmodule Membrane.Core.Helper.FastMap do
+  @moduledoc false
+
+  defmacro get_in(map, keys) do
+    get_in_code(map, keys)
+  end
```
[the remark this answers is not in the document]
mat-hek: `*_code` functions are intended to be used by macros that would build on top of FastMap, such as the ones in PadModel. In that case, they're slightly more convenient to use than the macros:
```elixir
FastMap.get_in_code(state, [:pads_data, pad_ref] ++ keys)
```
instead of
```elixir
quote do
  require FastMap
  FastMap.get_in(unquote(state), unquote([:pads_data, pad_ref] ++ keys))
end
```

These functions also allow the generated code to be modified before compiling. That said, I'm not certainly sure that it's the right approach

test/membrane/core/element/action_handler_test.exs
```diff
@@ -2,37 +2,35 @@ defmodule Membrane.Core.Element.ActionHandlerTest do
...
   alias Membrane.{ActionError, Buffer}
-  alias Membrane.Core.{Message, Playback}
-  alias Membrane.Core.Child.PadModel
+  alias Membrane.Core.Playback
   alias Membrane.Core.Element.State
   alias Membrane.Support.DemandsTest.Filter
   alias Membrane.Support.Element.{TrivialFilter, TrivialSource}
 
-  require Message
+  require Membrane.Core.Child.PadModel, as: PadModel
```
bblaszkow06, line 10: Lol, didn't know that was possible

https://user-images.githubusercontent.com/2119556/159253181-6773e9f3-7678-4dec-9ac2-0c4cd3189724.mp4

lib/membrane/core/helper/fast_map.ex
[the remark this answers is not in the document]
bblaszkow06: OK, this approach seems reasonable, the former snippet seems more readable and concise without all the unquotes.
However, I'd modify the naming. `get_in_code`, especially from the outside of this module without the context of `get_in` function, seems like it was supposed to get something from the code.
Maybe sth like `get_in_codegen`?

We may also think of making it a convention to use common prefix/suffix for all the compile-time invoked functions, as I think there are more such cases around pad definitions

bblaszkow06 (changes requested):
Now the review is finished :wink:

lib/membrane/core/helper/fast_map.ex
```diff
@@ -0,0 +1,166 @@
+defmodule Membrane.Core.Helper.FastMap do
+  @moduledoc false
+
+  defmacro get_in(map, keys) do
```
bblaszkow06, line 4: Not sure if you intended to do this within this PR, but since you added comments on private functions, maybe you could add the docs here as well?

bblaszkow06 (approved):
Since it might take a while before the next Elixir release, you could do 2 things for now:

* Add `.tool-versions` locked with dev version of Elixir
* Include the installation of the Elixir version from that file in CI pipeline

The above could be accompanied with the required Elixir version bump in `mix.exs`

Other than that, LGTM!
seal_of_approval_1_250x250

mickel8:
First benchmarks for one room, 21 participants, some very low quality video using TestRTC

https://drive.google.com/file/d/1KV-RtUXM8LIpC8UuUndDR_YhJre3KaIn/view?usp=sharing

Can't upload html files :(

test/membrane/core/helper/fast_map_test.exs
```diff
@@ -0,0 +1,65 @@
+defmodule Membrane.Core.Helper.FastMapTest do
+  use ExUnit.Case, async: true
+
+  require Membrane.Core.Helper.FastMap, as: FastMap
+
+  doctest FastMap
```
bblaszkow06, line 6: Fun fact (not that I would recommend that here): https://twitter.com/TylerAYoung/status/1508413319178297352

bblaszkow06 (changes requested):

bblaszkow06 (approved):

PR #411, opened 2022-03-22
Add playback actions

### LINKED ISSUE #337: Add `playback: :stopped | :prepared | :playing` actions in pipeline and deprecate `Pipeline.{stop, prepare, play}`
`Pipeline.{stop, prepare, play}` functions enforce some managing process over a pipeline. According to our experience, though, pipelines can usually manage themselves without any problems. We even sometimes do `Pipeline.play(self())`. To adjust our API, we could replace `Pipeline.{stop, prepare, play}` with `:playback` action. In cases when the pipeline should be managed externally, one could use it as described in #336.

varsill:
Adds `{:playback, :playing | :prepared | :stopped}` actions which can be returned by the pipeline.
Depricates Membrane.Pipeline.{prepare, play, stop} functions and these functions defined by default in modules, which `use Membrane.Pipeline`
closes #337 
Updates the tests in core so that they do not use deprecated functions.


varsill:
EXPLANATION: 
I have decided to choose a "soft" manner of deprecation. Membrane.Pipeline.{play, prepare, stop} functions are still available, however, using them in code leads to a `@deprecated` warning being printed.

I have needed to change the `Membrane.Pipeline.__before_compile__` (invoked during the `use Membrane.Pipeline`) implementation so that the `play/1, stop/1, prepare/1` methods are no longer delegated to the `Membrane.Pipeline` default implementations. It was necessary since `Membrane.Pipeline.{play, stop, prepare}` are now `@deprecated` and we do not want a warning while defining a custom pipeline module with `use Membrane.Pipeline`. However, changing `defdelegate` to a function definition led to a [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) increase of the `__before_compile__` function. `mix credo` complains about this and I couldn't find any way to decrease that complexity, so I have just turned off the `mix credo` for the line defining `__before_compile`.

lib/membrane/core/pipeline/action_handler.ex
```diff
@@ -65,6 +65,11 @@ defmodule Membrane.Core.Pipeline.ActionHandler do
     TimerController.stop_timer(id, state)
   end
 
+  defp do_handle_action({:playback, playback_state}, _cb, _params, state) do
+    Membrane.Core.PlaybackHandler.request_playback_state_change(self(), playback_state)
```
mat-hek, line 69: This unnecessarily introduces sending a message, better call `LifecycleController.handle_playback_state`. Also check if there's a good error when passing an invalid state

lib/membrane/pipeline.ex
```diff
@@ -354,23 +358,29 @@ defmodule Membrane.Pipeline do
...
         """
-        @spec stop(pid) :: :ok
-        defdelegate stop(pid), to: unquote(__MODULE__)
+        @spec stop(pid()) :: :ok
+        @deprecated "use pipeline's :playback action instead"
+        def stop(pid),
+          do: Membrane.Core.PlaybackHandler.request_playback_state_change(pid, :stopped)
       end
 
       unless Enum.any?(1..2, &Module.defines?(__MODULE__, {:stop_and_terminate, &1})) do
```
mat-hek, line 386: I'd rename this to `terminate`

test/membrane/element_test.exs
```diff
@@ -75,44 +75,43 @@ defmodule Membrane.ElementTest do
 
   describe "Start of stream" do
     test "causes handle_start_of_stream/3 to be called", %{pipeline: pipeline} do
-      Testing.Pipeline.play(pipeline)
+      Testing.Pipeline.execute_actions(pipeline, playback: :playing)
```
mat-hek, line 78: I'd make all these pipelines return `playback: :playing` in `init` instead

mat-hek (changes requested):

lib/membrane/core/pipeline/action_handler.ex
varsill: hmm didn't you mean `LifecycleController. change_playback_state/2`? As far as I see, `LifecycleController. change_playback_state/2` calls `PlaybackHandler.change_playback_state/3`, which finally calls `LifecycleController.handle_playback_state/3` - but wraps some logic concerning updating `state.playback`, etc.

varsill: Is this the desired error message? :P 
image

lib/membrane/pipeline.ex
varsill: good idea, done

varsill: I have also changed `Membrane.Core.Bin.stop_and_terminate` into `Membrane.Core.Bin.terminate`

test/membrane/element_test.exs
varsill: done

lib/membrane/core/pipeline/action_handler.ex
mat-hek: > hmm didn't you mean LifecycleController. change_playback_state/2

yes, sorry

lib/membrane/pipeline.ex
```diff
@@ -265,17 +265,22 @@ defmodule Membrane.Pipeline do
...
           :ok | {:error, :timeout}
+  @deprecated "use terminate/2 instead"
   def stop_and_terminate(pipeline, opts \\ []) do
-    blocking? = Keyword.get(opts, :blocking?, false)
-    timeout = Keyword.get(opts, :timeout, 5000)
-
-    ref = if blocking?, do: Process.monitor(pipeline)
-
-    PlaybackHandler.request_playback_state_change(pipeline, :terminating)
+    do_terminate(pipeline, opts)
```
mat-hek, line 270: couldn't we just call `terminate` here?

varsill: well, sure we could :D Done

mat-hek (approved):

PR #414, opened 2022-04-05
Refine testing pipeline

varsill:
Closes https://membraneframework.atlassian.net/browse/MC-46


lib/membrane/parent_spec.ex
```diff
@@ -457,6 +464,25 @@ defmodule Membrane.ParentSpec do
...
+  @doc """
+  Links subsequent children using default pads (linking `:input` to `:output` of
+  previous element).
+
+  ## Example
+
+      Membrane.ParentSpec.populate_links([el1: MembraneElement1, el2: MembraneElement2])
+  """
+  @spec populate_links(children :: children_spec_t()) :: links_spec_t()
+  def populate_links(children) when length(children) < 2 do
```
mat-hek, line 476:
```suggestion
  def populate_links([]) do
```

lib/membrane/parent_spec.ex
```diff
@@ -457,6 +464,25 @@ defmodule Membrane.ParentSpec do
...
+    []
+  end
+
+  def populate_links(children) when is_list(children) do
+    [h | t] = children |> Keyword.keys()
+    links = t |> Enum.reduce(link(h), &to(&2, &1))
+    [links]
+  end
```
mat-hek, line 484: - I guess this should create the children as well, now it ignores the values
- `populate_links` isn't quite meaningful, maybe something like `link_linear`?

lib/membrane/testing/pipeline.ex
```diff
@@ -152,33 +147,93 @@ defmodule Membrane.Testing.Pipeline do
...
+
+      :custom ->
+        module = Keyword.fetch!(pipeline_options, :module)
+        custom_args = Keyword.get(pipeline_options, :custom_args)
+        test_process = Keyword.get(pipeline_options, :test_process)
+        %{mode: mode, module: module, custom_args: custom_args, test_process: test_process}
+
+      other_mode ->
+        raise "Unknown testing pipeline mode #{other_mode}. Available modes are: :custom and :default"
+    end
```
mat-hek, line 217: Actually, we could just have the `module` key that would be either `:default` or a custom module, and it could default to `:default`. WDYT?

lib/membrane/testing/pipeline.ex
```diff
@@ -152,33 +147,93 @@ defmodule Membrane.Testing.Pipeline do
           }
   end
 
-  @spec start_link(Options.t(), GenServer.options()) :: GenServer.on_start()
-  def start_link(pipeline_options, process_options \\ []) do
+  @type default_pipeline_keywords_list_t :: [
```
mat-hek, line 150:
```suggestion
  @type default_pipeline_keyword_list_t :: [
```

It's keyword list, not keywords list ;)

mat-hek (changes requested):

lib/membrane/testing/pipeline.ex
varsill: indeed, done

lib/membrane/testing/pipeline.ex
varsill: I am up to it

varsill: done

lib/membrane/parent_spec.ex
varsill: done

mat-hek (approved):

PR #417, opened 2022-04-13
Better handle element stream arguments

varsill:
closes #https://github.com/membraneframework/membrane_core/issues/219


mat-hek (approved):
Let's wait with merging as we discussed

mat-hek (approved):

PR #420, opened 2022-04-15
Synchronous calls to pipeline

varsill:
closes https://github.com/membraneframework/membrane_core/issues/334


lib/membrane/pipeline.ex
```diff
@@ -421,6 +437,14 @@ defmodule Membrane.Pipeline do
...
           do: Membrane.Pipeline.terminate(pipeline, opts)
       end
+
+      unless Module.defines?(__MODULE__, {:call, 3}) do
+        @doc """
+        Synchronously calls the pipeline
+        """
+        @spec call(pid, any, timeout) :: :ok
+        defdelegate call(pipeline, message, timeout), to: unquote(__MODULE__)
+      end
```
mat-hek, line 447: I wouldn't add that. Some pipelines may want to only use their API to interact with them instead of doing `call` directly.

lib/membrane/pipeline/action.ex
```diff
@@ -89,6 +89,19 @@ defmodule Membrane.Pipeline.Action do
   Changes the playback state of the pipeline to the chosen one.
   """
 
+  @type reply_t :: {:reply, message :: any}
+  @typedoc """
+  Action that replies to a `Pipeline.call/2`. Can be returned only from the `handle_call/3` callback, in
+  which context the caller reference is available, with the `:from` key.
```
mat-hek, line 95:
```suggestion
  Action that replies to a `Pipeline.call/2`. Can be returned only from the `handle_call/3` callback, in
  which context the caller reference is available, under the `:from` key.
```

Also, add the full reference to the `Pipeline.call` and `handle_call` so they were clickable in docs

lib/membrane/pipeline/action.ex
```diff
@@ -89,6 +89,19 @@ defmodule Membrane.Pipeline.Action do
...
+  """
+
+  @type reply_to_t() :: {:reply_to, {GenServer.from(), message :: any}}
+  @typedoc """
+  Action that replies to a `Pipeline.call/2`. Useful when one does not want to reply in
+  `handle_call/3` callback. A caller reference is required to be passed, so one needs to save this
+  reference from the `handle_call/3` context, where it is available with the `:from` key.
+  """
```
mat-hek, line 103: Something is messed up with typedocs in this module. Each typedoc should be before the corresponding type definition. Also, the typedoc in line 79 refers to the `t` type defined in line 105

mat-hek (changes requested):

lib/membrane/pipeline/action.ex
varsill: done

lib/membrane/pipeline.ex
varsill: done

lib/membrane/pipeline/action.ex
varsill: done

lib/membrane/core/pipeline.ex
```diff
@@ -119,6 +121,21 @@ defmodule Membrane.Core.Pipeline do
     :ok = state.module.handle_shutdown(reason, state.internal_state)
   end
 
+  @impl GenServer
+  def handle_call(message, from, state) do
+    context = Component.callback_context_generator([:Pipeline], Call, state, from: from)
```
mat-hek, line 126:
```suggestion
    context = &CallbackContext.Call.from_state(&1, from: from)
```

lib/membrane/pipeline/action.ex
```diff
@@ -77,18 +77,29 @@ defmodule Membrane.Pipeline.Action do
...
-  Changes the playback state of the pipeline to the chosen one.
+  Action that replies to a `Membrane.Pipeline.call/3`. Can be returned only from the `c:Membrane.Pipeline.handle_call/3` callback, in
+  which context the caller reference is available, under the `:from` key.
+  """
+  @type reply_t :: {:reply, message :: any}
+
+  @typedoc """
+  Action that replies to a `Membrane.Pipeline.call/3`. Useful when one does not want to reply in
+  `c:Membrane.Pipeline.handle_call/3` callback. A caller reference is required to be passed, so one needs to save this
+  reference from the `c:Membrane.Pipeline.handle_call/3` context, where it is available under the `:from` key.
```
mat-hek, line 93:
```suggestion
  reference from the `Membrane.Pipeline.CallbackContext.Call`, where it is available under the `:from` key.
```

mat-hek (dismissed):

mat-hek (approved):

mickel8:
Brilliant PR!

PR #426, opened 2022-05-04
Parent error cleanup

mat-hek:


lib/membrane/core/parent/child_life_controller/startup_handler.ex
```diff
@@ -83,13 +83,14 @@ defmodule Membrane.Core.Parent.ChildLifeController.StartupHandler do
...
-          {:ok | {:error, any}, Parent.state_t()}
+  @spec add_children([ChildEntry.t()], Parent.state_t()) :: Parent.state_t()
   def add_children(children, state) do
-    children
-    |> Bunch.Enum.try_reduce(state, fn child, state ->
-      state |> ChildrenModel.add_child(child.name, child)
-    end)
+    children =
+      Enum.reduce(children, state.children, fn child, children ->
+        Map.put(children, child.name, child)
```
bblaszkow06, line 90: This gets rid of checks for the existence of the pad inside children. (It was at line 12 in child_model) Was that intended?

lib/membrane/core/parent/children_model.ex
```diff
@@ -1,66 +1,43 @@
...
-    {pid, children} = children |> Map.pop(child)
-
-    with {:ok, pid} <- pid |> Bunch.error_if_nil({:unknown_child, child}) do
-      state = %{state | children: children}
-      {{:ok, pid}, state}
+  @spec get_child_data!(Parent.state_t(), Child.name_t()) :: ChildEntry.t()
+  def get_child_data!(%{children: children}, child) do
+    case children do
+      %{^child => data} -> data
+      _children -> raise ParentError, "Unknown child #{inspect(child)}"
```
bblaszkow06, line 19: I wonder if this message will be clear enough for the user to figure out what's wrong. Can you think of example where this might appear and how the stacktrace will look? Plus, WDYT about adding the list of children that are there?

```suggestion
      _children -> raise ParentError, "Unknown child #{inspect(child)}. Current children: #{inspect(Map.keys(children))}"
```

lib/membrane/core/parent/children_model.ex
```diff
@@ -1,66 +1,43 @@
...
-        else
-          {:error, {:unknown_child, name}}
+  @spec update_children!(Parent.state_t(), [Child.name_t()], (ChildEntry.t() -> ChildEntry.t())) ::
+          Parent.state_t()
+  def update_children!(%{children: children} = state, children_names, mapper) do
+    children =
+      Enum.reduce(children_names, children, fn name, children ->
+        case children do
+          %{^name => data} -> %{children | name => mapper.(data)}
+          _children -> raise ParentError, "Unknown child #{inspect(name)}"
```
bblaszkow06, line 30: Like above. Maybe `UnknownChildError` for deduplication?

lib/membrane/core/pipeline.ex
```diff
@@ -5,10 +5,11 @@ defmodule Membrane.Core.Pipeline do
   alias __MODULE__.{ActionHandler, State}
   alias Membrane.Clock
   alias Membrane.Core.CallbackHandler
-  alias Membrane.Core.Parent.MessageDispatcher
-  alias Membrane.Core.Telemetry
+  alias Membrane.Core.TimerController
+  alias Membrane.Core.Parent.{ChildLifeController, LifecycleController}
```
bblaszkow06, line 9: Lexicographically, `Parent` < `Timer`. Do we have some guidelines regarding alias sorting?

bblaszkow06 (changes requested):

lib/membrane/core/pipeline.ex
mat-hek: they should be lexicographically sorted, but there's nothing enforcing that ¯\_(ツ)_/¯

lib/membrane/core/parent/child_life_controller/startup_handler.ex
mat-hek: There was a check for duplicate children, but we already do that at line 37 in the child life controller

mat-hek: https://github.com/membraneframework/membrane_core/pull/426/files/a7cd2d7e29aaa3a427966d801a03d10849615f5f#diff-9fc8a1f82166b1f51259297d76ef5b53ffce74fc633eb00060190baf97edd403R37

lib/membrane/core/parent/message_dispatcher.ex
```diff
@@ -1,95 +0,0 @@
-defmodule Membrane.Core.Parent.MessageDispatcher do
-  @moduledoc false
```
varsill, line 2: I am asking out of my curiosity, as I cannot find a reason for that action on my own - what was the purpose of removing Parent.MessageDispatcher, when we are duplicating the code for handling messages in pipeline and bin right now? Because, I believe, there was some reason :D

lib/membrane/core/pipeline/action_handler.ex
```diff
@@ -4,63 +4,62 @@ defmodule Membrane.Core.Pipeline.ActionHandler do
...
-  defp do_handle_action({action, _args}, :handle_init, _params, state)
-       when action not in [:spec, :log_metadata, :playback] do
-    {{:error, {:unknown_action, Membrane.Pipeline.Action}}, state}
-  end
-
-  defp do_handle_action({:forward, children_messages}, _cb, _params, state) do
-    Parent.ChildLifeController.handle_forward(Bunch.listify(children_messages), state)
+  @impl CallbackHandler
+  def handle_action({:forward, children_messages}, _cb, _params, state) do
+    :ok = Parent.ChildLifeController.handle_forward(Bunch.listify(children_messages), state)
```
varsill, line 20: `Enum.each`, which is the last and the only thing called in `ChildLifeController.handle_forward`, always returns `:ok`, so there is probably no point in matching it with `:ok` on the left side.

PR #431, opened 2022-06-02
Set `trap_exit` to `true` before killing Element.

mickel8:
This should guarantee calling `c:handle_shutdown/2`.

The current `Process.flag(:trap_exit, false)` was introduced in https://github.com/membraneframework/membrane_core/pull/418. Before #418 we were returning `{:stop, ...}` tuple which cause calling `c:terminate/2` which then calls `c:handle_shutdown/2`.


mickel8:
I can see that these changes break some tests which make me think that `Process.flag(:trap_exit, false)` was introduced on purpose :thinking:

mickel8:
Maybe instead of `Process.exit` we should call `Kernel.exit`.

PR #433, opened 2022-06-17
Metadata in discontinuity event

varsill:
This is a draft PR which introduces an experimental new way to define `Membrane.Event`.


varsill:
We have come to the conclusion that we do not need that generic approach. For more precise reason of our decision, seek here: https://membraneframework.atlassian.net/wiki/spaces/MEMBRANEFR/pages/13074437/Events+definition+in+Membrane+Core

PR #435, opened 2022-07-08
Time conversion for rationals

varsill:
This PR introduces function to convert desired number of time units (`seconds`, `hours` etc.) represented as a `%Ratio` structure.


lib/membrane/time.ex
```diff
@@ -239,21 +239,33 @@ defmodule Membrane.Time do
...
     end
 
-    to_fun_name = :"to_#{unit.plural}"
+    # credo:disable-for-next-line Credo.Check.Readability.Specs
+    def unquote(unit.plural)(number) do
+      if not Ratio.is_rational?(number),
+        do:
+          raise(
+            "Only integers and rationals can be converted with Membrane.Time.#{unquote(unit.plural)}"
+          )
```
mat-hek, line 254: use do..end notation and remove brackets from raise

lib/membrane/time.ex
```diff
@@ -272,4 +284,27 @@ defmodule Membrane.Time do
     unit = @units |> Enum.find(&(rem(time, &1.duration) == 0))
     {time |> div(unit.duration), unit}
   end
+
+  defp round_rational(ratio) do
+    ratio = make_rational(ratio)
+    trunced = Kernel.div(ratio.numerator, ratio.denominator)
```
mat-hek, line 290: maybe use `Ratio.trunc`?

varsill: done

PR #438, opened 2022-07-12
Bump Ratio to 3.0

varsill:
This PR updates the Ratio dependency do 3.0 and updates the code accordingly to the changes in API. This PR should be merged once we decide to update Ratio dependency in all the plugins in organization.


varsill:
Closing since Ratio 4.0 has been released ;)

varsill:
Reopening since Ratio v4.0.0 is only a release candidate ;)

lib/membrane/clock.ex
```diff
@@ -212,9 +212,9 @@ defmodule Membrane.Clock do
   end
 
   defp handle_clock_update(till_next, state) do
-    use Ratio
+    use Numbers, overload_operators: true
 
-    if till_next < 0 do
+    if Ratio.lt?(till_next, 0) do
```
mat-hek, line 217: why do we overload operators if we don't use them?

lib/membrane/clock.ex
```diff
@@ -227,10 +227,10 @@ defmodule Membrane.Clock do
   end
 
   defp do_handle_clock_update(till_next, state) do
-    use Ratio
+    use Numbers, overload_operators: true
     %{till_next: from_previous, clock_time: clock_time} = state
     clock_time = clock_time + from_previous
-    ratio = clock_time / (state.time_provider.() - state.init_time)
+    ratio = Ratio.div(Ratio.new(clock_time), Ratio.new(state.time_provider.() - state.init_time))
```
mat-hek, line 233: as above

mat-hek (changes requested):

lib/membrane/clock.ex
varsill: Well, we are using them - the overloaded operators are not only comparison operators, but also arithmetic operators

lib/membrane/clock.ex
```diff
@@ -34,11 +34,6 @@ defmodule Membrane.Clock do
 
   @type t :: pid
 
-  @typedoc """
-  Ratio of the Clock time to the reference time.
-  """
-  @type ratio_t :: Ratio.t() | non_neg_integer
-
```
mat-hek, line 41: I'd leave this type

```suggestion
  @typedoc """
  Ratio of the Clock time to the reference time.
  """
  @type ratio_t :: Ratio.t()
```

lib/membrane/clock.ex
```diff
@@ -227,10 +222,9 @@ defmodule Membrane.Clock do
   end
 
   defp do_handle_clock_update(till_next, state) do
-    use Ratio
     %{till_next: from_previous, clock_time: clock_time} = state
-    clock_time = clock_time + from_previous
-    ratio = clock_time / (state.time_provider.() - state.init_time)
+    clock_time = Ratio.add(clock_time, from_previous)
+    ratio = Ratio.new(clock_time, state.time_provider.() - state.init_time)
```
mat-hek, line 227: `div` is less confusing in this case IMO

```suggestion
    ratio = Ratio.div(clock_time, state.time_provider.() - state.init_time)
```

lib/membrane/core/timer.ex
```diff
@@ -48,7 +48,7 @@ defmodule Membrane.Core.Timer do
   end
 
   def tick(timer) do
-    use Ratio
+    use Numbers, overload_operators: true
```
mat-hek, line 51: let's be consistent and not use that

lib/membrane/time.ex
```diff
@@ -260,12 +262,10 @@ defmodule Membrane.Time do
     end
 
     # credo:disable-for-next-line Credo.Check.Readability.Specs
-    def unquote(unit.plural)(number) do
-      if not Ratio.is_rational?(number) do
-        raise "Only integers and rationals can be converted with Membrane.Time.#{unquote(unit.plural)}"
-      end
+    def unquote(unit.plural)(number) when Ratio.is_rational(number) do
+      use Numbers, overload_operators: true
```
mat-hek, line 266: as above

lib/membrane/clock.ex
varsill: done

lib/membrane/clock.ex
varsill: done

lib/membrane/time.ex
varsill: done here and also in clock_test.exs

mat-hek (approved):

PR #439, opened 2022-07-13
Specify Membrane.Time.non_neg_t() as spec for timers

bblaszkow06:


varsill (commented):
The `Membrane.Time.t()` is defined as follows:
```
  @type t :: integer
```
So we cannot substitute the `non_neg_integer` with that. I believe it wouldn't make sense to have, i.e. a timer with a negative interval.

However, we have the following type in `Membrane.Time`:
```
  @type non_neg_t :: non_neg_integer
```

varsill (approved):
lgtm 🥇

mat-hek (approved):

PR #441, opened 2022-07-21
Add Membrane.Time.to_timebase/2 function

varsill:


lib/membrane/time.ex
```diff
@@ -226,6 +226,14 @@ defmodule Membrane.Time do
     round(value / native_unit())
   end
 
+  @doc """
+  Returns timestamp in timebase units. Rounded to the nearest integer.
+  """
+  @spec to_timebase(number | Ratio.t(), number | Ratio.t()) :: integer
+  def to_timebase(timestamp, timebase) do
```
mat-hek, line 233:
```suggestion
  @spec round_to_timebase(t, t | Ratio.t()) :: integer
  def round_to_timebase(timestamp, timebase) do
```

lib/membrane/time.ex
```diff
@@ -226,6 +226,14 @@ defmodule Membrane.Time do
     round(value / native_unit())
   end
 
+  @doc """
+  Returns timestamp in timebase units. Rounded to the nearest integer.
```
mat-hek, line 230: an example would be useful

mat-hek (changes requested):

lib/membrane/time.ex
varsill: done

lib/membrane/time.ex
varsill: done

lib/membrane/time.ex
```diff
@@ -228,9 +228,15 @@ defmodule Membrane.Time do
 
   @doc """
   Returns timestamp in timebase units. Rounded to the nearest integer.
+
+  ## Examples:
+      iex> timestamp = 10 |> Membrane.Time.seconds()
+      iex> timebase = Ratio.new(Membrane.Time.second(), 30)
+      iex> Membrane.Time.round_to_timebase(timestamp, timebase)
+      300
```
mat-hek, line 236: you can make this a doctest

mat-hek (approved):

PR #447, opened 2022-09-06
Improved ChildrenSupervisor, ResourceGuard and Observability

mat-hek:
- [x] - improve the ChildrenSupervisor to be able to spawn utility processes as well
- [x] - add ResourceGuard
- [x] - tests improvements


lib/membrane/core/parent/child_life_controller/startup_utils.ex
```diff
@@ -142,10 +142,15 @@ defmodule Membrane.Core.Parent.ChildLifeController.StartupUtils do
...
           end
 
-          fn -> Core.Bin.start_link(params) end
+          Core.Bin
       end
 
-    with {:ok, child_pid} <- ChildrenSupervisor.start_child(supervisor, name, start_fun),
+    start_fun = fn supervisor ->
+      server_module.start_link(Map.put(params, :children_supervisor, supervisor))
+    end
```
bblaszkow06, line 150: OK finally found the place where this is defined. IMHO instead of lambda I'd pass the module and args

lib/membrane/utility_supervisor.ex
```diff
@@ -0,0 +1,30 @@
...
+  @moduledoc """
+  A supervisor that allows to start utility processes under the pipeline's
+  supervision tree.
+
+  The supervisor is spawned with each component and can be obtained from
+  callback contexts.
+
+  The supervisor never restarts any processes, it just makes sure they
+  terminate when the component that started them terminates. If restarting
+  is needed, a dedicated supervisor should be spawned under this supervisor.
```
bblaszkow06, line 11: I guess an example here wouldn't hurt

lib/membrane/resource_guard.ex
```diff
@@ -0,0 +1,106 @@
+defmodule Membrane.ResourceGuard do
+  @moduledoc """
+  Utility for handling resources that must be cleaned up after using.
+
+  This utility uses a separate process that allows to register functions
+  that are called when owner process (passed to `start_link/1`) dies for
+  any reason. Each Membrane component spawns its resource guard on startup
+  and provides it via callback context.
```
bblaszkow06, line 8: Example?

lib/membrane/resource_guard.ex
```diff
@@ -0,0 +1,106 @@
...
+    GenServer.start(__MODULE__, owner_pid, spawn_opt: [:link])
+  end
+
+  @doc """
+  Registers a resource cleanup function in the resource guard.
+
+  Registered functions are called in the order reverse to the registration order.
+  Return value of a registered function is ignored. If a `name` is passed,
+  the function can be cleaned up manually with `cleanup_resource/2`. Many
+  functions can be registered with the same name.
```
bblaszkow06, line 28:
```suggestion
  Registers a resource cleanup function in the resource guard.

  Registered functions are called in the order reverse to the registration order.
  A return value of a registered function is ignored. If a `name` is passed,
  the function can be cleaned up manually with `cleanup_resource/2`. Many
  functions can be registered with the same name.
```

lib/membrane/resource_guard.ex
bblaszkow06, line 8:
```suggestion
  Utility for handling resources that must be cleaned up after use.

  This utility uses a separate process that allows registering functions
  that are called when the owner process (passed to `start_link/1`) dies for
  any reason. Each Membrane component spawns its resource guard on startup
  and provides it via callback context.
```

bblaszkow06 (changes requested):

lib/membrane/resource_guard.ex
```diff
@@ -0,0 +1,106 @@
...
+    {:ok, %{guards: [], monitor: monitor}}
+  end
+
+  @impl true
+  def handle_info(Message.new(:register_resource, [function, name]), state) do
+    {:noreply, %{state | guards: [{function, name} | state.guards]}}
+  end
+
+  @impl true
+  def handle_info(Message.new(:cleanup_resource, name), state) do
```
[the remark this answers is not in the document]
mat-hek: Using bare processes has some problems - for example, they cannot be introspected with the observer. We don't use casts anywhere across the core. I don't see any benefit in using them, but if we choose to do so, we should be consistent.

lib/membrane/core/children_supervisor.ex
```diff
@@ -0,0 +1,198 @@
...
+        ) ::
+          Supervisor.on_start_child()
+  def start_link_utility(supervisor, child_spec) do
+    result = start_utility(supervisor, child_spec)
+
+    case result do
+      {:ok, pid, _info} -> Process.link(pid)
+      {:ok, pid} -> Process.link(pid)
+      _error -> :ok
+    end
```
varsill, line 54: [Not that important] `Process.link/1` returns true - just for the purpose of consistency the last match could be `_error -> true`. Btw. - are we anyhow informing about the error in `start_utility`? (Or perhaps there is just no reason to do so?)

lib/membrane/pipeline.ex
```diff
@@ -278,7 +278,7 @@ defmodule Membrane.Pipeline do
 
       setup_observability = Membrane.Core.Observability.setup_fun(:pipeline, name)
 
-      Membrane.Core.Parent.Supervisor.go_brrr(
+      Membrane.Core.Pipeline.Supervisor.go_brrr(
```
varsill, line 281: Are we sticking to `go_brrr`? :D

lib/membrane/utility_supervisor.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Membrane.UtilitySupervisor do
```
varsill, line 1: I don't think we have a test for utility supervisor

lib/membrane/core/children_supervisor.ex
mat-hek: I think `true` would look weird here. Not sure what you mean by informing about the error, but the result is returned  in the next line, no matter if error or not

lib/membrane/pipeline.ex
mat-hek: Why not? :P

lib/membrane/core/parent/child_life_controller/startup_utils.ex
mat-hek: done

lib/membrane/pipeline.ex
mat-hek: done

lib/membrane/resource_guard.ex
mat-hek: done

lib/membrane/resource_guard.ex
```diff
@@ -83,21 +96,24 @@ defmodule Membrane.ResourceGuard do
...
-      function.()
-      :ok
-    rescue
-      error -> {:error, error}
-    catch
-      error -> {:error, error}
+    {:ok, task} = Task.start_link(function)
+
+    receive do
+      {:EXIT, ^task, reason} -> reason
```
bblaszkow06, line 102: You forgot about the timeout.

bblaszkow06 (changes requested):

lib/membrane/pipeline.ex
mat-hek: image

lib/membrane/core/observability.ex
```diff
@@ -1,44 +1,45 @@
...
-  @spec setup_fun(:element | :bin | :pipeline, name :: term, Logger.metadata()) :: setup_fun()
-  def setup_fun(component_type, name, log_metadata \\ []) do
-    component_path = Membrane.ComponentPath.get()
+  @type config :: %{
+          optional(:parent_path) => ComponentPath.path_t(),
+          optional(:log_metadata) => Logger.metadata(),
+          name: term,
+          component_type: :element | :bin | :pipeline,
+          pid: pid()
+        }
```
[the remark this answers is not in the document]
mat-hek: I think it's overkill

lib/membrane/resource_guard.ex
mat-hek: done

lib/membrane/resource_guard.ex
```diff
@@ -95,11 +101,16 @@ defmodule Membrane.ResourceGuard do
...
-  defp cleanup(function, name) do
+  defp cleanup(function, name, timeout) do
     {:ok, task} = Task.start_link(function)
 
     receive do
       {:EXIT, ^task, reason} -> reason
+    after
+      timeout ->
+        Membrane.Logger.error("Cleanup of resource #{inspect(name)} timed out, killing")
+        Process.exit(task, :kill)
```
bblaszkow06, line 112: This will put another `:EXIT` signal in your mailbox (i.e. `{:EXIT, task_pid, :killed}`). I'm not sure if that's not going a bit into the implementation details of `Task`s but you can unlink the task process before killing like `Task.ignore` does (https://github.com/elixir-lang/elixir/blob/c5151e6890b5ac8df13276459696f0f47a8e634b/lib/elixir/lib/task.ex#L850)

bblaszkow06 (changes requested):

varsill (approved):

bblaszkow06 (approved):

PR #450, opened 2022-09-29
Prevent crash when stopping timer without interval

bblaszkow06:


lib/membrane/core/timer.ex
```diff
@@ -28,8 +28,12 @@ defmodule Membrane.Core.Timer do
   end
 
   @spec stop(t) :: :ok
-  def stop(timer) do
-    Process.cancel_timer(timer.timer_ref)
+  def stop(%__MODULE__{timer_ref: nil}) do
```
mat-hek, line 31:
```suggestion
  def stop(%__MODULE__{interval: :no_interval}) do
```

I think that will be more meaningful

mat-hek (approved):

PR #453, opened 2022-10-06
Get rid of espec

varsill:
This PR removes the ESpec dependency, removes unecessary ESpec tests and rewrites the last two significant ESpec tests (input queue test and caps matcher test) with ExUnit.


test/membrane/core/element/input_queue_test.exs
```diff
@@ -1,9 +1,187 @@
...
+             }
+
+      message =
+        Message.new(:demand, context.target_queue_size, for_pad: context.linked_output_ref)
+
+      assert_received ^message
+    end
+
+    test "send the demand if toilet is enabled not", context do
+      flush()
```
mat-hek, line 53: Each exunit test is run in a separate process, so I think it's not needed ;)

mix.exs
```diff
@@ -16,9 +16,8 @@ defmodule Membrane.Mixfile do
       name: "Membrane Core",
       source_url: link(),
       docs: docs(),
-      aliases: ["test.all": "do espec, test", docs: ["docs", &copy_assets/1]],
```
bblaszkow06, line 19: I'd remove "test.all" altogether.

mat-hek (approved):

mat-hek:
Let's merge this after all the playback stuff is merged

mix.exs
```diff
@@ -16,16 +16,15 @@ defmodule Membrane.Mixfile do
...
+      aliases: [docs: ["docs", &copy_assets/1]],
       preferred_cli_env: [
-        espec: :test,
-        "test.all": :test,
+        test: :test,
```
bblaszkow06, line 21: This isn't needed, `test` task already specifies test as its preferred env (this option allows to override it).
```suggestion
```

bblaszkow06 (approved):

mix.exs
varsill: done

PR #454, opened 2022-10-11
Match caps in bin's pads

FelonEkonom:
Related Jira tickets: 
 * https://membraneframework.atlassian.net/browse/MC-53
 * https://membraneframework.atlassian.net/browse/MC-76


lib/membrane/core/bin/pad_controller.ex
```diff
@@ -160,7 +160,12 @@ defmodule Membrane.Core.Bin.PadController do
           StructureParser.raw_endpoint_t(),
           StructureParser.raw_endpoint_t(),
           %{initiator: :parent}
-          | %{initiator: :sibling, other_info: PadModel.pad_info_t() | nil, link_metadata: map},
+          | %{
+              initiator: :sibling,
+              other_info: PadModel.pad_info_t() | nil,
+              link_metadata: map,
+              parents_accepted_caps: [struct() | module() | tuple()]
```
varsill, line 167: Are you sure these specification is correct? What if the bin's caps specification is, for instance, `:any`?

lib/membrane/core/element/caps_controller.ex
```diff
@@ -50,10 +50,21 @@ defmodule Membrane.Core.Element.CapsController do
...
+
     context = &CallbackContext.Caps.from_state(&1, pad: pad_ref)
 
-    if Caps.Matcher.match?(accepted_caps, caps) do
+    [accepted_caps | parents_accepted_caps]
+    |> Enum.any?(&(not Caps.Matcher.match?(&1, caps)))
+    |> if do
+      raise """
+      Received caps: #{inspect(caps)} that are not specified in def_input_pad
+      for pad #{inspect(pad_ref)}. Specs of accepted caps are:
```
varsill, line 64: How about adding information about the element name / module for which the caps matching failed? So that the user wouldn't need to traverse the nested bins one by one to find the problematic specification.

lib/membrane/element/pad_data.ex
```diff
@@ -3,7 +3,9 @@ defmodule Membrane.Element.PadData do
   Struct describing current pad state.
 
   The public fields are:
-    - `:accepted_caps` - specification of possible caps that are accepted on the pad.
+    - `:accepted_caps` - specification of possible caps that are accepted on the pad in the element.
+      See `Membrane.Caps.Matcher` for more information.
+    - `:parents_accepted_caps` - specification of possible caps that are accepted on the pad in the ancestors of the element.
```
varsill, line 8: I wouldn't refer to "ancestors" of the element - why not to simply call them parents?

lib/membrane/core/bin/pad_controller.ex
varsill: I see that the `parents_accepted_caps` is aggregated out of `state.pads_info[direction].accepted_caps`, which in fact, are specified as `any` :O https://github.com/membraneframework/membrane_core/blob/4ad63c58a6d733a2e808f4a17f855ab262e52bf5/lib/membrane/core/child/pad_model.ex#L56

varsill (commented):
Good job!
However, apart from some small changes I have suggested, I believe we need a test that checks if it also works for nested bins

lib/membrane/element/pad_data.ex
FelonEkonom: In this case, by "ancestors" I mean not only parents, but also "grandparents" and their ancestors as well. Naming them `parents` wouldn't suggest only closest parents?

lib/membrane/core/child/pads_specs.ex
```diff
@@ -82,6 +50,15 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
                      )
       unquote(pad_opts_typedef)
+
+      if not Module.get_attribute(__MODULE__, :has_caps_match_spec, false) do
+        Module.put_attribute(__MODULE__, :has_caps_match_spec, true)
+        @spec caps_match?(any(), any()) :: boolean()
+      end
+
+      def caps_match?(unquote(pad_name), caps) do
+        match?(unquote(caps_pattern), caps)
```
mat-hek, line 60: This doesn't seem to support a list of patterns. If someone wants to match two different formats, it's going to be tough to do without that possibility. We may also consider making use of elixir's `or` operator instead of a list, WDYT @varsill ?

Also, passing just a format module should work. It can be done by converting it to a match on a struct of that module

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -159,16 +159,21 @@ defmodule Membrane.Core.Bin.PadController do
           Pad.direction_t(),
           StructureParser.raw_endpoint_t(),
           StructureParser.raw_endpoint_t(),
-          %{initiator: :parent}
-          | %{initiator: :sibling, other_info: PadModel.pad_info_t() | nil, link_metadata: map},
+          %{initiator: :parent, parents_with_pads: [{module(), Pad.name_t()}]}
```
mat-hek, line 162:
```suggestion
          %{initiator: :parent, ancestors_pads: [{module(), Pad.name_t()}]}
```

lib/membrane/core/element/caps_controller.ex
```diff
@@ -50,26 +50,36 @@ defmodule Membrane.Core.Element.CapsController do
...
+        ActionHandler,
+        %{context: context} |> Map.merge(params),
+        [pad_ref, caps],
+        state
+      )
+
+    PadModel.set_data!(state, pad_ref, :caps, caps)
+  end
+
+  @spec ensure_caps_match(Pad.direction_t(), [{module(), Pad.name_t()}], Caps.t()) :: :ok
```
mat-hek, line 72: I'd extract that `[{module(), Pad.name_t()}]` to a new type in this module

lib/membrane/core/element/caps_controller.ex
```diff
@@ -50,26 +50,36 @@ defmodule Membrane.Core.Element.CapsController do
...
+  def ensure_caps_match(direction, modules_with_pads, caps) do
```
mat-hek, line 73:
```suggestion
  def ensure_caps_match!(direction, modules_with_pads, caps) do
```

lib/membrane/core/element/caps_controller.ex
```diff
@@ -50,26 +50,36 @@ defmodule Membrane.Core.Element.CapsController do
...
+    for {module, pad_name} <- modules_with_pads do
+      if not module.caps_match?(pad_name, caps) do
```
mat-hek, line 75: It's a unique occasion to use unless :P

```suggestion
      unless module.caps_match?(pad_name, caps) do
```

test/membrane/core/element/action_handler_test.exs
```diff
@@ -363,39 +361,39 @@ defmodule Membrane.Core.Element.ActionHandlerTest do
...
+    #     @module.handle_action(
+    #       caps_action(:output),
+    #       :handle_info,
+    #       %{},
+    #       state
+    #     )
+    #   end
+
+    #   refute_received Message.new(:caps, @mock_caps, for_pad: :other_ref)
+    # end
```
mat-hek, line 396: fix or remove

mat-hek (changes requested):

test/support/caps_test/inner_source_bin.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Membrane.Support.CapsTest.InnerSourceBin do
+  @moduledoc """
+  Bin used in caps test.
+  Has caps pattern for `:output` pad.
```
varsill, line 4:
```suggestion
It has a caps pattern defined for the `:output` pad.
```

test/support/caps_test/outer_sink_bin.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Membrane.Support.CapsTest.OuterSinkBin do
+  @moduledoc """
+  Bin used in caps test.
+  Has caps pattern for `:input` pad.
```
varsill, line 4:
```suggestion
It has a caps pattern defined for the `:input` pad.
```

test/support/caps_test/outer_source_bin.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Membrane.Support.CapsTest.OuterSourceBin do
+  @moduledoc """
+  Bin used in caps test.
+  Has caps pattern for `:output` pad.
```
varsill, line 4: It has a caps pattern defined for the `:output` pad.

test/membrane/integration/caps_test.exs
```diff
@@ -0,0 +1,105 @@
...
+      pipeline = pipeline_with_source_bins()
+      send_caps(pipeline, FormatAcceptedByOuterBins)
+      assert_source_down()
+    end
+
+    test "output caps patterns in outer bins" do
+      pipeline = pipeline_with_source_bins()
+      send_caps(pipeline, FormatAcceptedByInnerBins)
+      assert_source_down()
+    end
```
varsill, line 63: You could add here also a test to make sure there is a crash when the format is not accepted by the elements themselves.

test/membrane/integration/caps_test.exs
```diff
@@ -0,0 +1,105 @@
...
+    assert_receive(
+      {:DOWN, _ref, :process, ^source_pid, {%Membrane.CapsMatchError{}, _stacktrace}}
+    )
+  end
+
+  defp assert_sink_down() do
+    assert_receive({:sink_pid, sink_pid})
+    Process.monitor(sink_pid)
+    assert_receive({:DOWN, _ref, :process, ^sink_pid, {%Membrane.CapsMatchError{}, _stacktrace}})
+  end
```
varsill, line 104: How about parametrizing it so that to have `assert_down(name)` function?

lib/membrane/core/child/pads_specs.ex
FelonEkonom: IMO using `or` or `|` operator to have a collection of checked patterns is ok, but using a list could be confusing because it might be mistaken with the match on the whole list. Beyond this, are you sure, we need to support passing just a format module? It could lead to potential ambiguity when a value of a field also should be a struct, and writing a match on a struct is not much harder than passing only a module name

lib/membrane/core/element/action_handler.ex
```diff
@@ -294,21 +294,17 @@ defmodule Membrane.Core.Element.ActionHandler do
 
     pad_data = PadModel.get_data!(state, pad_ref)
 
-    withl data: %{direction: :output, pid: pid, other_ref: other_ref} <- pad_data,
-          caps: true <- Caps.Matcher.match?(pad_data.accepted_caps, caps) do
+    with %{direction: :output, pid: pid, other_ref: other_ref} <- pad_data do
+      %{name: pad_name, parents_with_pads: parents_with_pads} = pad_data
```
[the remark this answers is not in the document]
varsill: Well, I thought only of fetching: `:name`,  `:parents_with_pads` all along with `:direction`, `:pid` and `other_ref`

lib/membrane/core/child/pads_specs.ex
mat-hek: `or` is not hard to handle:
```elixir
def or_to_list({:or, _meta, [a, b]}), do: or_to_list(a) ++ or_to_list(b)
def or_to_list(term), do: [term]
```
I'd rather worry about the fact that it normally accepts booleans, not patterns

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -159,16 +159,21 @@ defmodule Membrane.Core.Bin.PadController do
...
+          | %{
+              initiator: :sibling,
+              other_info: PadModel.pad_info_t() | nil,
+              link_metadata: map,
+              parents_with_pads: [{module(), Pad.name_t()}]
```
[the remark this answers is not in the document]
FelonEkonom: @mat-hek what is your impression? IMO `parents_with_pads` is ok, because it suggests, that we deal with parents and **their** pads (not parents and **some** pads)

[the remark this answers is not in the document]
mat-hek: I'd stick to `ancestors_pads` :P

lib/membrane/core/child/pads_specs.ex
FelonEkonom: Because of operators precedence, it is not a good idea to use `|` or `or`, because in both cases expression like `a | b when b == 2 | c` is equal to `(a | b) when (b == 2 | c)`. Another option, beyond list, is to use our custom function call (something like `alternative(a, b when b == 2, c)`) and convert it to list during parsing AST

test/support/caps_test/source.ex
```diff
@@ -0,0 +1,42 @@
...
+    {:ok, %{test_pid: test_pid}}
+  end
+
+  @impl true
+  def handle_playing(_ctx, %{caps: caps} = state) do
+    {{:ok, caps: {:output, caps}}, state}
+  end
+
+  @impl true
+  def handle_playing(_ctx, state) do
```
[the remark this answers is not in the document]
FelonEkonom: Unfortunately, it sometimes crashes without this match

lib/membrane/core/bin/pad_controller.ex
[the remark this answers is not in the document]
FelonEkonom: Yeah, but we hold there not only pads, but related modules as well, so maybe `ancestors_with_pads`?

lib/membrane/core/child/pads_specs.ex
mat-hek: That's a valid point, I'd stick to the list then. That's what the voice of people said too, as well as that passing a bare module shall be supported :P

lib/membrane/core/bin/pad_controller.ex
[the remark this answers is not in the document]
mat-hek: Doesn't matter much to me

test/support/caps_test/source.ex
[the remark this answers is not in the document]
mat-hek: can't you just pass the caps via options?

mat-hek (changes requested):

lib/membrane/bin/pad_data.ex
```diff
@@ -21,6 +21,7 @@ defmodule Membrane.Bin.PadData do
           ref: Membrane.Pad.ref_t(),
           options: Membrane.ChildrenSpec.pad_options_t(),
           availability: Membrane.Pad.availability_t(),
+          caps_pattern: String.t(),
```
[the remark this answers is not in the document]
FelonEkonom: It is used to generate docs

lib/membrane/core/child/pads_specs.ex
```diff
@@ -56,21 +34,34 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
-    specs =
-      specs
-      |> Keyword.put(:options, escaped_pad_opts)
+    specs = Keyword.put(specs, :options, escaped_pad_opts)
+    {caps_pattern, specs} = Keyword.pop!(specs, :caps)
+    specs = [{:caps_pattern, Macro.to_string(caps_pattern)} | specs]
+
+    case_statement_clauses =
+      case caps_pattern do
+        {:alternative, _meta, args} -> args
```
mat-hek, line 46: I think it would introduce more confusion than profit in the end, so I'd remove that, at least for now

lib/membrane/pad.ex
```diff
@@ -139,13 +138,18 @@ defmodule Membrane.Pad do
...
           optional(:demand_unit) => Buffer.Metric.unit_t(),
           :direction => direction_t(),
           :options => nil | Keyword.t(),
           optional(:demand_mode) => demand_mode_t()
         }
 
+  @typedoc """
+  Type describing module with a name of a pad defined in it.
+  """
+  @type module_with_pad_t() :: {module(), name_t()}
```
mat-hek, line 151: I don't think we need this type. At least it shouldn't be public

lib/membrane/pad.ex
```diff
@@ -129,7 +129,6 @@ defmodule Membrane.Pad do
   @type common_spec_options_t ::
           {:availability, availability_t()}
           | {:mode, mode_t()}
-          | {:caps, Caps.Matcher.caps_specs_t()}
```
mat-hek, line 132: This shouldn't disappear and the docs should say how the caps specification should look like

mat-hek (changes requested):

lib/membrane/pad.ex
FelonEkonom: https://github.com/membraneframework/membrane_core/pull/454#discussion_r998057448

FelonEkonom: Do you have an idea for a better name for this type?

lib/membrane/pad.ex
```diff
@@ -128,6 +128,7 @@ defmodule Membrane.Pad do
   """
   @type common_spec_options_t ::
           {:availability, availability_t()}
+          | {:caps_pattern, code :: any()}
```
mat-hek, line 131: This still doesn't tell how the caps specification should actually look like. Let's create `caps_pattern_t` and describe that in its typedoc

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -159,12 +160,12 @@ defmodule Membrane.Core.Bin.PadController do
           Pad.direction_t(),
           StructureParser.raw_endpoint_t(),
           StructureParser.raw_endpoint_t(),
-          %{initiator: :parent, ancestors_with_pads: [Pad.module_with_pad_t()]}
+          %{initiator: :parent, ancestors_with_pads: CapsController.caps_validation_params_t()}
```
mat-hek, line 163: So ancestors_with_pads can now become caps_validation_params ;)

lib/membrane/element/pad_data.ex
```diff
@@ -27,16 +25,17 @@ defmodule Membrane.Element.PadData do
   @type private_field :: term()
 
   @type t :: %__MODULE__{
-          accepted_caps: Caps.Matcher.caps_specs_t(),
           availability: Pad.availability_t(),
           caps: Caps.t() | nil,
+          caps_pattern: String.t(),
```
varsill, line 30: `caps_pattern` should be listed as one of the public fields avaibale in the `PadData` structure, along with the description of that field

lib/membrane/pad.ex
```diff
@@ -95,6 +95,13 @@ defmodule Membrane.Pad do
   """
   @type availability_mode_t :: :static | :dynamic
 
+  @typedoc """
+  Describes pattern, that should be matched by caps send by element on specific pad.
+  Will not be evaluated during runtime, but used for matching struct passed in `:caps`
+  action.
+  """
+  @type caps_t :: any()
```
varsill, line 103: I am not sure if typespec `any()` should be used to describe a pattern - but I've no idea how it should be typespeced either :D

varsill (commented):
Some small things to reconsider

lib/membrane/element/pad_data.ex
mat-hek: Not sure if we should make it public. But if so, this should be rather kept as AST

lib/membrane/pad.ex
mat-hek: There's probably no better way, unfortunately. But I suggested a change that may improve it a bi

FelonEkonom:
Changes from here were moved to https://github.com/membraneframework/membrane_core/pull/463

PR #455, opened 2022-10-12
Fix filter aggregator

mat-hek:


test/membrane/filter_aggregator/unit_test.exs
```diff
@@ -1,384 +1,392 @@
...
+      assert map_size(pads) == 2
+
+      {{:ok, caps: {:output, %MockCaps{integer: 2}}}, %{state | state: :caps_sent}}
+    end)
+
+    expect(FilterB, :handle_playing, fn ctx_b, state ->
+      # ensure proper callbacks order
+      assert state == %{module: FilterB, state: :caps_sent}
+
+      assert %{
```
bblaszkow06, line 218: You have ommited the struct's module name

lib/membrane/bin/callback_context/init.ex
```diff
@@ -0,0 +1,6 @@
+defmodule Membrane.Bin.CallbackContext.Init do
+  @moduledoc """
+  Callback context for `c:Membrane.Bin.handle_init/2`.
+  """
+  use Membrane.Core.Bin.CallbackContext
+end
```
bblaszkow06, line 6: What's the reason for adding that? From the point of callback user it seems useless

mat-hek: It's needed if you want to register resources or spawn utility processes

bblaszkow06: But shouldn't that be done in `handle_setup` rather than init?

mat-hek: It depends. If we allow spawning children in `handle_init`, we should also allow spawning utils etc. Context in handle_init makes it more future-proof as well

varsill (approved):

lib/membrane/bin.ex
```diff
@@ -279,7 +280,11 @@ defmodule Membrane.Bin do
       def membrane_bin?, do: true
 
       @impl true
-      def handle_init(_options), do: {:ok, %{}}
+      def handle_init(_ctx, %opt_struct{} = options),
```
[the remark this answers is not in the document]
mat-hek: fixed

PR #456, opened 2022-10-14
Fix spawning children on remote nodes

mat-hek:


lib/membrane/core/bin.ex
```diff
@@ -52,31 +52,29 @@ defmodule Membrane.Core.Bin do
...
   """
+
   @spec start(options_t()) :: GenServer.on_start()
   def start(options),
     do: do_start(:start, options)
 
   defp do_start(method, options) do
     %{module: module, name: name, node: node, user_options: user_options} = options
 
-    if Membrane.Bin.bin?(module) do
```
varsill, line 62: Why that check was put there :o

varsill (approved):

lib/membrane/core/bin.ex
mat-hek: To make sure that the user passed the proper module :P but it's already checked in startup_utils

PR #467, opened 2022-10-28
only configure junit in test env

mat-hek:
Without this, the following warning is printed:
```
You have configured application :junit_formatter in your configuration file,
but the application is not available.

This usually means one of:

  1. You have not added the application as a dependency in a mix.exs file.

  2. You are configuring an application that does not really exist.

Please ensure :junit_formatter exists or remove the configuration.
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,4 +1,5 @@
 import Config
 
-config :junit_formatter,
-  include_filename?: true
+if Mix.env() == :test do
+  config :junit_formatter, include_filename?: true
+end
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -1,4 +1,5 @@
 import Config
 
-config :junit_formatter,
-  include_filename?: true
+if Mix.env() == :test do
```
bblaszkow06, line 3: In config you should rather use [`config_env()`](https://hexdocs.pm/elixir/1.14/Config.html#config_env/0)

```suggestion
if config_env() == :test do
```

bblaszkow06 (approved):

PR #469, opened 2022-11-02
Create a doc describing changes in 0.11

FelonEkonom:


mat-hek:
Looks good, but it's more like release notes. This guide should be more about examples, like in Oban:
image

and instead of saying that
> In `v0.11` `:caps` option passed to `def_input_pad/2` or `def_output_pad/2` has been deleted. Instead of it, we have added new option - `:accepted_format`.

it should be something like 'instead of using `:caps` option, use `:accepted_format` option

mat-hek (changes requested):
Add the page to docs

mat-hek (approved):

PR #471, opened 2022-11-04
Update type returned from callbacks

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-19

### DIFF
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -128,7 +128,7 @@ defmodule Membrane.Core.CallbackHandler do
     maybe_context =
       case handler_params do
         %{context: context_fun} -> [context_fun.(state)]
-        _else -> []
+        _params -> []
       end
 
     args = args ++ maybe_context ++ [internal_state]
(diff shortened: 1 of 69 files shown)


FelonEkonom:
@mat-hek @varsill according to our last discussion, I've decided to rescue `UndefinedFunctionError`, when it is raised because of lack of callback implementation and override it with `Membrane.CallbackHandler` with prettier message

lib/membrane/core/callback_handler.ex
```diff
@@ -125,27 +122,38 @@ defmodule Membrane.Core.CallbackHandler do
...
+
+    case callback_result do
+      {actions, _state} when is_list(actions) ->
+        callback_result
 
-    module
-    |> apply(callback, args)
-    |> parse_callback_result(module, callback)
+      _else ->
+        raise CallbackError, kind: :bad_return, callback: {module, callback}, val: callback_result
```
mat-hek, line 155:
```suggestion
      result ->
        raise CallbackError, kind: :bad_return, callback: {module, callback}, value: result
```

mat-hek (approved):

mat-hek (approved):

PR #476, opened 2022-11-17
Fix bunch config parser warnings

varsill:
This PR:
* updates the Bunch dependency to 1.5.0. 
* removes usage of a deprecated "Bunch.Config.parse/2` `require_if` option.
* removes `:demand_unit` option from `def_*_pad` macro in places where they are not expected to appear (as they could appear before, since the `require_if` option was not working as expected)
* Makes sure no Toilet is created in the `:push` mode


lib/membrane/core/child/pads_specs.ex
```diff
@@ -183,13 +183,10 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
-                demand_unit: [
-                  in: [:buffers, :bytes],
-                  require_if:
-                    &(&1.mode == :pull and &1.demand_mode != :auto and
-                        (component == :bin or direction == :input)),
-                  default: :buffers
-                ],
+                demand_unit:
+                  &if &1.mode == :pull and (component == :bin or direction == :input) do
+                    [in: [:buffers, :bytes], default: :buffers]
```
mat-hek, line 188:
```suggestion
                    [in: [:buffers, :bytes], default: :buffers, required?: &1.demand_mode == :manual]
```

mat-hek (changes requested):

lib/membrane/core/child/pads_specs.ex
varsill: done

mat-hek (approved):

PR #477, opened 2022-11-17
Implement running all cleanup functions in ResourceGuard

FelonEkonom:


lib/membrane/resource_guard.ex
```diff
@@ -98,6 +107,15 @@ defmodule Membrane.ResourceGuard do
     {:noreply, %{state | guards: guards}}
   end
 
+  @impl true
+  def handle_info(Message.new(:cleanup_all), state) do
+    for {function, tag, timeout} <- state.guards do
+      cleanup(function, tag, timeout)
+    end
```
mat-hek, line 114: this is already done in line 136, let's move it to a private function

mat-hek (approved):

PR #483, opened 2022-11-23
Conditional links

varsill:
This PR introduces conditional links mechanism. The following syntax is now supported:
```
  child(:a, A)
  |> link_if true do
    child(:b, B)
    |> link_if false do
      child(:c, C)
    end
  end
  |> child(:d, D)
```


varsill:
Closed since the functionality can be acheived with https://github.com/membraneframework/bunch/pull/42

PR #484, opened 2022-11-25
Fixed a deadlock when pipeline spawned a spec entailing two dependent specs in a bin

mat-hek:


FelonEkonom (commented):
These changes fix a bug occurring in RTP Plugin tests 👍

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -487,23 +496,34 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
-      Enum.reduce(Map.keys(pending_specs), state, &proceed_spec_startup/2)
-    else
-      state
+    {spec_ref, state} = pop_in(state, [:pending_specs, spec_ref])
+
+    case spec_ref do
+      nil ->
+        state
+
+      %{status: status} when status in @spec_dependency_requiring_statuses ->
```
varsill, line 505: If I get it correctly, that case happens only as a result of `handle_child_death` - perhaps it could be put there? (once called in `proceed_spec_startup`, the `cleanup_spec_startup` receives an already "cleaned" state, when it comes to the `pending_specs`)

test/membrane/integration/linking_test.exs
```diff
@@ -362,6 +362,60 @@ defmodule Membrane.Integration.LinkingTest do
...
+      def handle_pad_added(Pad.ref(:output, _id) = pad, _ctx, state) do
+        {[spec: get_child(:filter) |> bin_output(pad)], state}
+      end
+    end
+
+    defmodule MultiSpecBin.Filter do
+      use Membrane.Filter
+
+      def_input_pad :input,
+        availability: :on_request,
```
varsill, line 393: Perhaps the pads/one of them could be static ones, just to cover the broader scope of cases?

lib/membrane/core/parent/child_life_controller.ex
mat-hek: It's there to be sure that after the cleanup everything is cleaned up :P If we happen to call it from another place, it should work out of the box, regardless of the circumstances

varsill (approved):
lgtm 🥇

PR #485, opened 2022-11-29
Add :handle_tick implementation to the Zombie modules

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-111


varsill (commented):
Just a general comment.- shouldn't we iterate through all the callbacks in a module (i.e. `Bin.behaviour_info(:callbacks)`) and then check, if given callback was defined with `defoverridable` (with the `Module.overridable?` function) - if so, we should call it's `super()` function, and otherwise we should return `{[], %{}}`?

mat-hek:
@FelonEkonom I'd just cancel all the timers when switching to the zombie mode

mat-hek (changes requested):

mat-hek (approved):

PR #487, opened 2022-12-05
Implement :remove_link action

FelonEkonom:


varsill (changes requested):
General question - what is the expected behaviour when removing a link, with at least one endpoint being a static pad? Shouldn't we disallow such unlinking and raise some meaningful exception? 
Nevertheless, I believe we should check that case in tests.

FelonEkonom:
> General question - what is the expected behaviour when removing a link, with at least one endpoint being a static pad? Shouldn't we disallow such unlinking and raise some meaningful exception? Nevertheless, I believe we should check that case in tests.

@varsill if you will try to unlink static pad using `:remove_link` action, element/bin with static pad will raise `Membrane.PadError` with message `"Tried to unlink a static pad #{inspect(pad_ref)}. Static pads cannot be unlinked unless element is terminating"`

varsill:
> > General question - what is the expected behaviour when removing a link, with at least one endpoint being a static pad? Shouldn't we disallow such unlinking and raise some meaningful exception? Nevertheless, I believe we should check that case in tests.
> 
> @varsill if you will try to unlink static pad using `:remove_link` action, element/bin with static pad will raise `Membrane.PadError` with message `"Tried to unlink a static pad #{inspect(pad_ref)}. Static pads cannot be unlinked unless element is terminating"`

Ok, that seems fair enough ;) Probably we don't need to tests that though

test/membrane/integration/linking_test.exs
```diff
@@ -413,6 +435,49 @@ defmodule Membrane.Integration.LinkingTest do
...
+    end
+  end
+
+  defp assert_link_removed(pipeline, id) do
+    assert_pipeline_notified(pipeline, :source, {:handle_pad_removed, Pad.ref(:output, ^id)})
+    assert_pipeline_notified(pipeline, :sink, {:handle_pad_removed, Pad.ref(:input, ^id)})
+  end
+
+  defp refute_link_removed(pipeline, id) do
+    refute_pipeline_notified(pipeline, :source, {:handle_pad_removed, Pad.ref(:output, ^id)}, 10)
```
varsill, line 477: Perhaps we could make that delay configurable as a module tag: `@refute_delay 10`

varsill (approved):

lib/membrane/core/parent/child_life_controller/link_utils.ex
```diff
@@ -32,6 +32,28 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkUtils do
...
+        for endpoint <- [link.from, link.to] do
+          Message.send(endpoint.pid, :handle_unlink, endpoint.pad_ref)
+        end
+
+        links = Map.delete(state.links, link.id)
+        Map.put(state, :links, links)
+
+      nil ->
+        raise LinkError, """
+        Attempted to unlink pad #{inspect(pad_ref)} of child #{inspect(child_name)}, but this child does not have this pad
```
mat-hek, line 52:
```suggestion
        Attempted to unlink pad #{inspect(pad_ref)} of child #{inspect(child_name)}, but no such link exists
```
We should probably first check if such a child exists to give a better error when it doesn't

mat-hek (changes requested):

lib/membrane/bin/action.ex
```diff
@@ -42,6 +42,11 @@ defmodule Membrane.Bin.Action do
            | Membrane.Child.group_t()
            | [Membrane.Child.group_t()]}
 
+  @typedoc """
+  Action that removes link, which relates to specified child and pad.
```
mat-hek, line 46: Maybe let's note that only dynamic pads can be unlinked

mat-hek (changes requested):

varsill (approved):

lib/membrane/core/parent/child_life_controller/link_utils.ex
```diff
@@ -48,6 +49,12 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkUtils do
         Map.put(state, :links, links)
 
       nil ->
+        with %{^child_name => _child_entry} <- state.children do
+          raise ParentError, """
+          Attempted to unlink pad #{inspect(pad_ref)} of child #{inspect(child_name)}, but such a child does not exist
+          """
+        end
```
mat-hek, line 56: Doesn't this check the opposite?

mat-hek (changes requested):

mat-hek (approved):

PR #489, opened 2022-12-05
[MC-107] Rename `structure` into `specification`

varsill:


lib/membrane/testing/assertions.ex
```diff
@@ -58,9 +57,7 @@ defmodule Membrane.Testing.Assertions do
   @doc """
   Asserts that pipeline received or will receive a notification from the element
   with name `element_name` within the `timeout` period specified in milliseconds.
-
   The `notification_pattern` must be a match pattern.
-
```
mat-hek, line 63: and these

mat-hek (approved):

PR #490, opened 2022-12-06
Fix pad docs generation

mat-hek:
Generates pad options without injecting HTML, so that ex_doc doesn't throw warnings and it hopefully displays correctly both in web and iex

image

image

image

image

### DIFF
--- a/lib/membrane/core/child/pads_specs.ex
+++ b/lib/membrane/core/child/pads_specs.ex
@@ -246,15 +246,23 @@ defmodule Membrane.Core.Child.PadsSpecs do
     config_doc =
       config
       |> Enum.map(&generate_pad_property_doc/1)
-      |> Enum.map_join("\n", fn {k, v} ->
-        "<tr><td>#{k}</td> <td>#{v}</td></tr>"
+      |> Enum.map_join("\n", fn
+        {k, v} ->
+          if String.contains?(v, "\n") do
+            "#{k}:\n#{v}"
+          else
+            "#{k}: | #{v}"
+          end
+
+        nil ->
+          ""
       end)
 
     options_doc =
       if pad_opts != nil do
         quote do
           """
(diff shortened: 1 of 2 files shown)


lib/membrane/core/child/pads_specs.ex
```diff
@@ -246,15 +246,23 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
+      |> Enum.map_join("\n", fn
+        {k, v} ->
+          if String.contains?(v, "\n") do
+            "#{k}:\n#{v}"
+          else
+            "#{k}: | #{v}"
+          end
+
+        nil ->
+          ""
```
FelonEkonom, line 258:
```suggestion
|> Enum.reject(&match?({:name, _name}, &1)
|> Enum.map_join("\n", fn -> 
  {:accepted_formats_str, formats} ->
    formats =       
        Enum.map_join(
          formats,
          &"""
          ```
          #{&1}
          ```
          """
        )
    "#{k}:\n#{v}"

  {property, value} ->
    property =  
      property |> to_string() |> String.replace("_", " ") |> String.capitalize()          
    "#{property}: | #{`#{inspect(value)}`}"
end)
```
Maybe do something like this and delete  `|> Enum.map(&generate_pad_property_doc/1)` from line 248? I guess, that only case, when `v` in line 251 contains `"\n"` is case with `accepted_format`. You can delete `generate_pad_property_doc` as well.

FelonEkonom (changes requested):

lib/membrane/core/child/pads_specs.ex
FelonEkonom: If there are some `"\n"` in other cases than `:accepted_format_str`, ignore this suggestion

FelonEkonom (approved):

PR #496, opened 2022-12-12
Remove :playback action

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-108

### DIFF
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -13,7 +13,7 @@ defmodule Membrane.Bin.Action do
   @typedoc """
   Action that manages the end of the component setup.
 
-  By default, component setup ends with the end of `handle_setup/2` callback.
+  By default, component setup ends with the end of `c:Membrane.Bin.handle_setup/2` callback.
   If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
   is returned from antoher callback.
   """
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -10,6 +10,13 @@ defmodule Membrane.Bin.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
+  @typedoc """
+  Action that manages the end of the component setup.
+
+  By default, component setup ends with the end of `handle_setup/2` callback.
+  If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
+  is returned from antoher callback.
+  """
   @type setup_t :: {:setup, :incomplete | :complete}
 
   @typedoc """
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -10,7 +10,7 @@ defmodule Membrane.Bin.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
-  @type setup_t :: {:setup, :defer | :complete}
+  @type setup_t :: {:setup, :incomplete | :complete}
 
   @typedoc """
   Action that sends a message to a child identified by name.
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -17,7 +17,7 @@ defmodule Membrane.Bin.Action do
   If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
   is returned from antoher callback.
 
-  Untils the setup lasts, the component won't enter `:playing` playback state.
+  Untils the setup lasts, the component won't enter `:playing` playback.
   """
   @type setup_t :: {:setup, :incomplete | :complete}
 
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -10,6 +10,8 @@ defmodule Membrane.Bin.Action do
 
   alias Membrane.{Child, ChildrenSpec}
 
+  @type setup_t :: {:setup, :defer | :complete}
+
   @typedoc """
   Action that sends a message to a child identified by name.
   """
@@ -111,7 +113,8 @@ defmodule Membrane.Bin.Action do
   other parts of framework.
   """
   @type t ::
-          notify_child_t
+          setup_t
+          | notify_child_t
           | notify_parent_t
           | spec_t
           | remove_children_t
--- a/lib/membrane/bin/action.ex
+++ b/lib/membrane/bin/action.ex
@@ -16,6 +16,8 @@ defmodule Membrane.Bin.Action do
   By default, component setup ends with the end of `c:Membrane.Bin.handle_setup/2` callback.
   If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
   is returned from antoher callback.
+
+  Untils the setup lasts, the component won't enter `:playing` playback state.
   """
   @type setup_t :: {:setup, :incomplete | :complete}
 
--- a/lib/membrane/core/element/state.ex
+++ b/lib/membrane/core/element/state.ex
@@ -36,7 +36,7 @@ defmodule Membrane.Core.Element.State do
           resource_guard: Membrane.ResourceGuard.t(),
           subprocess_supervisor: pid,
           terminating?: boolean(),
-          setup_deferred?: boolean()
+          setup_incomplete_returned?: boolean()
         }
 
   defstruct [
@@ -57,7 +57,7 @@ defmodule Membrane.Core.Element.State do
     :resource_guard,
     :subprocess_supervisor,
     :terminating?,
-    :setup_deferred?
+    :setup_incomplete_returned?
   ]
 
   @doc """
@@ -94,7 +94,7 @@ defmodule Membrane.Core.Element.State do
       resource_guard: options.resource_guard,
       subprocess_supervisor: options.subprocess_supervisor,
       terminating?: false,
-      setup_deferred?: false
+      setup_incomplete_returned?: false
     }
     |> PadSpecHandler.init_pads()
   end
--- a/lib/membrane/core/element/state.ex
+++ b/lib/membrane/core/element/state.ex
@@ -34,7 +34,9 @@ defmodule Membrane.Core.Element.State do
           playback: Membrane.Playback.t(),
           playback_queue: Membrane.Core.Element.PlaybackQueue.t(),
           resource_guard: Membrane.ResourceGuard.t(),
-          subprocess_supervisor: pid
+          subprocess_supervisor: pid,
+          terminating?: boolean(),
+          setup_deferred?: boolean()
         }
 
   defstruct [
@@ -54,7 +56,8 @@ defmodule Membrane.Core.Element.State do
     :playback_queue,
     :resource_guard,
     :subprocess_supervisor,
-    :terminating?
+    :terminating?,
+    :setup_deferred?
   ]
 
   @doc """
@@ -90,7 +93,8 @@ defmodule Membrane.Core.Element.State do
       playback_queue: [],
       resource_guard: options.resource_guard,
       subprocess_supervisor: options.subprocess_supervisor,
-      terminating?: false
+      terminating?: false,
+      setup_deferred?: false
     }
     |> PadSpecHandler.init_pads()
   end
--- a/lib/membrane/core/element/state.ex
+++ b/lib/membrane/core/element/state.ex
@@ -36,7 +36,7 @@ defmodule Membrane.Core.Element.State do
           resource_guard: Membrane.ResourceGuard.t(),
           subprocess_supervisor: pid,
           terminating?: boolean(),
-          setup_incomplete_returned?: boolean()
+          setup_incomplete?: boolean()
         }
 
   defstruct [
@@ -57,7 +57,7 @@ defmodule Membrane.Core.Element.State do
     :resource_guard,
     :subprocess_supervisor,
     :terminating?,
-    :setup_incomplete_returned?
+    :setup_incomplete?
   ]
 
   @doc """
@@ -94,7 +94,7 @@ defmodule Membrane.Core.Element.State do
       resource_guard: options.resource_guard,
       subprocess_supervisor: options.subprocess_supervisor,
       terminating?: false,
-      setup_incomplete_returned?: false
+      setup_incomplete?: false
     }
     |> PadSpecHandler.init_pads()
   end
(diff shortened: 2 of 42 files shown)


lib/membrane/pipeline/action.ex
```diff
@@ -10,6 +10,8 @@ defmodule Membrane.Pipeline.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
+  @type setup_t :: {:setup, :defer | :complete}
```
varsill, line 13: A typedoc would be required

lib/membrane/playback.ex
```diff
@@ -1,8 +1,6 @@
 defmodule Membrane.Playback do
   @moduledoc """
   Playback defines whether media is processed within a pipeline (`:playing`) or not (`:stopped`).
-
-  Playback can be controlled with `t:Membrane.Pipeline.Action.playback_t/0`.
```
varsill, line 5: Perhaps it would be good to describe when the playback is changed with the deferred setup mechanism?

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -135,6 +136,8 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         links
       end)
 
+    assert_static_pads_linked(children_definitions, links)
```
varsill, line 139: NIT 
```suggestion
    :ok = assert_static_pads_linked(children_definitions, links)
```

varsill: With such an implementation, it's clear for the code reader that we haven't forgotten about the return value

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+        state
+
+      Component.is_element?(state) ->
+        Membrane.Logger.debug("Element initialized")
+        Message.send(state.parent_pid, :initialized, state.name)
+        state
+    end
+  end
+
+  @spec assert_operation_allowed(setup_operation_t(), atom(), boolean()) :: :ok | no_return()
```
varsill, line 50: do we need that spec and the `do_assert` subfunctions for a private function?

lib/membrane/bin/action.ex
```diff
@@ -10,6 +10,8 @@ defmodule Membrane.Bin.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
+  @type setup_t :: {:setup, :defer | :complete}
```
varsill, line 13: typespec would be welcomed

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+  defp assert_operation_allowed(operation, callback, deferred?) do
+    do_assert(operation, callback, deferred?)
+  end
+
+  defp do_assert(:defer, callback, true) do
+    raise PlaybackError, """
+    Action {:setup, :defer} was returned from callback #{inspect(callback)}, but setup has already been deferred
```
varsill, line 57: I believe with that clause we posses even more information, that could be put into the error message. 
We know, that it is due to a `setup: defer` being returned by `handle_setup`, when it has already been returned by the `handle_init`

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+
+      Component.is_element?(state) ->
+        Membrane.Logger.debug("Element initialized")
+        Message.send(state.parent_pid, :initialized, state.name)
+        state
+    end
+  end
+
+  @spec assert_operation_allowed(setup_operation_t(), atom(), boolean()) :: :ok | no_return()
+  defp assert_operation_allowed(operation, callback, deferred?) do
```
varsill, line 51:
```suggestion
  defp assert_operation_allowed!(operation, callback, deferred?) do
```

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+
+      Component.is_bin?(state) ->
+        Membrane.Logger.debug("Bin initialized")
+        Message.send(state.parent_pid, :initialized, state.name)
+        state
+
+      Component.is_element?(state) ->
+        Membrane.Logger.debug("Element initialized")
+        Message.send(state.parent_pid, :initialized, state.name)
+        state
```
varsill, line 46:
```suggestion
      Component.is_child?(state) ->
        Membrane.Logger.debug("Child initialized")
        Message.send(state.parent_pid, :initialized, state.name)
        state
```

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+    :ok = assert_operation_allowed(operation, callback, state.setup_deferred?)
+
+    case operation do
+      :defer -> defer_setup(state)
+      :complete -> complete_setup(state)
+    end
+  end
+
+  @spec defer_setup(Component.state_t()) :: Component.state_t()
+  def defer_setup(state) do
```
varsill, line 24: I don't think that function is used anywhere outside of that module

test/membrane/core/pipeline_test.exs
```diff
@@ -83,7 +87,7 @@ defmodule Membrane.Core.PipelineTest do
 
       assert_raise Membrane.ParentError, ~r/.*duplicate.*\[:a\]/i, fn ->
         ActionHandler.handle_action(
-          {:spec, [child(:a, Membrane.Testing.Source)]},
+          {:spec, [child(:a, TestBin)]},
```
varsill, line 90: Is it to cover broader scope of cases?

test/membrane/integration/child_spawn_test.exs
```diff
@@ -33,23 +33,51 @@ defmodule Membrane.Integration.ChildSpawnTest do
     end
```
varsill, line 33: # TO BE CONTINUED

lib/membrane/core/playback_controller.ex
FelonEkonom: Not really, returning `setup: :defer` two times in `actions` from a single callback would also cause such a error

lib/membrane/core/playback_controller.ex
FelonEkonom: Don't you think, that having log like `17:59:38.137 [debug] <0.309.0>/:bin/ Child initialized` is not as good, as with `Bin initialized`? IMO version with `Child` might be confusing, because you don't know, if it is about initializing bin, that prints this log, or one of it's children

test/membrane/core/pipeline_test.exs
FelonEkonom: This test case without this change fails because it does not link a static pad of `Membrane.Testing.Source`

lib/membrane/bin/action.ex
```diff
@@ -10,6 +10,15 @@ defmodule Membrane.Bin.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
+  @typedoc """
+  Action that manages the end of the component setup.
+
+  By default, component setup ends with the end of `handle_setup/2` callback.
```
mat-hek, line 16:
```suggestion
  By default, component setup ends with the end of `c:Membrane.Bin.handle_setup/2` callback.
```

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,69 @@
...
+
+      Component.is_element?(state) ->
+        Membrane.Logger.debug("Element initialized")
+        Message.send(state.parent_pid, :initialized, state.name)
+        state
+    end
+  end
+
+  @spec assert_operation_allowed(setup_operation_t(), atom(), boolean()) :: :ok | no_return()
+  defp assert_operation_allowed(:incomplete, :handle_setup, true) do
```
mat-hek, line 49: checking if action is returned from a valid callback should happen in the action handler

lib/membrane/core/playback_controller.ex
```diff
@@ -0,0 +1,69 @@
+defmodule Membrane.Core.PlaybackController do
```
mat-hek, line 1: Setup isn't a part of playback. I'd handle it in the lifecycle controller

lib/membrane/core/playback_controller.ex
mat-hek: It can be determined by the `/` at the end of the path. This applies to all logs.

lib/membrane/core/pipeline/state.ex
```diff
@@ -44,7 +44,7 @@ defmodule Membrane.Core.Pipeline.State do
                 pending_specs: %{},
                 playback: :stopped,
                 initialized?: false,
-                playing_requested?: false,
-                terminating?: false
+                terminating?: false,
+                setup_incomplete_returned?: false
```
mat-hek, line 48:
```suggestion
                setup_incomplete?: false
```

mat-hek (changes requested):

lib/membrane/core/playback_controller.ex
varsill: Ok, indeed ;)

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,6 +165,37 @@ defmodule Membrane.Core.Parent.ChildLifeController do
     proceed_spec_startup(spec_ref, state)
   end
 
+  defp assert_static_pads_linked!(children_definitions, links) do
```
mat-hek, line 168:
```suggestion
  defp assert_all_static_pads_linked!(children_definitions, links) do
```

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -162,6 +165,37 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    linked_pads_set =
```
mat-hek, line 169:
```suggestion
    pads_to_link =
```

mat-hek (approved):

lib/membrane/core/element/state.ex
```diff
@@ -90,7 +93,8 @@ defmodule Membrane.Core.Element.State do
       playback_queue: [],
       resource_guard: options.resource_guard,
       subprocess_supervisor: options.subprocess_supervisor,
-      terminating?: false
+      terminating?: false,
+      setup_incomplete?: false
```
varsill, line 97: NIT - are you sure about `setup_incomplete?` name? By default it's set to `false` which might be understood that the setup is already completed :D

lib/membrane/pipeline/action.ex
```diff
@@ -10,6 +10,15 @@ defmodule Membrane.Pipeline.Action do
 
   alias Membrane.{Child, ChildrenSpec, Pad}
 
+  @typedoc """
+  Action that manages the end of the component setup.
+
+  By default, component setup ends with the end of `c:Membrane.Pipeline.handle_setup/2` callback.
+  If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
+  is returned from antoher callback.
```
varsill, line 18: How about emphasizing that until the setup lasts, the `playback` won't be `:playing`?

test/membrane/integration/child_spawn_test.exs
```diff
@@ -75,18 +103,14 @@ defmodule Membrane.Integration.ChildSpawnTest do
     pipeline_pid =
       Testing.Pipeline.start_link_supervised!(
         module: PipelineWhichDoesntPlayOnStartup,
-        custom_args: [child(:source, %Testing.Source{output: [1, 2, 3]})]
+        custom_args: [child(:source, BinWithDynamicPads)]
       )
 
     spec =
-      child(:source, %Testing.Source{output: [1, 2, 3]}, get_if_exists: true)
+      child(:source, %Testing.Source{output: ["a", "b", "c"]}, get_if_exists: true)
```
varsill, line 110: How cam that test does not fail, since the source is sending `"a"`, `"b"` and `"c"` buffers, and then you are asserting on receiving buffers with payload of form `1`, `2` and `3`

test/membrane/remote_controlled/pipeline_test.exs
```diff
@@ -41,23 +40,14 @@ defmodule Membrane.RemoteControlled.PipelineTest do
...
-      child(:c, Membrane.Testing.Sink)
-    ]
-
-    links = [
-      ChildrenSpec.get_child(:a) |> ChildrenSpec.get_child(:b) |> ChildrenSpec.get_child(:c)
-    ]
-
-    actions = [{:spec, children ++ links}]
-    Pipeline.exec_actions(pipeline, actions)
-
```
varsill, line 60: why have we removed returning `spec` action from the test setup?

test/membrane/integration/defer_setup_test.exs
```diff
@@ -0,0 +1,167 @@
...
+      )
+
+    assert_pipeline_play(pipeline)
+
+    for bin <- [:bin_1, :bin_2] do
+      refute_child_playing(pipeline, bin)
+    end
+
+    for bin <- [:bin_a, :bin_b, :bin_c] do
+      refute_grandhild_playing(pipeline, :bin_1, bin)
```
varsill, line 107:
```suggestion
      refute_grandchild_playing(pipeline, :bin_1, bin)
```

varsill, line 107:
```suggestion
      refute_grandchild_playing(pipeline, :bin_1, bin)
```

varsill (commented):
General question concerning tests - why have you changed the components used in the tests so that to make them use dynamic pads? Why not to simply don't defer the setup and let all the static pads be linked during that setup?

lib/membrane/bin/action.ex
```diff
@@ -16,6 +16,8 @@ defmodule Membrane.Bin.Action do
   By default, component setup ends with the end of `c:Membrane.Bin.handle_setup/2` callback.
   If `{:setup, :incomplete}` is returned there, setup lasts until `{:setup, :complete}`
   is returned from antoher callback.
+
+  Untils the setup lasts, the component won't enter `:playing` playback state.
```
varsill, line 20: No more are we using the "playback state" term - please use "playback" :D

FelonEkonom:
> General question concerning tests - why have you changed the components used in the tests so that to make them use dynamic pads? Why not to simply don't defer the setup and let all the static pads be linked during that setup?

Because in this PR I have added checking, if spec returned from a parent links all static pads of its children

varsill (approved):

PR #497, opened 2022-12-16
Write function to get child pid in tests

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-134


lib/membrane/testing/utils.ex
```diff
@@ -0,0 +1,49 @@
...
+   * `{:error, :child_not_found}`, if parent does not have such a child
+   * `{:error, :parent_not_alive}`, if parent is not alive
+  """
+  @spec get_child_pid(pid(), Child.name_t(), Child.child_ref_options_t()) ::
+          {:ok, pid()} | {:error, :parent_not_alive | :child_not_found}
+  def get_child_pid(parent_pid, child_name, opts \\ []) do
+    child_ref = Child.ref(child_name, opts)
+
+    try do
+      case :sys.get_state(parent_pid) do
```
mat-hek, line 25: This is meant for debugging purposes and relies on the core internals. Instead, do a normal call to the testing pipeline and make it take the PID from the context.

mat-hek (changes requested):

lib/membrane/testing/pipeline.ex
```diff
@@ -218,6 +221,60 @@ defmodule Membrane.Testing.Pipeline do
...
+
+  Id second argument is a path of child references, function gets pid of
+  last a component pointed by this path.
+
+  Returns
+   * `{:ok, child_pid}`, if a child was succesfully found
+   * `{:error, :child_not_found}`, if there is no child under provided path
+   * `{:error, :pipeline_not_alive}`, if pipeline is not alive
+  """
+  @spec get_child_pid(pid(), child_ref_path :: Child.ref_t() | [Child.ref_t()]) ::
```
varsill, line 239: [JUST TO CONSIDER, it might not make any sense at all :D ] That implementation of the get_child_pid/2 would work perfectly well in case we put i.e. a bin's pid as a parent_pid. Perhaps that function should be provided by the `Testing.Utilities` module, rather then by just `Testing.Pipeline`?

lib/membrane/testing/pipeline.ex
```diff
@@ -218,6 +221,60 @@ defmodule Membrane.Testing.Pipeline do
...
+          {:ok, pid()} | {:error, :parent_not_alive | :child_not_found}
+  def get_child_pid(parent_pid, []), do: {:ok, parent_pid}
```
varsill, line 241: I am not huge fan of that clause - we are calling `get_child_pid` to get `parent`'s pid and what we receive is the `parent_pid` itself - I know that the user shouldn't call `get_child_pid(pid, [])` on his/her own, but in fact the user can do it - perhaps it would be better not to allow that by introducing private function and implementing such a `recursion break` inside that private function?

lib/membrane/testing/pipeline.ex
```diff
@@ -218,6 +221,60 @@ defmodule Membrane.Testing.Pipeline do
...
+  end
+
+  defp do_get_child_pid(parent_pid, child_ref) do
+    msg = Message.new(:get_child_pid, child_ref)
+
+    try do
+      GenServer.call(parent_pid, msg)
+    catch
+      :exit, {:noproc, {GenServer, :call, [^parent_pid, ^msg | _tail]}} ->
+        {:error, :pipeline_not_alive}
```
varsill, line 274: We are returning an error saying that `pipeline_not_alive`, but in fact it might be a bin who is not alive

lib/membrane/testing/utils.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Membrane.Testing.Utils do
+  @moduledoc """
+  Provides functions facilitating testing Membrane components.
```
varsill, line 3:
```suggestion
  Provides functions that facilitate testing Membrane components.
```

varsill (changes requested):

lib/membrane/testing/pipeline.ex
FelonEkonom: I have added this previously to `Testing.Utils` module (see there https://github.com/membraneframework/membrane_core/pull/497/files/b93b55e1086922792f4df64d1af77a149b0724a7), but I have removed it after @mat-hek suggestion, to put it in `Testing.Pipeline`

lib/membrane/testing/pipeline.ex
FelonEkonom: Yes, but if bin is not alive, probably it will not be present in its parent `state.children` map, so recursion will return it `{:error, :child_not_found}` in previous call. But still, I can change this error to sth matching this case

lib/membrane/testing/utils.ex
```diff
@@ -0,0 +1,41 @@
...
+          {:ok, pid()} | {:error, :parent_not_alive | :child_not_found}
+  def get_child_pid(parent_pid, child_ref) do
+    msg = Message.new(:get_child_pid, child_ref)
+
+    try do
+      GenServer.call(parent_pid, msg)
+    catch
+      :exit, {:noproc, {GenServer, :call, [^parent_pid, ^msg | _tail]}} ->
+        {:error, :parent_not_alive}
+    end
```
mat-hek, line 28: use Message.call

lib/membrane/testing/utils.ex
```diff
@@ -0,0 +1,41 @@
...
+  Gets pid of child `child_ref` of parent `parent_pid`.
+
+  Returns
+   * `{:ok, child_pid}`, if parent contains child with provided name
+   * `{:error, :child_not_found}`, if parent does not have such a child
+   * `{:error, :parent_not_alive}`, if parent is not alive
+  """
+  @spec get_child_pid(pid(), Child.ref_t()) ::
+          {:ok, pid()} | {:error, :parent_not_alive | :child_not_found}
+  def get_child_pid(parent_pid, child_ref) do
```
mat-hek, line 20: This was to be moved to `Membrane.Testing.Pipeline`

mat-hek (changes requested):

test/membrane/testing/get_child_pid_test.exs
```diff
@@ -0,0 +1,104 @@
+defmodule Membrane.Testing.GetChildPidTest do
```
mat-hek, line 1: now this can be moved to testing pipeline tests

mat-hek (approved):

lib/membrane/testing/pipeline.ex
```diff
@@ -218,6 +221,65 @@ defmodule Membrane.Testing.Pipeline do
...
+  @doc """
+  Returns the pid of the children process.
+
+  Accepts pipeline pid as a first argument and a child reference or a list
+  of child references representing a path as a second argument.
+
+  If second argument is a child reference, function gets pid of this child
+  from pipeline.
+
+  Id second argument is a path of child references, function gets pid of
```
varsill, line 233:
```suggestion
  If second argument is a path of child references, function gets pid of
```

lib/membrane/testing/pipeline.ex
varsill: Ok, let's leave it that way :D

varsill (changes requested):

varsill (approved):

PR #501, opened 2022-12-27
Demand units conversion autodemands bugfix

varsill:
This PR fix a bug within the demands unit conversion that was causing the demand unit not to be converted when used with autodemands

### DIFF
--- a/lib/membrane/core/element/pad_controller.ex
+++ b/lib/membrane/core/element/pad_controller.ex
@@ -138,7 +138,7 @@ defmodule Membrane.Core.Element.PadController do
       stream_format_validation_params: stream_format_validation_params
     } = link_props
 
-    {info, other_info} = resolve_demand_unit(info, other_info)
+    {info, other_info} = resolve_demand_units(info, other_info)
 
     toilet =
       cond do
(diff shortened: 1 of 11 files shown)


lib/membrane/core/element/pad_controller.ex
```diff
@@ -211,23 +215,32 @@ defmodule Membrane.Core.Element.PadController do
     end
   end
 
-  defp resolve_demand_unit(info, other_info) do
-    if info.mode == :pull and other_info.mode == :pull do
-      case info.direction do
-        :output ->
-          demand_unit =
-            if info[:demand_unit] == nil, do: other_info.demand_unit, else: info.demand_unit
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
```
mat-hek, line 218: maybe it's no longer needed?

lib/membrane/core/element/pad_controller.ex
```diff
@@ -211,23 +215,32 @@ defmodule Membrane.Core.Element.PadController do
...
+      output_info[:demand_mode] == :manual and output_info[:demand_unit] == nil and
+          input_info[:demand_mode] == :manual ->
+        {input_info.demand_unit, input_info.demand_unit}
+
+      output_info.mode == :push and input_info[:demand_mode] == :auto ->
+        {nil, :buffers}
+
+      true ->
+        {output_info[:demand_unit], input_info[:demand_unit]}
     end
```
mat-hek, line 244: Now it looks much better, but I'm wondering if this resolution doesn't come down to:
```elixir
output_unit = output_info[:demand_unit] || input_info[:demand_unit] || :buffers
input_unit = input_info[:demand_unit] || output_info[:demand_unit] || :buffers
```

lib/membrane/pad.ex
```diff
@@ -127,7 +127,13 @@ defmodule Membrane.Pad do
...
   Describes how an output pad should be declared inside an element.
   """
-  @type output_spec_t :: {name_t(), [common_spec_options_t | {:demand_mode, demand_mode_t()}]}
+  @type output_spec_t ::
+          {name_t(),
+           [
+             common_spec_options_t
+             | {:demand_mode, demand_mode_t()}
+             | {:demand_unit, Buffer.Metric.unit_t()}
+           ]}
```
mat-hek, line 136: - now it's the same as input_spec_t
- maybe we should move these types to child or element and bin?

lib/membrane/core/element/pad_controller.ex
varsill: Technically it would work, but in case of a pad in `push` mode we would have `demand_unit` equal to `buffers` :D 
Still, it looks like a really nice generalisation - I can rewrite it as: `    output_demand_unit = if output_info[:mode] == :push, do: nil, else: output_info[:demand_unit] || input_info[:demand_unit] || :buffers
`, WDYT?

lib/membrane/core/element/pad_controller.ex
varsill: With the current implementation it's still needed, with your suggestion it's not needed anymore

mat-hek (approved):

PR #504, opened 2023-01-04
Refactor callback contexts

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-138

[MC-138]: https://membraneframework.atlassian.net/browse/MC-138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/membrane/core/child/lifecycle_controller.ex
```diff
@@ -7,13 +7,13 @@ defmodule Membrane.Core.Child.LifecycleController do
   @spec handle_parent_notification(Membrane.ParentNotification.t(), Membrane.Core.Child.state_t()) ::
           Membrane.Core.Child.state_t()
   def handle_parent_notification(notification, state) do
-    context = Component.callback_context_generator(:child, ParentNotification, state)
+    # context = Component.callback_context(state)
```
mat-hek, line 10: leftover

lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -22,13 +22,13 @@ defmodule Membrane.Core.Parent.LifecycleController do
   @spec handle_setup(Parent.state_t()) :: Parent.state_t()
   def handle_setup(state) do
     Membrane.Logger.debug("Setup")
-    context = Component.callback_context_generator(:parent, Setup, state)
+    # context = Component.callback_context(state)
```
mat-hek, line 25: Another leftover. Check if there are more

lib/membrane/core/element/callback_context.ex
```diff
@@ -1,27 +1,25 @@
...
 
-  use Membrane.Core.CallbackContext,
-    pads: %{Membrane.Pad.ref_t() => Membrane.Element.PadData.t()},
-    clock: Membrane.Clock.t() | nil,
-    parent_clock: Membrane.Clock.t() | nil,
-    name: Membrane.Element.name_t(),
-    playback: Membrane.Playback.t(),
-    resource_guard: Membrane.ResourceGuard.t(),
-    utility_supervisor: Membrane.UtilitySupervisor.t()
+  @type option_t ::
```
mat-hek, line 4: - It seems that the singular form is not needed (we can leave `options_t` only)
- This should be called `optional_fields_t` or `callback_specific_fields_t`

mat-hek (changes requested):
AFAIR we decided to document callback-specific fields in the callbacks' docs

lib/membrane/core/element/callback_context.ex
FelonEkonom: I will change a name, but singular form is used in `Membrane.Core.Component`

mat-hek: Only to create the plural form again :P

FelonEkonom: Yes, but otherwise we would have to duplicate every option in code and put eg. `{:incoming_demand, non_neg_integer()}` in `Membrane.Core.Element.CallbackContext` and in `Membrane.Core.Componenet` as well

FelonEkonom: If you know, how to do it without repeating every key and value type in these 2 modules, let me know

mat-hek: In `Membrane.Core.Component` there should be an alternative of lists instead of a list with alternative:
```
options_t :: Membrane.Core.Element.CallbackContext.options_t | Membrane.Core.Bin.CallbackContext.options_t | Membrane.Core.Pipeline.CallbackContext.options_t
```

mat-hek (approved):

PR #508, opened 2023-01-12
Unify mode and demand_mode options of pads

varsill:


lib/membrane/pad.ex
```diff
@@ -43,38 +43,36 @@ defmodule Membrane.Pad do
 
   @typedoc """
   Describes how an element sends and receives data.
-  Modes are strictly related to pad directions:
 
-  - `:push` output pad - element can send data through such pad whenever it wants.
-  - `:push` input pad - element has to deal with data whenever it comes through
+  That parameter is strictly connected with the concept of a pad mode.
```
mat-hek, line 47: I would remove pad mode and not refer to it at all, or we still end up with many terms describing a single concept

lib/membrane/core/element/action_handler.ex
```diff
@@ -383,18 +383,18 @@ defmodule Membrane.Core.Element.ActionHandler do
...
 
-      %{mode: :push} ->
+      %{flow_control: :push} ->
         raise ElementError,
               "Tried to request a demand on pad #{inspect(pad_ref)} working in push mode"
 
-      %{demand_mode: :auto} ->
+      %{flow_control: :auto} ->
         raise ElementError,
               "Tried to request a demand on pad #{inspect(pad_ref)} that has demand mode set to auto"
```
mat-hek, line 399: demand mode is no more ;) there are more occurrences, search for `demand mode`

lib/membrane/pad.ex
```diff
@@ -43,38 +43,36 @@ defmodule Membrane.Pad do
...
-
-  @typedoc """
-  Defines the mode of handling and requesting demand on pads.
+  The available values for `:flow_control` parameter are:
+  - `:manual` - meaning that the pad works in a pull mode and the demand is manually handled and requested.
+  For more info, see `Membrane.Element.Action.demand_t`, `Membrane.Element.Action.redemand_t`
+  and `c:Membrane.Element.WithOutputPads.handle_demand/5`.
+  - `:auto` - meaning that the pad works in a pull mode and the demand is managed automatically:
+  the core ensures that there's demand on each input pad (that has `demand_mode` set to `:auto`)
+  whenever there's demand on all output pads (that have `demand_mode` set to `:auto`).
```
mat-hek, line 70: demand_mode -> flow_control

lib/membrane/core/filter_aggregator/context.ex
```diff
@@ -47,7 +47,7 @@ defmodule Membrane.Core.FilterAggregator.Context do
     }
   end
 
-  defp ensure_auto_demands!(%{name: name, mode: :pull, demand_mode: mode}, module)
+  defp ensure_auto_demands!(%{name: name, flow_control: mode}, module)
```
mat-hek, line 50:
```suggestion
  defp ensure_flow_auto!(%{name: name, flow_control: mode}, module)
```

mat-hek (changes requested):

lib/membrane/core/filter_aggregator/context.ex
varsill: done

lib/membrane/pad.ex
varsill: done

lib/membrane/core/element/action_handler.ex
varsill: indeed ;) done

mat-hek (approved):

PR #509, opened 2023-01-12
remove _t suffix from types

mat-hek:
Used the following regex (in VS Code) :D
from `_t([^a-z]|$)`
to `$1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 1.0.0
  * Introduce `:remove_link` action in pipelines and bins.
  * Add children groups - a mechanism that allows refering to multiple children with a single identifier. 
- * Rename `remove_child_t` action into `remove_children_t` and allow for removing a children group with a single action.
+ * Rename `remove_child` action into `remove_children` and allow for removing a children group with a single action.
  * Add an ability to spawn anonymous children.
  * Replace `Membrane.Time.round_to_<unit_name>` with `Membrane.Time.as_<unit_name>/2` with second argument equal `:round`. Rename `Membrane.Time.round_to_timebase` to `Membrane.Time.divide_by_timebase/2`. [#494](https://github.com/membraneframework/membrane_core/pull/494)
  * Remove `:playback` action. Introduce `:setup` action. [#496](https://github.com/membraneframework/membrane_core/pull/496)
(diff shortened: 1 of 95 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## 1.0.0
```
varsill, line 3: I would add a point saying that we have renamed types there

guides/upgrading/v0.11.md
```diff
@@ -292,7 +292,7 @@ Rename `Membrane.Time.to_<unit name>/1` into `Membrane.Time.round_to_<unit name>
 ```
 
 ## Update the children definitions
-Children defintions syntax (previously known as `ParentSpec`, after the name of a structure used to define children), that was used in `Membrane.Pipeline.Action.spec_t` and `Membrane.Bin.Action.spec_t` actions, has changed. 
+Children defintions syntax (previously known as `ParentSpec`, after the name of a structure used to define children), that was used in `Membrane.Pipeline.Action.spec` and `Membrane.Bin.Action.spec` actions, has changed.
```
varsill, line 295: I am not sure we should change it there

lib/membrane/core/pipeline/zombie.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Membrane.Core.Pipeline.Zombie do
   @moduledoc false
-  # When a pipeline returns Membrane.Pipeline.Action.terminate_t()
+  # When a pipeline returns Membrane.Pipeline.Action.terminate()
   # and becomes a zombie-ie-ie oh oh oh oh oh oh oh, ay, oh, ya ya
```
varsill, line 4: I wasn't aware that one is still there :D

varsill (changes requested):
We need to preserve that regex for the posterity :D

lib/membrane/core/pipeline/zombie.ex
mat-hek: It was actually added in 0.11 IIRC :P

guides/upgrading/v0.11.md
mat-hek: Why so?

lib/membrane/core/pipeline/zombie.ex
varsill: I know, but I thought it was changed before merging :D

guides/upgrading/v0.11.md
varsill: Well, we are refering to somthing `that WAS used in ...` - back in the time, it was `Membrane.Pipeline.Action.spec_t`, not `Membrane.Pipeline.Action.spec` :D

varsill (approved):
LGTM 🥇

PR #517, opened 2023-01-26
Update processed specs, when children or link is down

FelonEkonom:
Rekated Jira ticket: https://membraneframework.atlassian.net/browse/MC-148

### DIFF
--- a/lib/membrane/core/bin/pad_controller.ex
+++ b/lib/membrane/core/bin/pad_controller.ex
@@ -80,8 +80,8 @@ defmodule Membrane.Core.Bin.PadController do
     state
   end
 
-  @spec remove_dynamic_pad(Pad.ref(), State.t()) :: State.t()
-  def remove_dynamic_pad(pad_ref, state) do
+  @spec remove_dynamic_pad!(Pad.ref(), State.t()) :: State.t()
+  def remove_dynamic_pad!(pad_ref, state) do
     case pad_ref do
       Pad.ref(_name, _id) ->
         Message.send(state.parent_pid, :child_pad_removed, [state.name, pad_ref])
--- a/lib/membrane/core/bin/pad_controller.ex
+++ b/lib/membrane/core/bin/pad_controller.ex
@@ -266,15 +266,15 @@ defmodule Membrane.Core.Bin.PadController do
   @doc """
   Handles situation where the pad has been unlinked (e.g. when connected element has been removed from the pipeline)
   """
-  @spec handle_unlink(Pad.ref(), :soft | :hard, Core.Bin.State.t()) :: Core.Bin.State.t()
-  def handle_unlink(pad_ref, mode, state) do
+  @spec handle_unlink(Pad.ref(), Core.Bin.State.t()) :: Core.Bin.State.t()
+  def handle_unlink(pad_ref, state) do
     with {:ok, %{availability: :on_request}} <- PadModel.get_data(state, pad_ref) do
       state = maybe_handle_pad_removed(pad_ref, state)
       endpoint = PadModel.get_data!(state, pad_ref, :endpoint)
       {pad_data, state} = PadModel.pop_data!(state, pad_ref)
 
       if endpoint do
-        Message.send(endpoint.pid, :handle_unlink, [endpoint.pad_ref, mode])
+        Message.send(endpoint.pid, :handle_unlink, endpoint.pad_ref)
         ChildLifeController.proceed_spec_startup(pad_data.spec_ref, state)
       else
         Membrane.Logger.debug("""
@@ -285,7 +285,7 @@ defmodule Membrane.Core.Bin.PadController do
         state
       end
     else
-      {:ok, %{availability: :always}} when state.terminating? or mode == :soft ->
+      {:ok, %{availability: :always}} when state.terminating? ->
         state
 
       {:ok, %{availability: :always}} ->
--- a/lib/membrane/core/bin/pad_controller.ex
+++ b/lib/membrane/core/bin/pad_controller.ex
@@ -80,8 +80,8 @@ defmodule Membrane.Core.Bin.PadController do
     state
   end
 
-  @spec remove_dynamic_pad!(Pad.ref(), State.t()) :: State.t()
-  def remove_dynamic_pad!(pad_ref, state) do
+  @spec remove_pad!(Pad.ref(), State.t()) :: State.t()
+  def remove_pad!(pad_ref, state) do
     cond do
       Pad.is_dynamic_pad_ref(pad_ref) ->
         Message.send(state.parent_pid, :child_pad_removed, [state.name, pad_ref])
--- a/lib/membrane/core/bin/pad_controller.ex
+++ b/lib/membrane/core/bin/pad_controller.ex
@@ -82,15 +82,15 @@ defmodule Membrane.Core.Bin.PadController do
 
   @spec remove_dynamic_pad!(Pad.ref(), State.t()) :: State.t()
   def remove_dynamic_pad!(pad_ref, state) do
-    case pad_ref do
-      Pad.ref(_name, _id) ->
+    cond do
+      Pad.is_dynamic_pad_ref(pad_ref) ->
         Message.send(state.parent_pid, :child_pad_removed, [state.name, pad_ref])
         PadModel.delete_data!(state, pad_ref)
 
-      name when is_atom(name) and state.terminating? ->
+      Pad.is_static_pad_ref(pad_ref) and state.terminating? ->
         state
 
-      name when is_atom(name) ->
+      Pad.is_static_pad_ref(pad_ref) ->
         raise Membrane.PadError,
               "Tried to unlink bin static pad #{inspect(pad_ref)}. Static pads cannot be unlinked unless bin is terminating"
     end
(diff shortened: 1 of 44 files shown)


lib/membrane/core/parent/child_life_controller/link_utils.ex
```diff
@@ -33,22 +34,48 @@ defmodule Membrane.Core.Parent.ChildLifeController.LinkUtils do
...
+        %Endpoint{child: {Membrane.Bin, :itself}} = bin_endpoint ->
+          PadController.remove_dynamic_pad!(bin_endpoint.pad_ref, state)
+
+        %Endpoint{} = endpoint ->
           Message.send(endpoint.pid, :handle_unlink, endpoint.pad_ref)
-        end
+          state
+      end
+
+    state = ChildLifeController.remove_link_from_spec(link.id, state)
```
varsill, line 52: [SUGGESTION] I am not biggest fan of these circular dependencies :D 
We are calling `handle_child_pad_removed()` from the ChildLifeController, and inside of that function we are calling another function from `ChildLifeController`. How about putting `remove_link_from_spec()` in `LinkUtils`?

varsill (changes requested):

lib/membrane/pipeline.ex
```diff
@@ -135,6 +135,19 @@ defmodule Membrane.Pipeline do
             ) ::
               callback_return
 
+  @doc """
+  Callback invoked when a child removes its pad.
+
+  Removing child's pad due to return `t:Membrane.Bin.Action.remove_link()`
```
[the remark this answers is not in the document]
FelonEkonom: No, this version is correct. Now bin can remove his own, external pad by unlinking his child with this pad

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -80,6 +80,22 @@ defmodule Membrane.Core.Bin.PadController do
     state
   end
 
+  @spec remove_dynamic_pad!(Pad.ref(), State.t()) :: State.t()
+  def remove_dynamic_pad!(pad_ref, state) do
```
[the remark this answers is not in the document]
FelonEkonom: This function fails, if passed pad is static - only exception is when component is in `terminating` state. What name do you suggest?

lib/membrane/testing/pipeline.ex
```diff
@@ -482,8 +514,8 @@ defmodule Membrane.Testing.Pipeline do
 
   defp eval_injected_module_callback(callback, args, state)
 
-  defp eval_injected_module_callback(_callback, _args, %State{module: nil} = state),
-    do: {[], state}
+  defp eval_injected_module_callback(_callback, _args, %State{module: nil}),
+    do: {[], nil}
```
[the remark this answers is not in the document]
FelonEkonom: This change fixes weird, nested multiple times, custom state in Testing.Pipeline, where custom module is `nil`

lib/membrane/core/parent/child_life_controller/link_utils.ex
FelonEkonom: I agree, but I would refactor this along with crash groups

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -564,21 +691,18 @@ defmodule Membrane.Core.Parent.ChildLifeController do
     %{pid: child_pid} = ChildrenModel.get_child_data!(state, child_name)
 
     with {:ok, group} <- CrashGroupUtils.get_group_by_member_pid(child_pid, state) do
+      existing_members =
+        group.members
+        |> Enum.filter(&Map.has_key?(state.children, &1))
+
+      state = remove_children_from_specs(existing_members, state)
+      state = Enum.reduce(existing_members, state, &LinkUtils.unlink_element/2)
```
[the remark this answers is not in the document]
FelonEkonom: I removed the line with `LinkUtils.unlink_element`, because elements are already unlinked in `crash_all_group_members/3`

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -193,58 +209,61 @@ defmodule Membrane.Core.Bin.PadController do
...
+          other_endpoint,
+          params
+        ])
+
+      state = PadModel.set_data!(state, endpoint.pad_ref, :linked?, true)
+      state = PadModel.set_data!(state, endpoint.pad_ref, :endpoint, child_endpoint)
+      state = ChildLifeController.proceed_spec_startup(spec_ref, state)
+      {reply, state}
+    else
+      {:error, :unknown_pad} ->
```
mat-hek, line 264: - handle this error in element & parent
- add a comment on when this can happen here and in similarly weird places

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -522,24 +522,154 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    spec_dependencies
+    |> Enum.map(fn {spec_ref, spec_children} ->
+      {
+        spec_ref,
+        Enum.reject(spec_children, &(&1 in children_removed_from_spec))
+      }
+    end)
+    |> Enum.reject(&match?({_ref, []}, &1))
+    |> Map.new()
+  end
```
mat-hek, line 645: - don't proceed with a dependent spec even if the dependency is removed
- simplify the code by iterating through children first and making use of spec_ref in child and link data

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -522,24 +519,121 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    removed_children = Bunch.listify(removed_children) |> MapSet.new()
+
+    removed_links_ids =
+      state.links
+      |> Map.values()
+      |> Enum.filter(
+        &(MapSet.member?(removed_children, &1.from.child) or
+            MapSet.member?(removed_children, &1.to.child))
+      )
+      |> MapSet.new(& &1.id)
```
mat-hek, line 546: instead of getting id, we could get the spec ref

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -522,24 +519,121 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+
+    updated_specs =
+      state.pending_specs
+      |> Enum.filter(fn {_ref, spec_data} ->
+        not MapSet.disjoint?(spec_data.children_names, removed_children) or
+          not MapSet.disjoint?(spec_data.links_ids, removed_links_ids)
+      end)
```
mat-hek, line 553: we can get links specs as stated in the comment above and children specs from children data, so that this won't be needed

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -522,24 +519,121 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+        |> MapSet.new()
+
+      state =
+        put_in(
+          state,
+          [:pending_specs, spec_ref],
+          %{spec_data | links_ids: links_ids, awaiting_responses: awaiting_responses}
+        )
+
+      proceed_spec_startup(spec_ref, state)
```
mat-hek, line 607: It still duplicates the code from `remove_children_from_specs`

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -621,15 +711,15 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
-      |> Enum.filter(fn {_ref, data} -> spec_ref in data.dependent_specs end)
+      |> Enum.filter(fn {_ref, data} -> MapSet.member?(data.dependent_specs, spec_ref) end)
       |> Map.new(fn {ref, data} ->
         {ref, Map.update!(data, :dependent_specs, &MapSet.delete(&1, spec_ref))}
       end)
 
-    state = %{state | pending_specs: Map.merge(state.pending_specs, dependent_specs)}
-    dependent_specs |> Map.keys() |> Enum.reduce(state, &proceed_spec_startup/2)
+    state = %{state | pending_specs: Map.merge(state.pending_specs, dependencies)}
+    dependencies |> Map.keys() |> Enum.reduce(state, &proceed_spec_startup/2)
```
mat-hek, line 722: seems these are dependent_specs again

lib/membrane/pipeline.ex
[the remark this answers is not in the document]
mat-hek: I'm not sure if this changed, but I still have a hard time understanding this. The doc for this callback in Bin is even more sophisticated since there is a 'parent' and 'child' bin involved and it's hard to determine which is which. Maybe let's focus on when this callback _is_ called and provide some example/s.

lib/membrane/testing/pipeline.ex
```diff
@@ -466,6 +475,29 @@ defmodule Membrane.Testing.Pipeline do
...
   end
 
+  @impl true
+  def handle_child_pad_removed(child, pad, ctx, state) do
+    if state.raise_on_child_pad_removed? do
+      raise """
+      Child #{inspect(child)} removed it's pad #{inspect(pad)}. If you want to
+      handle such a scenario, pass `raise_on_child_pad_removed?: false` option to
+      `Membrane.Testing.Pipeline.start_*/2` or pass there a pipeline module
+      implementing this callback via `:name` option.
```
mat-hek, line 485: `name` option?

mat-hek (changes requested):

mat-hek (approved):

varsill (approved):

PR #519, opened 2023-01-30
Update Membrane.Pipeline.terminate/2 behavior

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-149

### DIFF
--- a/lib/membrane/pipeline.ex
+++ b/lib/membrane/pipeline.ex
@@ -61,7 +61,7 @@ defmodule Membrane.Pipeline do
   use Bunch
 
   alias __MODULE__.{Action, CallbackContext}
-  alias Membrane.{Child, Pad, PipelineTerminationError}
+  alias Membrane.{Child, Pad, PipelineError}
 
   require Membrane.Logger
   require Membrane.Core.Message, as: Message
@@ -301,37 +301,43 @@ defmodule Membrane.Pipeline do
     * `asynchronous?` - if set to `true`, pipline termination won't be blocking and
       will be executed in the process, which pid is returned as function result. If
       set to `false`, pipeline termination will be blocking and will be executed in
-      the process, that called this function.
+      the process that called this function. Defaults to `false`.
     * `timeout` - tells how much time (ms) to wait for pipeline to get gracefully
       terminated. Defaults to 5000.
     * `force?` - if set to `true` and pipeline is still alive after `timeout`,
-      pipeline will be killed using `Process.exit/2` with second argument set to
-      `:kill`, and function will return `{:error, :timeout}`. If set to `false` and
-      pipeline is still alive after `timeout`, function will raise an error.
-      Defaults to `false`.
+      pipeline will be killed using `Process.exit/2` with reason `:kill`, and function
+      will return `{:error, :timeout}`. If set to `false` and pipeline is still alive
+      after `timeout`, function will raise an error. Defaults to `false`.
 
   Returns:
     * `{:ok, pid}` - if option `asynchronous?: true` was passed.
     * `:ok` - if pipeline was gracefully terminated within `timeout`.
     * `{:error, :timeout}` - if pipeline was killed after a `timeout`.
   """
-  @spec terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
+  @spec terminate(pipeline :: pid, timeout: timeout(), force?: boolean(), asynchronous?: boolean()) ::
           :ok | {:ok, pid()} | {:error, :timeout}
   def terminate(pipeline, opts \\ []) do
-    {asynchronous?, opts} = Keyword.pop(opts, :asynchronous?, false)
+    [asynchronous?: asynchronous?] ++ opts =
+      Keyword.validate!(opts,
+        asynchronous?: false,
+        force?: false,
+        timeout: 5000
+      )
+      |> Enum.sort()
 
     if asynchronous? do
-      Task.start(__MODULE__, :run_terminate, [pipeline, opts])
+      Task.start(__MODULE__, :do_terminate, [pipeline, opts])
     else
-      run_terminate(pipeline, opts)
+      do_terminate(pipeline, opts)
     end
   end
 
-  @spec run_terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
+  @doc false
+  @spec do_terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
           :ok | {:error, :timeout}
-  def run_terminate(pipeline, opts) do
-    timeout = Keyword.get(opts, :timeout, 5000)
-    force? = Keyword.get(opts, :force?, false)
+  def do_terminate(pipeline, opts) do
+    timeout = Keyword.get(opts, :timeout)
+    force? = Keyword.get(opts, :force?)
 
     ref = Process.monitor(pipeline)
     Message.send(pipeline, :terminate)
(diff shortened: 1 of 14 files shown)


lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
+
+  Accepts three options:
+    * `asynchronous?` - if set to `true`, pipline termination won't be blocking and
+      will be executed in the process, which pid is returned as function result. If
+      set to `false`, pipeline termination will be blocking and will be executed in
+      the process, that called this function.
+    * `timeout` - tells how much time (ms) to wait for pipeline to get gracefully
+      terminated. Defaults to 5000.
+    * `force?` - if set to `true` and pipeline is still alive after `timeout`,
+      pipeline will be killed using `Process.exit/2` with second argument set to
```
mat-hek, line 308:
```suggestion
      pipeline will be killed using `Process.exit/2` with reason
```

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
-    * `blocking?` - tells whether to stop the pipeline synchronously
-    * `timeout` - if `blocking?` is set to true it tells how much
-      time (ms) to wait for pipeline to get terminated. Defaults to 5000.
+  Terminates the pipeline.
+
+  Accepts three options:
+    * `asynchronous?` - if set to `true`, pipline termination won't be blocking and
+      will be executed in the process, which pid is returned as function result. If
+      set to `false`, pipeline termination will be blocking and will be executed in
+      the process, that called this function.
```
mat-hek, line 304:
```suggestion
      the process that called this function.
```

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
-          :ok | {:error, :timeout}
+  @spec terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
+          :ok | {:ok, pid()} | {:error, :timeout}
   def terminate(pipeline, opts \\ []) do
-    blocking? = Keyword.get(opts, :blocking?, false)
-    timeout = Keyword.get(opts, :timeout, 5000)
-
-    ref = if blocking?, do: Process.monitor(pipeline)
-    Message.send(pipeline, :terminate)
+    {asynchronous?, opts} = Keyword.pop(opts, :asynchronous?, false)
```
mat-hek, line 321: let's use `Keyword.validate!` to get all the options

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
       {:DOWN, ^ref, _process, _pid, _reason} ->
         :ok
     after
       timeout ->
-        {:error, :timeout}
+        if force? do
+          Process.exit(pipeline, :kill)
+          {:error, :timeout}
+        else
+          raise PipelineTerminationError, """
```
mat-hek, line 348: Having a separate error for that is an overkill, let's raise `PipelineError`

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
+    end
   end
 
-  @spec call(pid, any, integer()) :: :ok
-  def call(pipeline, message, timeout \\ 5000) do
-    GenServer.call(pipeline, message, timeout)
-  end
+  @spec run_terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
+          :ok | {:error, :timeout}
+  def run_terminate(pipeline, opts) do
```
mat-hek, line 332: This should have `@doc false`. `do_terminate` would be a better name too.

mat-hek (changes requested):

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
+    * `asynchronous?` - if set to `true`, pipline termination won't be blocking and
+      will be executed in the process, which pid is returned as function result. If
+      set to `false`, pipeline termination will be blocking and will be executed in
+      the process, that called this function.
+    * `timeout` - tells how much time (ms) to wait for pipeline to get gracefully
+      terminated. Defaults to 5000.
+    * `force?` - if set to `true` and pipeline is still alive after `timeout`,
+      pipeline will be killed using `Process.exit/2` with second argument set to
+      `:kill`, and function will return `{:error, :timeout}`. If set to `false` and
+      pipeline is still alive after `timeout`, function will raise an error.
```
varsill, line 310: [SOME THOUGHTS] Hmm should it raise under these circumstances? I think it would be better if we have simply returned the `:error` tuple, and let the user raise on her/his own

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
-  Gracefully terminates the pipeline.
-
-  Accepts two options:
-    * `blocking?` - tells whether to stop the pipeline synchronously
-    * `timeout` - if `blocking?` is set to true it tells how much
-      time (ms) to wait for pipeline to get terminated. Defaults to 5000.
+  Terminates the pipeline.
+
+  Accepts three options:
+    * `asynchronous?` - if set to `true`, pipline termination won't be blocking and
```
varsill, line 301: I believe the default for that option is `false` - it's worth mentioning in docs

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
+      Defaults to `false`.
+
+  Returns:
+    * `{:ok, pid}` - if option `asynchronous?: true` was passed.
+    * `:ok` - if pipeline was gracefully terminated within `timeout`.
+    * `{:error, :timeout}` - if pipeline was killed after a `timeout`.
   """
-  @spec terminate(pipeline :: pid, Keyword.t()) ::
-          :ok | {:error, :timeout}
+  @spec terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
```
varsill, line 318: You have missed the `asynchronous?` option in the spec

lib/membrane/pipeline.ex
```diff
@@ -295,43 +295,69 @@ defmodule Membrane.Pipeline do
...
+    else
+      run_terminate(pipeline, opts)
+    end
   end
 
-  @spec call(pid, any, integer()) :: :ok
-  def call(pipeline, message, timeout \\ 5000) do
-    GenServer.call(pipeline, message, timeout)
-  end
+  @spec run_terminate(pipeline :: pid, timeout: timeout(), force?: boolean()) ::
```
varsill, line 330: That one should be somehow made "private". I known you have made it public since you can use it both directly and with the asynchronous `Task`, but now we will have `Membrane.Pipeline.run_terminate/2` function in the API, with a defective options list ;D

lib/membrane/rc_pipeline.ex
```diff
@@ -392,7 +392,7 @@ defmodule Membrane.RCPipeline do
   def handle_terminate_request(_ctx, state) do
     pipeline_event = %Terminated{from: self()}
     send_event_to_controller_if_subscribed(pipeline_event, state)
-    {[], state}
+    {[terminate: :normal], state}
```
varsill, line 395: Oh, that's odd that that one wasn't terminating :O

varsill (changes requested):
One more thing - I haven't seen the usage of the `force?: true` option anywhere in the tests. Perhaps we should check if it's working with that option properly

mat-hek (approved):

varsill (approved):

PR #521, opened 2023-01-31
Crash groups in bins

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-143


lib/membrane/children_spec.ex
```diff
@@ -318,7 +318,7 @@ defmodule Membrane.ChildrenSpec do
 
   @type children_spec_options :: [
           group: Child.group(),
-          crash_group_mode: Membrane.CrashGroup.mode() | nil,
+          crash_group_mode: Membrane.CrashGroup.mode(),
```
mat-hek, line 321:
```suggestion
          crash_group_mode: :temporary | nil,
```

and we can remove `Membrane.CrashGroup` altogether.

FelonEkonom: It is used in 2 other places as well, so I can remove this `CrashGroup` module and move typespec to the another file, but I would still have a typespec describing `:temporary | nil`

mat-hek (approved):

PR #529, opened 2023-02-16
Revert "Update processed specs, when children or link is down"

FelonEkonom:
Reverts membraneframework/membrane_core#517


mat-hek:
wait, what?

PR #531, opened 2023-02-17
Add ability to set push flow

mat-hek:


lib/membrane/core/element/pad_controller.ex
```diff
@@ -124,12 +148,15 @@ defmodule Membrane.Core.Element.PadController do
...
          toilet,
+         metrics_data,
          %{initiator: :sibling} = link_props,
          state
        ) do
     %{other_info: other_info, link_metadata: link_metadata} = link_props
 
-    link_metadata = %{link_metadata | toilet: link_metadata.toilet || toilet}
+    link_metadata =
+      %{link_metadata | toilet: link_metadata.toilet || toilet}
```
FelonEkonom, line 158: Sentence like `%{a | b: c.f || e}` is quite unreadable IMO

lib/membrane/core/metrics.ex
```diff
@@ -0,0 +1,83 @@
...
+        key =
+          {unquote(metric), unquote(opts)[:component_path] || Membrane.ComponentPath.get(),
+           unquote(opts)[:id]}
+
+        [{_key, value} | _default] =
+          :ets.lookup(unquote(__MODULE__), key) ++ [{nil, unquote(init)}]
+
+        :ets.insert(unquote(__MODULE__), {key, unquote(fun).(value)})
+      end
+    end
```
FelonEkonom, line 33: The only call to `report_update` is adding number to sum, so you can use https://www.erlang.org/doc/man/ets.html#update_counter-4 instead. `:ets.update_counter/4` is atomic. There is an example, how to use it: https://github.com/membraneframework/membrane_telemetry_metrics/blob/a46281fd0c43f3f31600b24b96416b9e563b337b/lib/telemetry_metrics/reporter/sum.ex#L23

PR #537, opened 2023-03-13
Auto push full impl

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-163


lib/membrane/core/element/demand_controller.ex
```diff
@@ -5,122 +5,167 @@ defmodule Membrane.Core.Element.DemandController do
...
-          split_continuation_arbiter: &exec_handle_demand?(PadModel.get_data!(&1, pad_ref)),
-          context: context
-        },
-        [pad_ref, demand, data[:demand_unit]],
-        state
-      )
+  defp do_check_demand_counter(%{flow_control: :manual} = pad_data, state) do
+    with %{demand: demand, demand_counter: demand_counter} when demand <= 0 <- pad_data,
+         counter_value when counter_value > 0 and counter_value > demand <-
+           DemandCounter.get(demand_counter) do
```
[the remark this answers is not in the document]
varsill: I am back :D After inspecting InputQueue I realised that `demand_counter` in fact holds the `size` of a demand (expressed in the demand unit of the proceeding element's output pad). For me it would be much clearer to use `demand_size` instead of `demand` but I am not insisting since I'm aware that sometimes I'm extravagantly keen on descriptive names. 
With such a change we could have  `pad_demand_size`, `demand_counter` and `counters_demand_size`

lib/membrane/core/element/demand_controller.ex
```diff
@@ -5,122 +5,167 @@ defmodule Membrane.Core.Element.DemandController do
...
-  def send_auto_demand_if_needed(pad_ref, demand_decrease \\ 0, state) do
-    data = PadModel.get_data!(state, pad_ref)
+  @spec handle_outgoing_buffers(
+          Pad.ref(),
+          [Buffer.t()],
+          State.t()
+        ) :: State.t()
+  def handle_outgoing_buffers(pad_ref, buffers, state) do
+    pad_data = PadModel.get_data!(state, pad_ref)
+    buffers_size = Buffer.Metric.from_unit(pad_data.other_demand_unit).buffers_size(buffers)
```
varsill, line 112: Are you sure we should use `other_demand_unit` here? We are working on an `output` pad and the demand unit is expressed in the metric specified by the `output` pad.

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,354 @@
...
+    # The counter uses :atomics module under the hood.
+    # The module allows to create and modify the value of a counter in the same manner both when the counter is about to be accessed
+    # from the same node, and from different nodes.
+
+    @enforce_keys [:worker, :atomic_ref]
+    defstruct @enforce_keys
+
+    @type t :: %__MODULE__{worker: Worker.t(), atomic_ref: :atomics.atomics_ref()}
+
+    defguardp on_this_same_node_as_self(distributed_atomic)
```
varsill, line 71:
```suggestion
    defguardp on_the_same_node_as_self(distributed_atomic)
```

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,354 @@
...
+              when distributed_atomic.worker |> node() == self() |> node()
+
+    @spec new(integer() | nil) :: t
+    def new(initial_value \\ nil) do
```
varsill, line 75: [GENERAL IDEA] Do we really need to start the worker in case it we are operating on the same node?

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,354 @@
...
+
+  @type flow_mode :: DistributedFlowMode.flow_mode_value()
+
+  @enforce_keys [
+    :counter,
+    :receiver_mode,
+    :receiver_process,
+    :sender_mode,
+    :sender_process,
+    :sender_pad_ref,
```
varsill, line 187: These three are not in the `@type t` definition

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,354 @@
...
+    defp int_to_flow_mode(2), do: :pull
+
+    defp flow_mode_to_int(:to_be_resolved), do: 0
+    defp flow_mode_to_int(:push), do: 1
+    defp flow_mode_to_int(:pull), do: 2
+  end
+
+  @default_overflow_limit_factor -200
+  @default_buffered_decrementation_limit 1
+  @distributed_buffered_decrementation_limit 150
```
varsill, line 168: That one used to be configurable from the `pad` options - we can reconsider whether it should or shouldn't stay that way

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,354 @@
...
+
+  defstruct @enforce_keys ++ [buffered_decrementation: 0, toilet_overflowed?: false]
+
+  @spec new(
+          receiver_mode :: EffectiveFlowController.effective_flow_control(),
+          receiver_process :: Process.dest(),
+          receiver_demand_unit :: Membrane.Buffer.Metric.unit(),
+          sender_process :: Process.dest(),
+          sender_pad_ref :: Pad.ref(),
+          overflow_limit :: neg_integer() | nil
```
varsill, line 200: [NIT] Perhaps we should be more consistent with the naming and have either `toilet_capacity` or the `overflow_limit`?

lib/membrane/core/element/pad_controller.ex
```diff
@@ -149,34 +156,35 @@ defmodule Membrane.Core.Element.PadController do
...
+      EffectiveFlowController.pad_effective_flow_control(endpoint.pad_ref, state)
+
+    demand_counter =
+      DemandCounter.new(
+        pad_effective_flow_control,
+        self(),
+        input_demand_unit || :buffers,
+        other_endpoint.pid,
+        other_endpoint.pad_ref,
+        -300
```
varsill, line 181: I believe we should have `toilet_capacity` passed here

lib/membrane/core/element/demand_counter.ex
varsill, line 1: `DemandCounter` consists of `DistributedFlowMode` and `DistributedAtomic`. `DistributedFlowMode` is built on top of `DistributedAtomic` and `DistributedAtomic` uses `Worker`. 
How about moving the `DistributedAtomic` to the "private" part of the `DemandCounter` (i.e. putting it in some kind of a separate namespace), along with the `Worker`, leaving the `DistributedFlowMode` the way it is and adding a `DistributedCounter` built on top of `DistributedAtomic` (on the same "level" as the `DistributedFlowMode`)? I believe that the responsibilities of the modules would be then slightly less mixed.

varsill (changes requested):

lib/membrane/core/element/input_queue.ex
```diff
@@ -24,31 +25,30 @@ defmodule Membrane.Core.Element.InputQueue do
           {:event | :stream_format, any} | {:buffers, list, pos_integer, pos_integer}
   @type output :: {:empty | :value, [output_value]}
 
+  @type queue_item() :: Buffer.t() | Event.t() | struct() | atom()
```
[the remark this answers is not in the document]
FelonEkonom: `:atom` was for `start`/`end_of_stream`, as far as I remember

lib/membrane/core/element/demand_handler.ex
```diff
@@ -155,25 +109,35 @@ defmodule Membrane.Core.Element.DemandHandler do
...
-        state =
-          case action do
-            :supply -> do_supply_demand(pad_ref, state)
-            :redemand -> handle_redemand(pad_ref, state)
-          end
+      state.handle_demand_loop_counter >= @handle_demand_loop_limit ->
+        Message.self(:resume_handle_demand_loop)
+        %{state | handle_demand_loop_counter: 0}
+
+      state.delayed_demands == MapSet.new() ->
```
mat-hek, line 126:
```suggestion
      Enum.empty?(state.delayed_demands) ->
```

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,127 @@
+defmodule Membrane.Core.Element.EffectiveFlowController do
```
mat-hek, line 1: Let's add description what is effective flow control and how it changes

lib/membrane/element/pad_data.ex
```diff
@@ -39,13 +39,16 @@ defmodule Membrane.Element.PadData do
           other_ref: private_field,
           input_queue: private_field,
           demand: integer() | nil,
+          incoming_demand: integer() | nil,
```
mat-hek, line 42: why is it here?

lib/membrane/element/pad_data.ex
```diff
@@ -39,13 +39,16 @@ defmodule Membrane.Element.PadData do
...
           demand_unit: private_field,
           other_demand_unit: private_field,
           auto_demand_size: private_field,
           sticky_messages: private_field,
-          toilet: private_field,
+          demand_counter: private_field,
+          lacking_buffers: private_field,
```
mat-hek, line 48: describe somewhere what is it about

mat-hek (commented):
TBC

varsill (changes requested):

lib/membrane/core/element/demand_counter.ex
FelonEkonom: No, but in such a case the worker won't do any additional operations, so there is no extra performance overhead caused by it

lib/membrane/core/element/demand_handler.ex
FelonEkonom: Calling `Enum.empty?` for MapSet has linear complexity while comparing with `MapSet.new()` is performed in constant time. This is the same case as in `if length(my_list) == 0 do ...`, which would cause generating a warning with a suggestion to replace it with `if my_list == [] do ...`

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+    state = PadModel.set_data!(state, my_pad_ref, pad_data)
+
+    cond do
+      state.playback != :playing or pad_data.direction != :input or pad_data.flow_control != :auto ->
+        state
+
+      other_effective_flow_control == state.effective_flow_control ->
+        :ok =
+          PadModel.get_data!(state, my_pad_ref, :demand_counter)
+          |> DemandCounter.set_receiver_mode(state.effective_flow_control)
```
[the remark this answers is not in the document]
FelonEkonom: Because sender sets `receiver_mode` to `:to_be_resolved`, this line is to revert this change, when we don't have to resolve our own `effective_flow_control`

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+          State.t()
+        ) ::
+          State.t()
+  def handle_other_effective_flow_control(my_pad_ref, other_effective_flow_control, state) do
+    pad_data = PadModel.get_data!(state, my_pad_ref)
+    pad_data = %{pad_data | other_effective_flow_control: other_effective_flow_control}
+    state = PadModel.set_data!(state, my_pad_ref, pad_data)
+
+    cond do
+      state.playback != :playing or pad_data.direction != :input or pad_data.flow_control != :auto ->
```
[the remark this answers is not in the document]
FelonEkonom: Because we will resolve `effective_flow_control` on `handle_playing` anyway

lib/membrane/core/element.ex
```diff
@@ -215,6 +216,20 @@ defmodule Membrane.Core.Element do
     {:noreply, state}
   end
 
+  defp do_handle_info(
+         Message.new(:other_effective_flow_control_resolved, [my_pad_ref, effective_flow_control]),
+         state
+       ) do
+    state =
+      EffectiveFlowController.handle_other_effective_flow_control(
```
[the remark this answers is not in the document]
FelonEkonom: Do you have any other ideas for this function name? `handle_other_effective_flow_control_resolved` is quite long :(

lib/membrane/core/element/demand_counter.ex
FelonEkonom: I have put every module in the separate file, inside `core/element/demand_counter` directory

test/membrane/core/element/demand_counter_test.exs
```diff
@@ -0,0 +1,134 @@
...
+      end)
+
+    :ok = DemandCounter.set_sender_mode(demand_counter, :push)
+    :ok = DemandCounter.set_receiver_mode(demand_counter, :pull)
+    _demand_counter = DemandCounter.decrease(demand_counter, 1000)
+
+    assert_receive {:DOWN, ^monitor_ref, :process, _pid, _reason}
+  end
+
+  test "if buffering decrementation works properly with distribution" do
```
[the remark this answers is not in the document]
FelonEkonom: I wouldn't say that, in this test I check if buffering decrementation mechanism works properly when `sender_process` is on different node, than process that have created `DemandCounter` instance, in this case that is test process

lib/membrane/core/element/demand_controller.ex
FelonEkonom: I haven't fixed this bug yet, but I will do it before merging this PR

lib/membrane/core/element/input_queue.ex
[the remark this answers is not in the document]
varsill: but `start`/`end_of_stream` are both implementing `Membrane.EventProtocol` and therefore they are `Event.t`

lib/membrane/core/element/demand_handler.ex
mat-hek: not true https://github.com/elixir-lang/elixir/blob/cbb4fd2b3d55487b43b370bba30ed04c379a59c0/lib/elixir/lib/map_set.ex#L412

lib/membrane/core/element/demand_controller/auto_flow_utils.ex
```diff
@@ -0,0 +1,78 @@
...
+  alias Membrane.Core.Element.{
+    DemandCounter,
+    State
+  }
+
+  require Membrane.Core.Child.PadModel, as: PadModel
+  require Membrane.Pad, as: Pad
+
+  @lacking_buffer_size_lowerbound 200
+  @lacking_buffer_size_upperbound 400
```
mat-hek, line 13: `auto_demand_size` doesn't seem to affect these

mat-hek (changes requested):
TBC :D

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    state = %{state | effective_flow_control: new_effective_flow_control}
+
+    Enum.each(state.pads_data, fn
+      {_ref, %{flow_control: :auto, direction: :output} = pad_data} ->
+        :ok = DemandCounter.set_sender_mode(pad_data.demand_counter, new_effective_flow_control)
+        :ok = DemandCounter.set_receiver_mode(pad_data.demand_counter, :to_be_resolved)
+
+        Message.send(pad_data.pid, :other_effective_flow_control_resolved, [
```
varsill, line 121: Are you sure that one is correct? You have set the sender mode for the `output` pad (of the sender) and then sent a message to the sender that the `other_effective_flow_control_resolved` and it's not `other_effective_flow_control` that has been resolved.

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,136 @@
...
+    end
+  end
+
+  @spec handle_other_effective_flow_control(
+          Pad.ref(),
+          effective_flow_control(),
+          State.t()
+        ) ::
+          State.t()
+  def handle_other_effective_flow_control(my_pad_ref, other_effective_flow_control, state) do
```
varsill, line 62: Do we need to stick to the `other_*` nomenclature, like with `other_effective_flow_control`? I believe `handle_other_effective_flow_control` is always called for the `my_pad_ref ` being the `input` pad. Changing `other_effective_flow_control` to `output_effective_flow_control` could help make the code more clear, as the reader wouldn't need to search the context to find out if something is an input pad or an output pad. I think there are more changes of that type that can be done after you have simplified the links resolving.

varsill (commented):
TBC

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -0,0 +1,136 @@
...
+  @spec handle_input_pad_added(Pad.ref(), State.t()) :: State.t()
+  def handle_input_pad_added(pad_ref, state) do
+    with %{pads_data: %{^pad_ref => %{flow_control: :auto, direction: :input} = pad_data}} <-
+           state do
+      handle_other_effective_flow_control(
+        pad_ref,
+        pad_data.other_effective_flow_control,
+        state
+      )
+    end
```
mat-hek, line 53: I'd rather have it in the pad controller

lib/membrane/core/element/input_queue.ex
```diff
@@ -291,47 +273,43 @@ defmodule Membrane.Core.Element.InputQueue do
...
-    Sending demand of size #{inspect(to_demand)} to output #{inspect(linked_output_ref)}
+    Increasing DemandCounter linked to  #{inspect(input_queue.linked_output_ref)} by #{inspect(diff)}
     """
     |> mk_log(input_queue)
     |> Membrane.Logger.debug_verbose()
 
-    Message.send(demand_pid, :demand, to_demand, for_pad: linked_output_ref)
-    %__MODULE__{input_queue | demand: demand - to_demand}
-  end
+    lacking_buffer_size = lacking_buffer_size + diff
```
mat-hek, line 294:
```suggestion
```

lib/membrane/core/element/input_queue.ex
```diff
@@ -291,47 +273,43 @@ defmodule Membrane.Core.Element.InputQueue do
...
+    :ok = DemandCounter.increase(demand_counter, diff)
 
-  defp send_demands(input_queue, _demand_pid, _linked_output_ref) do
-    input_queue
+    %{input_queue | lacking_buffer_size: lacking_buffer_size}
```
mat-hek, line 297:
```suggestion
    %{input_queue | lacking_buffer_size: lacking_buffer_size + diff}
```

mat-hek (changes requested):

lib/membrane/core/element/effective_flow_controller.ex
varsill: After discussion we have decided to postpone it for a future refactor

lib/membrane/core/element/effective_flow_controller.ex
varsill: Apparently `pad_data.pid` is a pid of the other side of the link - I think we should rename it to i.e. `other_pid`, but let's do it later

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -234,7 +232,7 @@ defmodule Membrane.Core.Bin.PadController do
 
       child_endpoint = %{child_endpoint | pad_props: pad_props}
 
-      if params.initiator == :sibling do
+      if direction == :input do
```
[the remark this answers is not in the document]
FelonEkonom: This line does not change anything. I will remove field `initiator` from `params` in the another PR, because it have become redundant, since we always do `handle_link` calls in order parent -> output sibling -> input sibling

lib/membrane/core/element/effective_flow_controller.ex
FelonEkonom: This message is sent to the receiver, not sender

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,213 @@
...
+  @spec set_sender_mode(t, EffectiveFlowController.effective_flow_control()) :: :ok
+  def set_sender_mode(%__MODULE__{} = demand_counter, mode) do
+    DistributedFlowMode.put(
+      demand_counter.sender_mode,
+      mode
+    )
+  end
+
+  @spec get_sender_mode(t) :: flow_mode()
+  def get_sender_mode(%__MODULE__{} = demand_counter) do
```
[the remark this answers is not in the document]
FelonEkonom: it is not really flow control, because this function can also return `:to_be_resolved`

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,213 @@
...
+    DistributedFlowMode.get(demand_counter.sender_mode)
+  end
+
+  @spec set_receiver_mode(t, flow_mode()) :: :ok
+  def set_receiver_mode(%__MODULE__{} = demand_counter, mode) do
```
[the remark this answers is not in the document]
FelonEkonom: as above

lib/membrane/core/element/demand_counter.ex
```diff
@@ -0,0 +1,213 @@
...
+        toilet_capacity \\ nil,
+        throttling_factor \\ nil
+      ) do
+    %DistributedAtomic{worker: worker} = counter = DistributedAtomic.new()
+
+    throttling_factor =
+      cond do
+        throttling_factor != nil -> throttling_factor
+        node(sender_process) == node(worker) -> @default_throttling_factor
+        true -> @distributed_default_throttling_factor
```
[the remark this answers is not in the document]
mat-hek: but do we need it?

mat-hek (changes requested):

lib/membrane/core/element/demand_counter.ex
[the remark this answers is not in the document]
FelonEkonom: Or we will have `@default_throttling_factor 1`, either we will have to check on every `DemandCounter.decrease`, if the sender is on this same node as the receiver.

lib/membrane/core/element/demand_controller.ex
```diff
@@ -91,7 +89,9 @@ defmodule Membrane.Core.Element.DemandController do
   @spec decrease_demand_by_outgoing_buffers(Pad.ref(), [Buffer.t()], State.t()) :: State.t()
   def decrease_demand_by_outgoing_buffers(pad_ref, buffers, state) do
     pad_data = PadModel.get_data!(state, pad_ref)
-    buffers_size = Buffer.Metric.from_unit(pad_data.other_demand_unit).buffers_size(buffers)
+
+    demand_unit = pad_data.demand_unit || pad_data.other_demand_unit || :buffers
```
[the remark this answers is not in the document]
FelonEkonom: Yes, but this alternative matters when output pad or both sides of link has `:push` flow control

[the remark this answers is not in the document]
mat-hek: So maybe we should calculate the demand unit no matter the flow control (since the demand counter is always there)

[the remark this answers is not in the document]
mat-hek: > IMO it does not look good to have non-nil demand_unit when flow_control is `:push`.

IMO it looks even worse to have demand_counter and not have demand_unit there 🤔 And I see no point in resolving it each time when we can do it once.

Maybe the field should be `buffer_size_unit`, not `demand_unit` :D

lib/membrane/core/element/demand_counter/distributed_atomic.ex
```diff
@@ -19,7 +19,7 @@ defmodule Membrane.Core.Element.DemandCounter.DistributedAtomic do
   @spec new(integer() | nil) :: t
   def new(initial_value \\ nil) do
     atomic_ref = :atomics.new(1, [])
-    {:ok, worker} = Worker.start_link()
+    {:ok, worker} = Worker.start_link(self())
```
mat-hek, line 22: let's start it under the supervisor

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -85,16 +85,17 @@ defmodule Membrane.Core.Element.EffectiveFlowController do
...
 
   defp set_effective_flow_control(
          effective_flow_control,
+         _last_changed_pad,
          %{effective_flow_control: effective_flow_control} = state
        ),
        do: state
 
-  defp set_effective_flow_control(new_effective_flow_control, state) do
+  defp set_effective_flow_control(new_effective_flow_control, last_changed_pad, state) do
```
mat-hek, line 98:
```suggestion
  defp set_effective_flow_control(new_effective_flow_control, triggerer_pad, state) do
```

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -105,19 +106,41 @@ defmodule Membrane.Core.Element.EffectiveFlowController do
...
         Message.send(
           pad_data.pid,
-          :other_effective_flow_control_resolved,
+          :sender_effective_flow_control_resolved,
           [pad_data.other_ref, new_effective_flow_control]
         )
 
         state
 
+      {pad_ref, %{direction: :input} = pad_data}, state when last_changed_pad != nil ->
```
mat-hek, line 125:
```suggestion
      {pad_ref, %{direction: :input} = pad_data}, state ->
```

mat-hek (changes requested):

lib/membrane/core/element/effective_flow_controller.ex
```diff
@@ -6,8 +6,8 @@ defmodule Membrane.Core.Element.EffectiveFlowController do
   # :pull mode. If the element's effective flow control is set to :push, then all of its auto pads work in :push. Analogically,
   # if the element effective flow control is set to :pull, auto pads also work in :pull.
 
-  # If element A is linked via its input auto pads only to the :push output pads, then effective flow control of
-  # element A will be set to :push. Otherwise, if element A is linked via its input auto pads to at least one
+  # If element A is linked via its input auto pads only to the :push output pads of other elements, then effective flow
+  # control of element A will be set to :push. Otherwise, if element A is linked via its input auto pads to at least one
   # :pull output pad, element A will set itss effective flow control to :pull and will forward this information
```
varsill, line 11:
```suggestion
  # :pull output pad, element A will set its effective flow control to :pull and will forward this information
```

FelonEkonom:
> 🎉

awesome

lib/membrane/core/element/demand_controller.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Membrane.Core.Element.DemandController do
   @moduledoc false
 
-  # Module handling changes in values of output pads demand counters
+  # Module handling changes in values of output pads demand atomics
```
varsill, line 4:
```suggestion
  # Module handling changes in values of output pads atomic demand
```
To be fair, I believe it should be written that way, but I don't find it anyhow crucial :D

varsill (approved):
There is one small comment concerning the docs, but generally speaking:
LGTM 🥇

PR #538, opened 2023-03-15
Fix the bug with pipeline persisting in Zombie mode after self-terminating with `terminate: :normal` action

varsill:
Some additional small changes (removing unused variables and aliases from tests) have also been performed since my credo has been failing.

### DIFF
--- a/lib/membrane/core/pipeline/action_handler.ex
+++ b/lib/membrane/core/pipeline/action_handler.ex
@@ -95,8 +95,6 @@ defmodule Membrane.Core.Pipeline.ActionHandler do
 
   @impl CallbackHandler
   def handle_action({:terminate, :normal}, _cb, _params, state) do
-    state = %{state | terminating?: true}
-
     case LifecycleController.handle_terminate(state) do
       {:continue, state} -> state
       {:stop, _state} -> exit(:normal)
--- a/lib/membrane/core/pipeline/action_handler.ex
+++ b/lib/membrane/core/pipeline/action_handler.ex
@@ -95,6 +95,8 @@ defmodule Membrane.Core.Pipeline.ActionHandler do
 
   @impl CallbackHandler
   def handle_action({:terminate, :normal}, _cb, _params, state) do
+    state = %{state | terminating?: true}
+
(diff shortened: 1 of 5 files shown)


test/membrane/core/pipeline_test.exs
```diff
@@ -129,6 +129,16 @@ defmodule Membrane.Core.PipelineTest do
     end)
   end
 
+  test "Pipeline should be able to terminate itself with terminate: :normal action when it has already spawned children" do
+    {:ok, _supervisor, pid} = Testing.Pipeline.start(module: TestPipeline)
```
FelonEkonom, line 133:
```suggestion
    pid = Testing.Pipeline.start_supervised!(module: TestPipeline)
```

lib/membrane/core/pipeline/action_handler.ex
```diff
@@ -95,6 +95,8 @@ defmodule Membrane.Core.Pipeline.ActionHandler do
 
   @impl CallbackHandler
   def handle_action({:terminate, :normal}, _cb, _params, state) do
+    state = %{state | terminating?: true}
```
FelonEkonom, line 98: callbacks above don't change state, so maybe move this line to `LifecycleController.handle_terminate/1`?

varsill: Well, the only usage of `Parent.LifecycleController.handle_terminate/1` is here, so we can safely move it ;)

test/membrane/element_test.exs
```diff
@@ -75,16 +75,16 @@ defmodule Membrane.ElementTest do
...
   end
 
-  test "play", %{pipeline: pipeline} do
+  test "play", %{} do
     TestFilter.assert_callback_called(:handle_playing)
   end
 
   describe "Start of stream" do
-    test "causes handle_start_of_stream/3 to be called", %{pipeline: pipeline} do
+    test "causes handle_start_of_stream/3 to be called", %{} do
```
mat-hek, line 83:
```suggestion
    test "causes handle_start_of_stream/3 to be called" do
```

mat-hek (approved):

PR #539, opened 2023-03-15
Make sure enumerable with all elements being `Membrane.Buffer.t()`, passed as `:output` parameter for `Membrane.Testing.Source` won't get rewrapper in `Membrane.Buffer.t()` struct.

varsill:


FelonEkonom (changes requested):

FelonEkonom:
IMO it would be better, to put element of `output` into a `%Membrane.Buffer{}` if this single element is not a `%Membrane.Buffer{}`, without looking at the rest of the `output`, to avoid scenario, when we have `%Membrane.Buffer{}` inside another `%Membrane.Buffer{}`

lib/membrane/testing/source.ex
```diff
@@ -74,8 +77,9 @@ defmodule Membrane.Testing.Source do
       {initial_state, generator} when is_function(generator) ->
         {[], opts |> Map.merge(%{generator_state: initial_state, output: generator})}
 
-      _enumerable_output ->
-        {[], opts}
+      enumerable_output ->
+        all_buffers_in_output = Enum.all?(enumerable_output, &match?(%Membrane.Buffer{}, &1))
+        {[], opts |> Map.merge(%{all_buffers_in_output?: all_buffers_in_output})}
```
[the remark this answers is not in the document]
varsill: done

lib/membrane/testing/source.ex
```diff
@@ -124,15 +128,21 @@ defmodule Membrane.Testing.Source do
...
-    {payloads, output} = Enum.split(output, size)
-    buffers = Enum.map(payloads, &%Buffer{payload: &1})
+    {to_output_now, rest} = Enum.split(output, size)
+
+    buffers =
+      if state.all_buffers_in_output? do
+        to_output_now
+      else
+        Enum.map(to_output_now, &%Buffer{payload: &1})
+      end
```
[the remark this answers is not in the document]
varsill: I am not sure it looks anyhow nicer:
```
buffers = Bunch.then_if(to_output_now, state.all_buffers_in_output?, fn unwrapped_buffers -> Enum.map(unwrapped_buffers, &%Buffer{payload: &1}) end)
```
`then_if` is helpful when we are having a pipeline and here there is no need for a pipeline. What is more, I would need to use nested anonymous functions, that prevents me from using the `&` syntax

mat-hek:
> IMO it would be better, to put element of output into a %Membrane.Buffer{} if this single element is not a %Membrane.Buffer{}, without looking at the rest of the output, to avoid scenario, when we have %Membrane.Buffer{} inside another %Membrane.Buffer{}

I think I agree it's best just to wrap each term that's not a buffer into a buffer

lib/membrane/testing/source.ex
```diff
@@ -40,14 +40,16 @@ defmodule Membrane.Testing.Source do
...
                 default: {0, &__MODULE__.default_buf_gen/2},
                 description: """
-                If `output` is an enumerable with `t:Membrane.Payload.t/0` then
-                buffer containing those payloads will be sent through the
-                `:output` pad and followed by `t:Membrane.Element.Action.end_of_stream/0`.
+                If `output` is an enumerable with elements being `t:Membrane.Payload.t/0`,
+                then these elements that are not `t:Membrane.Buffer.t/0` structs,
+                will be put into `t:Membrane.Buffer.t/0` struct's `payload` field and such a list of
+                `t:Membrane.Buffer.t/0` buffers will be sent via `:output` pad, followed by
+                `t:Membrane.Element.Action.end_of_stream/0`.
```
mat-hek, line 50: Can you try to write that in a simpler way?

varsill: done

lib/membrane/testing/source.ex
```diff
@@ -40,14 +40,16 @@ defmodule Membrane.Testing.Source do
...
                 default: {0, &__MODULE__.default_buf_gen/2},
                 description: """
-                If `output` is an enumerable with `t:Membrane.Payload.t/0` then
-                buffer containing those payloads will be sent through the
-                `:output` pad and followed by `t:Membrane.Element.Action.end_of_stream/0`.
+                If the `output` is an enumerable of elements that are of type `t:Membrane.Payload.t/0`,
+                then any of those elements that are not already `t:Membrane.Buffer.t/0` structures will
+                be enclosed in a `t:Membrane.Buffer.t/0` structure, and a list of these `t:Membrane.Buffer.t/0`
+                structures will be sent through the `:output` pad, followed by
+                `t:Membrane.Element.Action.end_of_stream/0` action.
```
mat-hek, line 50:
```suggestion
                If `output` is an Enumerable, then each element of it will be sent
                through the `:output` pad, followed by `t:Membrane.Element.Action.end_of_stream/0`.
                Each element of the Enumerable must be either `t:Membrane.Buffer.t/0`
                or `t:Membrane.Payload.t/0`. In the latter case, it will be automatically wrapped into
                `t:Membrane.Buffer.t/0` before sending.
```

mat-hek (approved):

PR #540, opened 2023-03-20
Improve performance test

varsill:
This PR improves the formatting of the metric printouts and solves the problem with final_memory metric


.circleci/config.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - run: cp -r benchmark/ ~/benchmark_backup/
       - run: cp mix.exs ~/benchmark_backup/
       - run: docker run -e MIX_ENV=benchmark -v ./:/root/app -v ~/results:/root/results -w /root/app membraneframeworklabs/docker_membrane mix run benchmark/run.exs /root/results/results.res
-      - run: git checkout master
+      - run: git checkout -f master
```
FelonEkonom, line 159: why?

benchmark/compare.exs
```diff
@@ -20,17 +20,17 @@ defmodule Benchmark.Compare do
...
           """
           METRIC: #{metric_module}
-          #{inspect(Map.get(test_case_results, metric_module), pretty: true)}
-          vs
-          #{inspect(Map.get(test_case_results_ref, metric_module), pretty: true)}
-
+          used to be:
+          #{inspect(Map.get(test_case_results_ref, metric_module), pretty: true, limit: :infinity)}
+          now is:
+          #{inspect(Map.get(test_case_results, metric_module), pretty: true, limit: :infinity)}
```
FelonEkonom, line 26: What about having `master branch` and `feature branch` labels, instead of `used to be` and `now is`? To be precise, we are not comparing `feature branch` `HEAD`  with `feature branch` `HEAD~10`, but with current `master`, that does not have to be behind `feature branch` `HEAD` in commits history

FelonEkonom (changes requested):

.circleci/config.yml
varsill: Because on this feature branch I have needed to delete a file which still exists on master (`benchmark/metrics/final_memory.ex`)

mat-hek (approved):

FelonEkonom (approved):

PR #545, opened 2023-04-18
Fix bug in make_cannonical, make Child.ref a macro

FelonEkonom:
Related Jira tickets:
 - https://membraneframework.atlassian.net/browse/MC-169
 - https://membraneframework.atlassian.net/browse/MC-170

### DIFF
--- a/lib/membrane/child.ex
+++ b/lib/membrane/child.ex
@@ -29,8 +29,16 @@ defmodule Membrane.Child do
 
   defmacro ref(name, options) do
     case __CALLER__.context do
-      :match -> ref_in_a_match(name, options)
-      _not_match -> ref_outside_a_match(name, options)
+      :match ->
+        Macro.expand(options, __CALLER__)
+        |> Access.fetch(:group)
+        |> case do
+          {:ok, group_ast} -> ref_in_a_match(name, group_ast)
+          :error -> raise "Improper options. The options must be in form of [group: group]."
+        end
+
+      _not_match ->
+        ref_outside_a_match(name, options)
     end
   end
 
@@ -40,7 +48,7 @@ defmodule Membrane.Child do
     end
   end
 
-  defp ref_in_a_match(name, group: group) do
+  defp ref_in_a_match(name, group) do
     quote do
       {
         unquote(__MODULE__),
--- a/lib/membrane/child.ex
+++ b/lib/membrane/child.ex
@@ -6,7 +6,6 @@ defmodule Membrane.Child do
   import Membrane.Bin, only: [is_bin_name?: 1]
 
   alias Membrane.{Bin, Element}
-  @default_ref_options [group: [default: nil]]
 
   @type name :: Element.name() | Bin.name()
   @type group() :: any()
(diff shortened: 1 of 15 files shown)


lib/membrane/child.ex
```diff
@@ -21,22 +20,68 @@ defmodule Membrane.Child do
...
+    end
+  end
+
+  defp ref_in_a_match(name) do
+    quote do
+      unquote(name)
+    end
+  end
+
+  defp ref_in_a_match(name, group: group) do
```
varsill, line 43: Right now, when the user passes an improper list of options to the `Child.ref` in `match` context, an error saying that the "`ref_in_a_match` function couldn't be matched" is raised. How about adding the following function clause:
```
 defp ref_in_a_match(_name, improper_options) do
    raise "Improper options: #{inspect(improper_options)}. The options must be in form of [group: group]."
  end
```

varsill: Perhaps I am running into the future to much, but I also wonder if we should allow the user to match on the whole options list, that is - to fetch the `opts` from the following match case: `Child.ref(child_name, opts) = child`.

lib/membrane/child.ex
```diff
@@ -21,22 +20,68 @@ defmodule Membrane.Child do
...
+  defmacro ref(name, options) do
+    case __CALLER__.context do
+      :match -> ref_in_a_match(name, options)
+      _not_match -> ref_outside_a_match(name, options)
+    end
+  end
+
+  defp ref_in_a_match(name) do
+    quote do
+      unquote(name)
```
varsill, line 39: why would we want to allow for such a match: `Child.ref({Membrane.Child, _, _}) = {Membrane.Child, :group, :name}`?

varsill (changes requested):

test/membrane/integration/accepted_format_test.exs
```diff
@@ -89,6 +89,7 @@ defmodule Membrane.StreamFormatTest do
   defp assert_down(module) do
     assert_receive({:my_pid, ^module, pid})
     Process.monitor(pid)
-    assert_receive({:DOWN, _ref, :process, ^pid, {%Membrane.StreamFormatError{}, _stacktrace}})
+    assert_receive({:DOWN, _ref, :process, ^pid, reason})
+    assert match?({%Membrane.StreamFormatError{}, _stacktrace}, reason) or reason == :noproc
```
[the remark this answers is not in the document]
FelonEkonom: When `pid` dies before `Process.monitor`

lib/membrane/element/base.ex
```diff
@@ -243,13 +243,20 @@ defmodule Membrane.Element.Base do
 
   Options:
     - `:bring_pad?` - if true (default) requires and aliases `Membrane.Pad`
+    - `:bring_child?` - if true (default) requires and aliases `Membrane.Child`
```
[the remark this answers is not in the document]
FelonEkonom: Element still can use it to match on it's own name

lib/membrane/child.ex
FelonEkonom: I don't, but I cannot check in inside a match :shit:

varsill (approved):

test/membrane/integration/accepted_format_test.exs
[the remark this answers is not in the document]
mat-hek: This approach smells, if the reason is `:noproc` then the element could have died for any reason or even never started. The appropriate way would be to monitor the process when we're sure it should be still alive or use start_link

mat-hek (approved):

mat-hek (approved):

PR #551, opened 2023-05-08
Add typedocs for @type t types

varsill:


lib/membrane/time.ex
```diff
@@ -16,6 +16,9 @@ defmodule Membrane.Time do
   @compile {:inline,
             native_units: 1, native_unit: 0, nanoseconds: 1, nanosecond: 0, second: 0, seconds: 1}
 
+  @typedoc """
+  The time represented in Membrane's native time units - nanoseconds.
```
mat-hek, line 20: the underlying unit technically shouldn't be relied on

lib/membrane/time.ex
```diff
@@ -17,7 +17,7 @@ defmodule Membrane.Time do
             native_units: 1, native_unit: 0, nanoseconds: 1, nanosecond: 0, second: 0, seconds: 1}
 
   @typedoc """
-  The time represented in Membrane's native time units - nanoseconds.
+  The time represented in Membrane's native time units.
```
mat-hek, line 20: actually the word 'native' usually relates to OS in this context, so I'd remove it as well :P
```suggestion
  The time represented in Membrane's time units.
```

mat-hek (approved):

PR #552, opened 2023-05-08
Implement Debug.Filter and Debug.Sink

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-174


test/membrane/integration/simple_elements_test.exs
```diff
@@ -0,0 +1,85 @@
...
+    @impl true
+    def handle_parent_notification({:send_buffers, number}, _ctx, state) do
+      buffers =
+        Enum.map(1..number, fn i ->
+          %Buffer{payload: inspect(i)}
+        end)
+
+      {[buffer: {:output, buffers}], state}
+    end
+  end
```
mat-hek, line 34: Does this do anything beyond the testing source?

mat-hek (changes requested):

lib/membrane/simple/filter.ex
```diff
@@ -0,0 +1,68 @@
...
+  alias Membrane.StreamFormat
+
+  def_input_pad :input, accepted_format: _any, flow_control: :auto
+  def_output_pad :output, accepted_format: _any, flow_control: :auto
+
+  @spec identity(any()) :: any()
+  def identity(arg), do: arg
+
+  def_options handle_buffer: [spec: (Buffer.t() -> Buffer.t()), default: &__MODULE__.identity/1],
+              handle_event: [spec: (Event.t() -> Event.t()), default: &__MODULE__.identity/1],
```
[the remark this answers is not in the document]
mat-hek: BTW, you can use ` &Function.identity/1

test/membrane/integration/simple_elements_test.exs
FelonEkonom: This source works in `push`, while Testing Source works in `pull`

mat-hek: does it matter in this use case?

mat-hek: It wouldn't be more than you provide via the `output` option

lib/membrane/lambda_filter.ex
```diff
@@ -1,16 +1,11 @@
-defmodule Membrane.Simple.Filter do
+defmodule Membrane.LambdaFilter do
```
mat-hek, line 1: ❤️

lib/membrane/lambda_filter.ex
```diff
@@ -27,14 +22,30 @@ defmodule Membrane.Simple.Filter do
...
+                Function with arity 1, that maps buffers handled by this filter.
+                Result of this function is passed to the `:buffer` action on `:output` pad.
+                """
+              ],
+              handle_event: [
+                spec: (Event.t() -> Event.t() | [Event.t()]),
+                default: &Function.identity/1,
+                description: """
+                Function with arity 1, that maps events handled by this filter.
+                Result of this function is passed to the `:event` action on the opposite pad to
```
mat-hek, line 38: hmm, this way we cannot know whether the event is going downstream or upstream, which is pretty important

lib/membrane/debug/filter.ex
```diff
@@ -0,0 +1,84 @@
...
+
+  @impl true
+  def handle_event(pad, event, _ctx, state) do
+    _ingored = state.handle_event.(event)
+
+    opposite_pad =
+      case pad do
+        :input -> :output
+        :output -> :input
+      end
```
mat-hek, line 74: you can use the forward action instead

lib/membrane/debug/filter.ex
```diff
@@ -0,0 +1,84 @@
+defmodule Membrane.Debug.Filter do
+  @moduledoc """
+  Membrane Filter, that can be used to create a child that will be used to debug data flowing thouth pipeline.
```
mat-hek, line 3:
```suggestion
  Membrane Filter, that can be used to create a child that will be used to debug data flowing thouth pipeline.
  
```

mat-hek (approved):

PR #553, opened 2023-05-10
Raise on attempting to remove a not existing child

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-176

### DIFF
--- a/lib/membrane/core/parent/child_life_controller.ex
+++ b/lib/membrane/core/parent/child_life_controller.ex
@@ -541,16 +541,17 @@ defmodule Membrane.Core.Parent.ChildLifeController do
       MapSet.new(state.children, fn {_ref, data} -> data.group end)
       |> MapSet.delete(nil)
 
-    Enum.find(removed_children_or_groups, fn name ->
-      not Map.has_key?(state.children, name) and not MapSet.member?(children_groups, name)
+    removed_children_or_groups
+    |> Enum.reject(fn name ->
+      Map.has_key?(state.children, name) or MapSet.member?(children_groups, name)
     end)
     |> case do
-      nil ->
+      [] ->
         :ok
 
-      child_ref ->
+      children_refs ->
         raise Membrane.ParentError, """
-        Trying to remove child #{inspect(child_ref)}, while such a child or children group does not exist.
+        Trying to remove children #{Enum.map_join(children_refs, ", ", &inspect/1)}, while such children or children groups do not exist.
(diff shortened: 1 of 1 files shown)


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -500,6 +500,25 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
     children_or_children_groups = Bunch.listify(children_or_children_groups)
 
+    all_children_and_children_groups =
+      state.children
+      |> Enum.flat_map(fn {ref, %{group: group}} -> [ref, group] end)
+      |> List.delete(nil)
+      |> Enum.uniq()
+
+    children_or_children_groups
+    |> Enum.find(&(&1 not in all_children_and_children_groups))
```
mat-hek, line 510: instead of searching through all children, we could just search in the ones that are to be removed

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
FelonEkonom: value `refs` in line 522 contains only references to the children that we have to remove, while variable `children_or_children_groups` might contain not only children refs, but child groups names as well, so I am not convinced if this change will make the code simpler

mat-hek: but we can easily get the groups when retrieving refs

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -553,6 +536,27 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    Enum.find(removed_children_or_groups, fn name ->
+      not Map.has_key?(state.children, name) and not MapSet.member?(children_groups, name)
+    end)
+    |> case do
+      nil ->
+        :ok
+
+      child_ref ->
+        raise Membrane.ParentError, """
+        Trying to remove child #{inspect(child_ref)}, while such a child or children group does not exist.
```
mat-hek, line 553:
```suggestion
    removed_children_or_groups
    |> Enum.reject(fn name ->
      Map.has_key?(state.children, name) or MapSet.member?(children_groups, name)
    end)
    |> case do
      [] ->
        :ok

      children_refs ->
        raise Membrane.ParentError, """
        Trying to remove children #{Enum.map_join(", ", &inspect/1)}, while such children or children groups do not exist.
```

mat-hek (approved):

PR #568, opened 2023-06-15
Stop calling handle_demand after end of stream

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-204

### DIFF
--- a/test/membrane/integration/demands_test.exs
+++ b/test/membrane/integration/demands_test.exs
@@ -6,6 +6,7 @@ defmodule Membrane.Integration.DemandsTest do
   import Membrane.ChildrenSpec
   import Membrane.Testing.Assertions
 
+  alias Membrane.Testing
   alias Membrane.Buffer
   alias Membrane.Support.DemandsTest.Filter
   alias Membrane.Testing.{Pipeline, Sink, Source}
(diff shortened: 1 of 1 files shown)


lib/membrane/core/element/action_handler.ex
```diff
@@ -47,45 +47,56 @@ defmodule Membrane.Core.Element.ActionHandler do
   defguardp is_demand_size(size) when is_integer(size) or is_function(size)
 
   @impl CallbackHandler
-  def handle_action({action, _}, :handle_init, _params, _state)
-      when action not in [:latency, :notify_parent] do
-    raise ActionError, action: action, reason: {:invalid_callback, :handle_init}
+  def handle_action(action, callback, params, state) do
+    state = Map.update!(state, :callback_depth_counter, &(&1 + 1))
```
mat-hek, line 51: instead of that, I'd set a `handling_callback?` flag in the callback handler

test/membrane/core/element/action_handler_test.exs
```diff
@@ -25,6 +25,9 @@ defmodule Membrane.Core.Element.ActionHandlerTest do
         playback: :stopped,
         synchronization: %{clock: nil, parent_clock: nil},
         delayed_demands: MapSet.new(),
+        callback_depth_counter: 0,
+        pads_to_snapshot: MapSet.new(),
+        pads_to_snapshot: MapSet.new(),
```
mat-hek:
```suggestion
```

mat-hek (changes requested):

mat-hek (approved):

PR #574, opened 2023-06-30
Refactor code related to crash groups and linking mechanism

FelonEkonom:
Related Jira ticekt: https://membraneframework.atlassian.net/browse/MC-191


lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -2,86 +2,141 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
...
-      state = Bunch.Access.delete_in(state, [:crash_groups, group_name])
+  def handle_crash_group_member_death(
+        child_name,
+        %CrashGroup{triggered?: true} = group,
+        _reason,
+        state
+      ) do
+    all_members_dead? =
+      Enum.all?(group.members, fn member ->
+        member == child_name or not Map.has_key?(state.children, member)
```
mat-hek, line 51: if a child is not in `state.children`, shouldn't it have been removed from `group.members` already?

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -2,86 +2,141 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
...
-    %CrashGroup{alive_members_pids: alive_members_pids} = state.crash_groups[group_name]
+  @spec get_child_crash_group(Child.name(), Parent.state()) :: {:ok, CrashGroup.t()} | :error
+  def get_child_crash_group(child_name, state) do
+    %{group: group_name} = ChildrenModel.get_child_data!(state, child_name)
+    Map.fetch(state.crash_groups, group_name)
+  end
+
+  @spec handle_crash_group_member_death(Child.name(), CrashGroup.t(), any(), Parent.state()) ::
+          Parent.state()
+  def handle_crash_group_member_death(child_name, crash_group_data, reason, state)
```
mat-hek, line 41: this has so many clauses, can we make it a bit less verbose?

lib/membrane/core/element/pad_controller.ex
```diff
@@ -359,21 +360,28 @@ defmodule Membrane.Core.Element.PadController do
...
   end
 
-  defp init_pad_direction_data(%{direction: :input}, _props, metadata, _state),
+  defp merge_pad_direction_data(pad_data, metadata, state) do
+    merge_pad_data(pad_data, &init_pad_direction_data(&1, metadata, state))
+  end
+
+  defp merge_pad_mode_data(pad_data, props, other_info, state) do
+    merge_pad_data(pad_data, &init_pad_mode_data(&1, props, other_info, state))
+  end
```
mat-hek, line 369: These add another level of complexity, which is one level too much IMO. If you don't like passing a function to `merge_pad_data`, I'd rather remove it and rename each `init_pad_*_data` function to `merge_pad_*_data` and explicitly call `Map.merge` there.

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
FelonEkonom: When a child dies with a reason different than `:normal`, then it lasts in `group.members`, until the whole group dies

lib/membrane/core/element/pad_controller.ex
```diff
@@ -137,17 +130,18 @@ defmodule Membrane.Core.Element.PadController do
     end
   end
 
-  defp do_handle_link(:input, endpoint, other_endpoint, info, link_props, state) do
+  defp do_handle_link(:input, endpoint, other_endpoint, pad_info, link_props, state) do
```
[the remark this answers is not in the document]
varsill: the same applies for the `do_handle_link(:output, ...)`

lib/membrane/core/element/pad_controller.ex
```diff
@@ -301,86 +297,88 @@ defmodule Membrane.Core.Element.PadController do
...
-        |> Enum.reduce(state, fn other_data, state ->
-          PadModel.update_data!(state, other_data.ref, :associated_pads, &[data.ref | &1])
+        |> Enum.filter(&(&1.direction != pad_data.direction and &1.flow_control == :auto))
+        |> Enum.reduce(state, fn opposite_pad_data, state ->
+          PadModel.update_data!(
+            state,
+            opposite_pad_data.ref,
+            :associated_pads,
+            &[pad_data.ref | &1]
+          )
```
varsill, line 343: [NIT] I would opt for moving that part of code to some separate function, e.g. `update_associated_pads` or something like that.

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -265,15 +264,14 @@ defmodule Membrane.Core.Parent.ChildLifeController do
     state = %{state | children: Map.merge(state.children, Map.new(children, &{&1.name, &1}))}
 
     children_names = children |> Enum.map(& &1.name)
-    children_pids = children |> Enum.map(& &1.pid)
 
     # adding crash group to state
     state =
-      if options.crash_group_mode != nil do
+      if options.crash_group_mode != nil or Map.has_key?(state.crash_groups, options.group) do
```
varsill, line 270: I am not sure I get it - shouldn't it be: `if options.crash_group_mode != nil and (not Map.has_key?(state.crash_groups, options.group))`
?

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -2,86 +2,117 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
...
+      |> Enum.all?(&(not Map.has_key?(state.children, &1)))
+
+    if all_members_dead? do
+      cleanup_crash_group(group.name, state)
+    else
+      state
+    end
+  end
+
+  defp trigger_crash_group(crash_initiator, %CrashGroup{} = group, state) do
```
varsill, line 71: I am not the biggest fan of the `trigger_crash_group` name - I would opt at least for `trigger_crash_group_crash`, but it might sound funny :D 
Perhaps you have some better alternative?

varsill: and when it comes to `triggered?` - I would rename it to `crash_triggered?` or something like that

varsill (changes requested):

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -2,86 +2,123 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
...
+  def get_child_crash_group(child_name, state) do
+    %{group: group_name} = ChildrenModel.get_child_data!(state, child_name)
+    Map.fetch(state.crash_groups, group_name)
+  end
+
+  @spec handle_crash_group_member_death(Child.name(), CrashGroup.t(), any(), Parent.state()) ::
+          Parent.state()
+  def handle_crash_group_member_death(child_name, %CrashGroup{} = group, :normal, state) do
+    members = List.delete(group.members, child_name)
+    state = put_in(state, [:crash_groups, group.name, :members], members)
```
mat-hek, line 49: I forgot again why we remove a member if it crashed with the reason `normal` and we don't otherwise. I think it should be refactored to make it more explicit or I'd add a comment.
BTW, I'd merge this and the subsequent clause anyway

mat-hek (changes requested):
Looks good 👍

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
FelonEkonom: Why would you merge these two caluses? IMO operations executed in these clauses are quite different and it is cleaner to keep them separate

mat-hek (approved):

test/membrane/core/element_test.exs
```diff
@@ -130,10 +130,10 @@ defmodule Membrane.Core.ElementTest do
           %Endpoint{pad_spec: :output, pad_ref: :output, pad_props: %{options: []}, child: :this},
           output_other_endpoint,
           %{
-            other_info: other_info,
+            # other_info: other_info,
```
varsill, line 133: ;)

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -30,6 +26,16 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
     )
   end
 
+  @spec extend_crash_group(
+          Child.group(),
+          ChildrenSpec.crash_group_mode(),
+          [Child.name()],
+          Parent.state()
+        ) :: Parent.state()
+  def extend_crash_group(group_name, _mode, children, state) do
```
varsill, line 35: I would remove `mode` argument here
```suggestion
  def extend_crash_group(group_name, children, state) do
```

test/membrane/core/element_test.exs
```diff
@@ -270,14 +270,14 @@ defmodule Membrane.Core.ElementTest do
                    pad_props: %{options: [], toilet_capacity: nil, throttling_factor: nil}
                  },
                  %{
-                   other_info: %{
-                     direction: :input,
+                   output_pad_info: %{
+                     direction: :output,
```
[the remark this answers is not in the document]
FelonEkonom: Without this fix, this test doesn't pass

varsill (approved):

PR #579, opened 2023-07-18
Improve `in_progress_memory` metric in performance benchmark

varsill:
This PR:
* Adds synchronous GC launch before the initial memory measurement in benchmark
* Removes calculation of the `initial_memory` in `in_progress_memory` metric

### DIFF
--- a/benchmark/metric/in_progress_memory.ex
+++ b/benchmark/metric/in_progress_memory.ex
@@ -33,7 +33,7 @@ defmodule Benchmark.Metric.InProgressMemory do
 
   @impl true
   def start_meassurement(_opts \\ nil) do
-    Process.list() |> Enum.each(:erlang.garbage_collect(&1))
+    Process.list() |> Enum.each(&:erlang.garbage_collect(&1))
     initial_memory = :erlang.memory(:total)
 
     task =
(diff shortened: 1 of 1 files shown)


benchmark/metric/in_progress_memory.ex
```diff
@@ -33,23 +33,23 @@ defmodule Benchmark.Metric.InProgressMemory do
...
+        do_loop([])
       end)
 
     task
   end
 
-  defp do_loop(acc, initial_memory) do
-    acc = acc ++ [:erlang.memory(:total) - initial_memory]
+  defp do_loop(acc) do
+    acc = acc ++ [:erlang.memory(:total)]
```
mat-hek, line 47: let's prepend & reverse

benchmark/metric/in_progress_memory.ex
```diff
@@ -33,23 +33,23 @@ defmodule Benchmark.Metric.InProgressMemory do
 
   @impl true
   def start_meassurement(_opts \\ nil) do
-    initial_memory = :erlang.memory(:total)
+    Process.list() |> Enum.each(&:erlang.garbage_collect(&1))
```
mat-hek, line 36:
```suggestion
    Process.list() |> Enum.each(&:erlang.garbage_collect/1)
```

mat-hek (approved):

PR #586, opened 2023-07-21
Implement soft corcking

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-213

### DIFF
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -137,43 +137,28 @@ defmodule Membrane.Core.Element.ActionHandler do
   end
 
   @impl CallbackHandler
-  def handle_action({:redemand, out_refs}, cb, params, state)
-      when is_list(out_refs) do
-    Enum.reduce(out_refs, state, fn out_ref, state ->
-      handle_action({:redemand, out_ref}, cb, params, state)
+  def handle_action({action, pads_refs}, cb, params, state)
+      when action in [:redemand, :pause_auto_demand, :resume_auto_demand] and is_list(pads_refs) do
+    Enum.reduce(pads_refs, state, fn pad_ref, state ->
+      handle_action({action, pad_ref}, cb, params, state)
     end)
   end
 
   @impl CallbackHandler
   def handle_action({:redemand, out_ref}, cb, _params, %State{type: type} = state)
-      when type in [:source, :filter, :endpoint] and is_pad_ref(out_ref) and
-             {type, cb} != {:filter, :handle_demand} do
+      when type in [:source, :filter, :endpoint] and {type, cb} != {:filter, :handle_demand} do
     handle_redemand(out_ref, state)
   end
 
   @impl CallbackHandler
-  def handle_action({:pause_auto_demand, in_refs}, cb, params, state) when is_list(in_refs) do
-    Enum.reduce(in_refs, state, fn in_ref, state ->
-      handle_action({:pause_auto_demand, in_ref}, cb, params, state)
-    end)
-  end
-
-  @impl CallbackHandler
   def handle_action({:pause_auto_demand, in_ref}, _cb, _params, %State{type: type} = state)
-      when type in [:sink, :filter, :endpoint] and is_pad_ref(in_ref) do
+      when type in [:sink, :filter, :endpoint] do
     DemandController.AutoFlowUtils.pause_demands(in_ref, state)
   end
 
   @impl CallbackHandler
-  def handle_action({:resume_auto_demand, in_refs}, cb, params, state) when is_list(in_refs) do
-    Enum.reduce(in_refs, state, fn in_ref, state ->
-      handle_action({:resume_auto_demand, in_ref}, cb, params, state)
-    end)
-  end
-
-  @impl CallbackHandler
   def handle_action({:resume_auto_demand, in_ref}, _cb, _params, %State{type: type} = state)
-      when type in [:sink, :filter, :endpoint] and is_pad_ref(in_ref) do
+      when type in [:sink, :filter, :endpoint] do
     DemandController.AutoFlowUtils.resume_demands(in_ref, state)
   end
 
--- a/lib/membrane/core/element/action_handler.ex
+++ b/lib/membrane/core/element/action_handler.ex
@@ -70,8 +70,8 @@ defmodule Membrane.Core.Element.ActionHandler do
                :stream_format,
                :demand,
                :redemand,
-               :pause_auto_demands,
-               :resume_auto_demands,
+               :pause_auto_demand,
+               :resume_auto_demand,
                :forward,
                :end_of_stream
              ] do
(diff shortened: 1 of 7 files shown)


test/membrane/integration/demands_test.exs
```diff
@@ -133,4 +133,82 @@ defmodule Membrane.Integration.DemandsTest do
...
+    def handle_demand(:output, _size, _unit, _ctx, state) do
+      Process.sleep(10)
+      {[buffer: {:output, %Membrane.Buffer{payload: ""}}, redemand: :output], state}
+    end
+  end
+
+  defmodule PausingSink do
+    use Membrane.Sink
+
+    def_input_pad :input, accepted_format: _any, flow_control: :auto, availability: :always
```
varsill, line 161: [NIT] I think that one is default ;)
```suggestion
    def_input_pad :input, accepted_format: _any, flow_control: :auto
```

test/membrane/integration/demands_test.exs
```diff
@@ -133,4 +133,82 @@ defmodule Membrane.Integration.DemandsTest do
...
+    end
+
+    @impl true
+    def handle_playing(_ctx, state) do
+      {[stream_format: {:output, %StreamFormat{}}], state}
+    end
+
+    @impl true
+    def handle_demand(:output, _size, _unit, _ctx, state) do
+      Process.sleep(10)
```
varsill, line 153: let's parametrize that `10` as an module's attribute

test/membrane/integration/demands_test.exs
```diff
@@ -189,19 +194,20 @@ defmodule Membrane.Integration.DemandsTest do
...
 
     for _i <- 1..10 do
       # during sleep below source should send about 100 buffers
-      Process.sleep(100 * 10)
+      Process.sleep(100 * RedemandingSource.sleep_time())
 
       Testing.Pipeline.execute_actions(pipeline, notify_child: {:sink, :pause_auto_demand})
 
       assert_pipeline_notified(pipeline, :sink, {:buff_no, buff_no})
       assert buff_no > 70
```
varsill, line 207: Do you have some kind of an idea how that value (`70`) and the other one used in assertion (`25`) is correlated with the number of buffers we expect to be sent (`100`)? Is it a linear correlation? If so, we can parametrize the `100` and use some kind of a coefficient to get the values used in assertions

lib/membrane/core/element/action_handler.ex
```diff
@@ -149,6 +151,32 @@ defmodule Membrane.Core.Element.ActionHandler do
     handle_redemand(out_ref, state)
   end
 
+  @impl CallbackHandler
+  def handle_action({:pause_auto_demand, in_refs}, cb, params, state) when is_list(in_refs) do
+    Enum.reduce(in_refs, state, fn in_ref, state ->
+      handle_action({:pause_auto_demand, in_ref}, cb, params, state)
```
varsill, line 157: I believe now you are repeating that pattern three times (`pause_auto_demand`, `pause_auto_demand` and `redemand`) - how about adding some helper function?

lib/membrane/core/element/action_handler.ex
```diff
@@ -149,6 +151,32 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+    end)
+  end
+
+  @impl CallbackHandler
+  def handle_action({:pause_auto_demand, in_ref}, _cb, _params, %State{type: type} = state)
+      when type in [:sink, :filter, :endpoint] and is_pad_ref(in_ref) do
```
varsill, line 163: [GENERAL THOUGHT] I am not sure if the user benefits from the fact that we have `is_pad_ref` guard here (and in a few other places in action handler).
Depending on whether user mistakenly passes a term that follows the pad name's pattern or not, he will either receive `Unknown pad` error, or `Unknown action` error.

test/membrane/integration/demands_test.exs
varsill, line 1: How about adding a test with a "mixing" element (the one that has provoked use to think about that "soft corcking" solution)?
In an ideal situation I would check if there is indeed no need to buffer the incoming buffers from a pad, which sends the buffers more frequently than the other.
However, I believe that can wait for the "hard blocking" solution with that test, 
but as for now, I would at least check such a scenario: 
1) The mixer passes the buffers received on its inputs exactly the moment they are received
2) The mixer stops autodemands on one of its pads
3) The element after the mixer should stop receiving buffers from the source connected with the "corcked" pad after some time, but it should still receive the buffers from the source connected to the "uncorcked" pad

lib/membrane/element/pad_data.ex
```diff
@@ -34,6 +34,7 @@ defmodule Membrane.Element.PadData do
           name: Pad.name(),
           ref: Pad.ref(),
           options: %{optional(atom) => any},
+          auto_demand_stopped?: boolean(),
```
varsill, line 37: You need to add the description of the `auto_demand_stopped?` flag to the `PadData` moduledoc since now that field is public

lib/membrane/element/action.ex
```diff
@@ -91,6 +91,31 @@ defmodule Membrane.Element.Action do
...
 
+  @typedoc """
+  Pauses auto-demanding on the specific pad.
+
+  The pad must have input direction and work in `:auto` flow control mode.
+
+  This action does not guarantee that no more buffers will arrive on the specific pad,
+  but ensures, that demand on this pad will not increase until returning
+  `#{inspect(__MODULE__)}.resume_auto_demand()` action. Number of buffers, that will
+  arrive on the pad, depends on the behaviour of the elements ealier in the pipeline.
```
varsill, line 102:
```suggestion
  arrive on the pad, depends on the behaviour of the elements earlier in the pipeline.
```

lib/membrane/element/action.ex
```diff
@@ -91,6 +91,31 @@ defmodule Membrane.Element.Action do
...
+
+  When auto-demanding is already paused, this action will have no effect.
```
varsill, line 104:
```suggestion
  When auto-demanding is already paused, this action has no effect.
```

lib/membrane/element/action.ex
```diff
@@ -91,6 +91,31 @@ defmodule Membrane.Element.Action do
...
+  @type pause_auto_demand :: {:pause_auto_demand, Pad.ref() | [Pad.ref()]}
+
+  @typedoc """
+  Resumes auto-demanding on the specific pad.
+
+  The pad must have input direction and work in `:auto` flow control mode.
+
+  This action reverts the effects of `#{inspect(__MODULE__)}.pause_auto_demand()` action.
+
+  When auto demanding is not paused, this action will have no effect.
```
varsill, line 115:
```suggestion
  When auto demanding is not paused, this action has no effect.
```

lib/membrane/core/element/demand_controller/auto_flow_utils.ex
```diff
@@ -7,13 +7,58 @@ defmodule Membrane.Core.Element.DemandController.AutoFlowUtils do
...
+              "Action #{inspect(action)} can only be returned for input pads, but #{inspect(pad_ref)} is an output pad"
+
+      %{flow_control: flow_control} ->
+        raise Membrane.ElementError,
+              "Action #{inspect(action)} can only be returned for pads with :auto flow control, but #{inspect(pad_ref)} pad flow control is #{inspect(flow_control)}"
+    end
+  end
+
+  @spec set_auto_demand_stopped_flag(Pad.ref(), boolean(), State.t()) :: State.t()
+  defp set_auto_demand_stopped_flag(pad_ref, new_value, state) do
```
varsill, line 47: [NIT] I think it will be easier to read it that way
```suggestion
  defp set_auto_demand_stopped_flag(pad_ref, stopped?, state) do
```

varsill (changes requested):

test/membrane/integration/demands_test.exs
FelonEkonom: I have added comments describing why I used these numbers

test/membrane/integration/demands_test.exs
FelonEkonom: Done

lib/membrane/core/element/demand_controller/auto_flow_utils.ex
```diff
@@ -7,13 +7,58 @@ defmodule Membrane.Core.Element.DemandController.AutoFlowUtils do
...
+
+  @spec set_auto_demand_paused_flag(Pad.ref(), boolean(), State.t()) :: State.t()
+  defp set_auto_demand_paused_flag(pad_ref, paused?, state) do
+    {old_value, state} =
+      PadModel.get_and_update_data!(state, pad_ref, :auto_demand_paused?, &{&1, paused?})
+
+    if old_value == paused? do
+      operation = if paused?, do: "pause", else: "resume"
+
+      Membrane.Logger.debug(
```
mat-hek, line 54:
```suggestion
      Membrane.Logger.debug_verbose(
```

test/membrane/integration/demands_test.exs
```diff
@@ -212,9 +218,80 @@ defmodule Membrane.Integration.DemandsTest do
...
+    @impl true
+    def handle_buffer(pad, buffer, ctx, state) do
+      buffer = %Membrane.Buffer{buffer | metadata: pad}
+
+      {pad_counter, state} = get_and_update_in(state, [:pads_counters, pad], &{&1, &1 + 1})
+
+      pad_upperbound = Map.get(state.pads_upperbounds, pad, :infinity)
+
+      actions =
+        if pad_counter > pad_upperbound and not ctx.pads[pad].auto_demand_paused? do
```
varsill, line 257: that `pad_counter > pad_upperbound` comparison is quite tricky :D

varsill (changes requested):

mat-hek (approved):
Looks great!

test/membrane/integration/demands_test.exs
FelonEkonom: term `some_integer < :infinity` is always equal `true`

test/membrane/integration/demands_test.exs
```diff
@@ -251,26 +251,25 @@ defmodule Membrane.Integration.DemandsTest do
...
-        Map.values(ctx.pads)
-        |> Enum.filter(&(&1.direction == :output and not &1.end_of_stream?))
-        |> Enum.map(&{:end_of_stream, &1.ref})
-
-      {actions, state}
+      if ctx.pads.output.end_of_stream? do
+        {[], state}
+      else
+        {[end_of_stream: :output], state}
+      end
```
varsill, line 272: If it's ok that the eos will be send on funnel's output right after first eos from input is received, we can leave it that way

varsill (approved):

PR #587, opened 2023-08-01
Fix elixir 1.15 compatibility on 0.10

WojciechBarczynski:


WojciechBarczynski:
After merging https://github.com/membraneframework/membrane_demo/pull/241 won't be useful.

PR #590, opened 2023-08-17
Make error in `LinkUtils.handle_child_pad_removed` more descriptive

sgfn:
Related issue: [RTC-315](https://membraneframework.atlassian.net/browse/RTC-315)

Right now, this fails with a match error. I blindly copied over the implementation from `remove_link`, you should change the messages as needed :smile:

[RTC-315]: https://membraneframework.atlassian.net/browse/RTC-315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mat-hek:
I suppose it's the case when a child removed its pad, but the parent removed it earlier. Thus, we should just do nothing. WDYT @FelonEkonom ?

FelonEkonom:
> I suppose it's the case when a child removed its pad, but the parent removed it earlier. Thus, we should just do nothing. WDYT @FelonEkonom ?

Agree, I will do it

PR #591, opened 2023-08-21
Fix stream format error message, fix race condition in handling child pad removed

FelonEkonom:
Related Jira tickets:
 * https://membraneframework.atlassian.net/browse/MC-219
 * https://membraneframework.atlassian.net/browse/MC-221


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -625,9 +623,17 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         )
 
       LinkUtils.handle_child_pad_removed(child, pad, state)
+    else
+      state
     end
   end
 
+  defp corresponding_link_exists?(child, pad, state) do
```
mat-hek, line 631: you can use `LinkUtils.get_link`

mat-hek (changes requested):

mat-hek (approved):

PR #592, opened 2023-08-28
Fix test failing on elixir 1.15

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-226


lib/membrane/logger.ex
```diff
@@ -36,12 +36,15 @@ defmodule Membrane.Logger do
...
+      end
+
       if is_function(message, 0) do
         fn ->
-          case message.() do
-            {content, metadata} -> {[prefix, content], metadata}
-            content -> [prefix, content]
-          end
+          message.()
+          |> message_result_mapper.()
```
mat-hek, line 47: why not just case?

test/membrane/testing/pipeline_test.exs
```diff
@@ -262,4 +271,13 @@ defmodule Membrane.Testing.PipelineTest do
...
     end
   end
+
+  defp otp_lower_than_26() do
+    System.otp_release()
+    |> Integer.parse()
+    |> case do
+      {otp_version, _version_sufix} when otp_version < 26 -> true
+      _else -> false
+    end
```
mat-hek, line 281:
```suggestion
    String.to_integer(System.otp_release()) < 26
```

lib/membrane/logger.ex
FelonEkonom: Because of credo error caused by dept = 3. This is a workaround to avoid it

test/membrane/testing/pipeline_test.exs
FelonEkonom: If `System.otp_release()` returns something like `26.1.3`, `String.to_integer()` will raise

lib/membrane/logger.ex
mat-hek: So I'd disable this warning locally (or increase dept to 3 globally?) or find another way to write this

mat-hek (approved):

PR #595, opened 2023-09-05
Fix flaky tests on Elixir 1.15

FelonEkonom:


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -681,7 +681,7 @@ defmodule Membrane.Core.Parent.ChildLifeController do
         Terminating.
         """)
 
-        {:error, {:membrane_child_crash, child_name}}
+        {:error, {:membrane_child_crash, {child_name, reason}}}
```
mat-hek, line 684: how about less nesting?

```suggestion
        {:error, {:membrane_child_crash, child_name, reason}}
```

test/membrane/integration/demands_test.exs
```diff
@@ -20,6 +20,9 @@ defmodule Membrane.Integration.DemandsTest do
   end
 
   defp test_pipeline(pid) do
+    assert_sink_playing(pid, :sink)
+    # assert_pipeline_notified(pid, :sink, :playing)
```
mat-hek, line 24: leftover

PR #601, opened 2023-09-25
Delete handle_buffers_batch callback

FelonEkonom:
Related Jira ticket: https://membraneframework.atlassian.net/browse/MC-238

### DIFF
--- a/lib/membrane/core/element/buffer_controller.ex
+++ b/lib/membrane/core/element/buffer_controller.ex
@@ -91,7 +91,7 @@ defmodule Membrane.Core.Element.BufferController do
   end
 
   @doc """
-  Executes `handle_buffers_batch` callback.
+  Executes `handle_buffer` callback.
   """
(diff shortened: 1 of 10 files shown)


lib/membrane/core/element/buffer_controller.ex
```diff
@@ -113,12 +113,16 @@ defmodule Membrane.Core.Element.BufferController do
...
   defp do_exec_buffer_callback(pad_ref, buffers, state) do
-    CallbackHandler.exec_and_handle_callback(
-      :handle_buffers_batch,
-      ActionHandler,
-      %{context: &CallbackContext.from_state/1},
-      [pad_ref, buffers],
-      state
-    )
+    buffers
+    |> List.wrap()
```
mat-hek, line 117: I guess we should always have a list here

mat-hek (approved):

PR #614, opened 2023-09-29
Sort component state fields in the error logs in the order from the most to the least important

### LINKED ISSUE #616: Implement `Inspect` protocol in component states
This will make error logs much prettier

FelonEkonom:
Closes #616


lib/membrane/core/inspect.ex
```diff
@@ -0,0 +1,90 @@
...
+    :supplying_demand?,
+    :handling_action?,
+    :stalker,
+    :resource_guard,
+    :subprocess_supervisor,
+    :children_log_metadata,
+    :handle_demand_loop_counter,
+    :demand_size,
+    :playback_queue,
+    :pads_to_snapshot
```
mat-hek, line 36:
```suggestion
    :children,
    :links,
    :crash_groups,
    :pending_specs,
    :synchronization,
    :delayed_demands,
    :effective_flow_control,
    :initialized?,
    :terminating?,
    :setup_incomplete?,
    :supplying_demand?,
    :handling_action?,
    :stalker,
    :resource_guard,
    :subprocess_supervisor,
    :children_log_metadata,
    :handle_demand_loop_counter,
    :demand_size,
    :pads_to_snapshot,
    :playback_queue,
    :pads_data
```

lib/membrane/core/inspect.ex
```diff
@@ -0,0 +1,90 @@
...
+      field_doc_extractor,
+      break: :strict
+    )
+  end
+
+  @spec ensure_all_struct_fields_inspected!(module()) :: :ok | no_return()
+  def ensure_all_struct_fields_inspected!(state_module) do
+    state_module.__info__(:struct)
+    |> Enum.map(& &1.field)
+    |> Enum.filter(&(&1 not in @fields_order))
```
mat-hek, line 67:
```suggestion
    |> Enum.reject(&(&1 in @fields_order))
```

lib/membrane/core/inspect.ex
```diff
@@ -0,0 +1,90 @@
...
+    |> case do
+      [] ->
+        :ok
+
+      fields ->
+        raise """
+        Fields #{inspect(fields)} from #{inspect(state_module)} structure are omitted in @fields_order module attribute in #{inspect(__MODULE__)}. Add them to @fields_order, to define where they will occur in the error logs.
```
mat-hek, line 74:
```suggestion
        Fields #{inspect(fields)} from #{inspect(state_module)} structure are omitted in @fields_order module attribute in #{inspect(__MODULE__)}.\
        Add them to @fields_order, to define where they will occur in the error logs.
```

mat-hek (changes requested):

lib/membrane/core/inspect.ex
FelonEkonom: Are you sure we want to have `pads_data` at the end, even after things like `delayed_demands` or `handle_demand_loop_counter`?

mat-hek (changes requested):
As we spoke, let's use the order of fields as specified in each state module

mat-hek (approved):

PR #626, opened 2023-10-04
Fix clock selection

mat-hek:


FelonEkonom (commented):
As we spoke, let's delete `clock_provider` from spec options and move this functionality to a new action

FelonEkonom (approved):

Issue #622, opened 2023-10-02
Handle board automatization bot


michalb666:
@FelonEkonom - do uzupełnienia

FelonEkonom:
- [ ] Auto close issue opened beyond membrane_core with a proper comment
- [ ] Update shadow text in the place, where somebody would put an issue description 
- [ ] Put newly created issue in membrane_core on this board

PR #639, opened 2023-10-12
Write github actions needed for auto closing issues opened beyond core

FelonEkonom:
Relates to #622


.github/actions/close_issue/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        otp-version: '26.1'
+        elixir-version: '1.15.6'
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - name: Create ticket and close issue
+      run: |
+        gh issue edit $ISSUE_URL --add-project "Smackore" --add-label closed-by-membrane-bot
+        sleep 10        
+        export TICKET_ID=$(gh project item-list 19 --owner membraneframework --format json --limit 10000000 | elixir get_ticket_id.exs "$ISSUE_URL" | awk '/TICKET_ID/{print $2}')
+        gh issue close $ISSUE_URL --comment "Only PRs welcome ;)" --reason "not planned"
```
mkaput, line 31: I don't think this is a polite wording of this message.

.github/actions/close_issue/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        otp-version: '26.1'
+        elixir-version: '1.15.6'
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - name: Create ticket and close issue
+      run: |
+        gh issue edit $ISSUE_URL --add-project "Smackore" --add-label closed-by-membrane-bot
+        sleep 10        
+        export TICKET_ID=$(gh project item-list 19 --owner membraneframework --format json --limit 10000000 | elixir get_ticket_id.exs "$ISSUE_URL" | awk '/TICKET_ID/{print $2}')
```
mkaput, line 30:
```suggestion
        export TICKET_ID=$(gh project item-list 19 --owner membraneframework --format json --limit 10000000 | elixir get_ticket_id.exs "$ISSUE_URL")
```

why not just print ticket id without unnecessary stuff?

mkaput, line 30: 19 is a magic constant, consider at least extract it into a constant env var

scripts/get_ticket_id.exs
mkaput, line 1: Python is preinstalled in GH actions, why not just use it or `jq` ?

.github/actions/close_issue/action.yml
FelonEkonom: Sorry, leftover ;)

.github/actions/close_issue/action.yml
FelonEkonom: Because `.exs` script sometimes prints also logs from `Mix.install/1`, like `Resolving Hex dependencies...`

scripts/get_ticket_id.exs
FelonEkonom: Do you mean to use it instead of elixir?

mkaput: yes, setup-elixir is just an unnecessary cost here

FelonEkonom: But does this cost hurts us anyhow?

mkaput: Frustration that action is working longer than it should. I don't think you'll saturate concurrency limits, given this runs on Linux runners only.

mkaput (commented):
I'm leaving the issue of this script being written in Elixir to you. I'm not the owner of this code, so I'm not clicking accept, but this looks okay-ish to my eyes.

mat-hek (approved):

.github/actions/close_issue/action.yml
mat-hek: You could set `MIX_QUIET=1`, though it seems it doesn't work for `Mix.install` :P It may be worth to report an issue though

PR #685, opened 2023-11-27
Fix timer running late

varsill:
This PR: 
* guards the timer tick in `TimerController.handle_tick` with supplying_demand? flag

### DIFF
--- a/lib/membrane/core/element.ex
+++ b/lib/membrane/core/element.ex
@@ -256,6 +256,16 @@ defmodule Membrane.Core.Element do
   end
 
   defp do_handle_info(Message.new(:timer_tick, timer_id), state) do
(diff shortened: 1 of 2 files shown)


lib/membrane/core/element.ex
```diff
@@ -256,7 +256,10 @@ defmodule Membrane.Core.Element do
   end
 
   defp do_handle_info(Message.new(:timer_tick, timer_id), state) do
+    state = %{state | supplying_demand?: true}
```
mat-hek, line 259: Let's add a comment explaining why we do this

mat-hek (changes requested):

lib/membrane/core/element.ex
varsill: done

PR #708, opened 2024-01-17
Bug in handling actions returned from handle_tick

FelonEkonom:
Closes https://github.com/membraneframework/membrane_core/issues/695

### DIFF
--- a/test/membrane/integration/actions_handling_order_test.exs
+++ b/test/membrane/integration/actions_handling_order_test.exs
@@ -1,0 +1,40 @@
+defmodule Membrane.Integration.ActionsHandlingOrderTest do
+  use ExUnit.Case, async: true
+
+  import Membrane.ChildrenSpec
+  import Membrane.Testing.Assertions
+
+  defmodule DelayedPlayingPipeline do
+    use Membrane.Pipeline
+
+    @impl true
+    def handle_setup(_ctx, state) do
+      self() |> send(:wake_up_buddy)
+      {[setup: :incomplete, start_timer: {:one, Membrane.Time.second()}], state}
+    end
+
+    @impl true
+    def handle_info(:wake_up_buddy, _ctx, state) do
+      {[setup: :complete, timer_interval: {:one, Membrane.Time.seconds(10)}], state}
+    end
+
+    @impl true
+    def handle_playing(_ctx, state) do
+      {[timer_interval: {:one, Membrane.Time.seconds(5)}], state}
+    end
+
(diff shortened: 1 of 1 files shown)


lib/membrane/core/element/action_handler.ex
mat-hek, line 1: It seems this is not needed now: https://github.com/membraneframework/membrane_core/blob/fix-handling-actions-bug/lib/membrane/core/element/action_handler.ex#L309

mat-hek (changes requested):

lib/membrane/bin.ex
```diff
@@ -310,6 +312,9 @@ defmodule Membrane.Bin do
       @behaviour unquote(__MODULE__)
       @before_compile unquote(__MODULE__)
 
+      require Membrane.Core.Parent
+      Membrane.Core.Parent.bring_after_compile_check()
```
mat-hek, line 316:
```suggestion
@after_compile {Membrane.Core.Parent, :check_for_deprecated_callbacks}
```

lib/membrane/core/parent.ex
```diff
@@ -2,4 +2,27 @@ defmodule Membrane.Core.Parent do
...
+          :membrane_core v1.0.1, but it is implemented in #{inspect(env.module)}
+          """
+
+          IO.warn(warn_message, [])
+        end
+
+        :ok
+      end
+    end
+  end
```
mat-hek, line 27:
```suggestion
def check_for_deprecated_callbacks(env, _bytecode) do
  modules_whitelist = [Membrane.Testing.Pipeline]

  if env.module not in modules_whitelist and
       Module.defines?(env.module, {:handle_spec_started, 3}, :def) do
    warn_message = """
    Callback handle_spec_started/3 has been deprecated since \
    :membrane_core v1.0.1, but it is implemented in #{inspect(env.module)}
    """

    IO.warn(warn_message, [])
  end

  :ok
end
```

lib/membrane/pipeline.ex
```diff
@@ -204,9 +204,11 @@ defmodule Membrane.Pipeline do
             ) :: {[Action.common_actions()], state()}
 
   @doc """
+  This callback is deprecated since v1.0.1.
```
mat-hek, line 207: If we deprecate the callback, I'd go for releasing the current master as v1.0.1 and release this as v1.1.0-rc0 with auto flow queues

mat-hek (approved):

PR #745, opened 2024-02-08
lib/membrane/core/bin/pad_controller.ex: refactored Enum to Stream

bradhanks:
No one asked me to but I refactored Membrane.Core.Bin.PadController. respond_links/2  to use Stream. 
I enjoyed it as an exercise so feel free to just close this if you don't want it. :)

### DIFF
--- a/lib/membrane/core/bin/pad_controller.ex
+++ b/lib/membrane/core/bin/pad_controller.ex
@@ -186,7 +186,7 @@ defmodule Membrane.Core.Bin.PadController do
         end
       end,
       fn state -> {[state], state} end,
-      fn _ -> :ok end
+      fn _acc -> :ok end
     )
     |> Enum.to_list()
     |> List.first()


mat-hek:
Hi @bradhanks, can you elaborate on what improvements does this change bring? ;)

bradhanks:
I'm going to assert that it  marginally improves the application's performance and reduces memory usage, but really  I have an unhealthy relationship with Stream.transform. :)

bradhanks:
> Streams can indeed reduce memory usage, but this usually comes at the cost of CPU consumption, because Streams have more computational overhead than Enums. However, as you pointed out, the performance change will be marginal in this case, and I don't like `Stream.transform` as much as you do, so I'm going to close this ;)
> 
> 
> 
> If you'd like to improve the core's performance, you should investigate the hot path, which includes sending, processing and receiving buffers and demands ;)

That's great to know. Thanks for the feedback. 👍

PR #756, opened 2024-02-14
Timestamp queue POC

FelonEkonom:
Relates to https://github.com/membraneframework/membrane_core/issues/734


lib/membrane/timestamp_queue.ex
mat-hek, line 1: As we spoke, the following features would be useful
- adding pad before pushing anything on it
- specifying the minimal or average chunk size
- flushing everything from the queue
- queue live mode (separate PR)

lib/membrane/timestamp_queue.ex
```diff
@@ -0,0 +1,338 @@
...
+  If amount of buffers from pad associated with popped buffer just dropped below
+  `pause_demand_boundary`, the suggested actions list contains `t:Action.resume_auto_demand()`
+  action, otherwise it is an empty list.
+
+  If the queue cannot return any buffer, returns `:none` in it's place instead. Note, that
+  the queue doesn't have to be empty to be unable to return a buffer - sometimes queue has to
+  keep up to 1 buffer for each pad, to be able to work correctly.
+  """
+  @spec pop(t()) :: {[Action.resume_auto_demand()], popped_value() | :none, t()}
+  def pop(%__MODULE__{} = timestamp_queue) do
```
mat-hek, line 213: As we spoke, let's remove this function, at least for now

mat-hek (changes requested):

PR #779, opened 2024-03-18
Ensure handle_demand busy wait infinite loop doesn't occur & deprecate message_child

FelonEkonom:
Prerequisite for https://github.com/membraneframework/membrane_core/pull/769
Related to https://github.com/membraneframework/membrane_core/issues/706


test/membrane/integration/delayed_demands_loop_test.exs
```diff
@@ -0,0 +1,79 @@
...
+        get_child(:source)
+        |> via_in(:input, auto_demand_size: auto_demand_size)
+        |> child(Debug.Sink)
+      end)
+      |> Stream.take(sinks_number)
+      |> Enum.concat([child(:source, Source)])
+
+    pipeline = Testing.Pipeline.start_link_supervised!(spec: spec)
+
+    Process.sleep(1_000)
```
mat-hek, line 70: you can assert_start_of_stream on each sink instead

test/membrane/integration/delayed_demands_loop_test.exs
```diff
@@ -0,0 +1,79 @@
...
+
+    for _i <- 1..(auto_demand_size + 5) do
```
mat-hek, line 72: why this number?

FelonEkonom: To ensure, that this line 
https://github.com/membraneframework/membrane_core/blob/d1068aa67c75f9f8892b6b30e5b9637c05cb06e5/lib/membrane/core/element/demand_controller.ex#L79
will be executed

FelonEkonom: Now I think, that it would be good to have auto_demand_size <= delayed_demand_loop_counter_limit, to ensure, that counter won't reset after reaching linked code, I will change this

mat-hek: so let’s change the naming or add a comment indicating it’s to exceed the loop counter limit

test/membrane/integration/delayed_demands_loop_test.exs
```diff
@@ -54,7 +54,11 @@ defmodule Membrane.Test.DelayedDemandsLoopTest do
   end
 
   defp do_test(sinks_number) do
+    # auto_demand_size is smaller than delayed_demands_loop_counter_limit to ensure, that counter
+    # doesn't reset after calling handle_demand caused by reading positive atomic demand value
+    # during the snaphsot
```
mat-hek, line 59:
```suggestion
    # auto_demand_size is smaller than delayed_demands_loop_counter_limit, to ensure that
    # after calling handle_demand and making a snapshot, the atomic demand value is still positive
    # and the counter is not reset
```

PR #780, opened 2024-03-19
Bump Ratio to 4.0.1

FelonEkonom:


mix.exs
```diff
@@ -144,7 +144,7 @@ defmodule Membrane.Mixfile do
       {:qex, "~> 0.3"},
       {:telemetry, "~> 1.0"},
       {:bunch, "~> 1.6"},
-      {:ratio, "~> 3.0"},
+      {:ratio, ">= 4.0.1 and < 5.0.0"},
```
mat-hek, line 147: This is an equivalent of `"~>> 4.0.1"`. However, I think it should be

```suggestion
      {:ratio, "~> 3.0 or ~>> 4.0.1"},
```

so it doesn't block updating [plugins that depend on ratio](https://github.com/search?q=org%3Amembraneframework+%3Aratio+path%3Amix.exs&type=code)

mat-hek (changes requested):

mat-hek (approved):

PR #801, opened 2024-05-13
Implement handle_spec_setup_completed and handle_spec_playing

FelonEkonom:


lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -319,6 +319,11 @@ defmodule Membrane.Core.Parent.ChildLifeController do
     if Enum.all?(spec_data.children_names, &Map.fetch!(children, &1).initialized?) and
          Enum.empty?(spec_data.dependent_specs) do
       Membrane.Logger.debug("Spec #{inspect(spec_ref)} status changed to initialized")
+
+      state =
+        MapSet.to_list(spec_data.children_names)
+        |> StartupUtils.exec_handle_spec_setup_completed(state)
```
mat-hek, line 325: The problem here is that the children_names can be empty, and we call `handle_spec_started` with an empty list. This makes me lean towards having `handle_child_setup` and `handle_child_playing` instead.

mat-hek (changes requested):

lib/membrane/core/parent/child_life_controller.ex
```diff
@@ -610,6 +626,32 @@ defmodule Membrane.Core.Parent.ChildLifeController do
...
+    exec_child_playback_related_callbacks(:handle_child_setup_completed, children_names, state)
+  end
+
+  @spec handle_children_playing(MapSet.t(Child.name()) | [Child.name()], Parent.state()) ::
+          Parent.state()
+  def handle_children_playing(children_names, state) do
+    exec_child_playback_related_callbacks(:handle_child_playing, children_names, state)
+  end
+
+  defp exec_child_playback_related_callbacks(callback, children_names, state) do
```
mat-hek, line 641:
```suggestion
  defp exec_child_playback_related_callback(callback, children_names, state) do
```

lib/membrane/child_entry.ex
```diff
@@ -38,6 +41,7 @@ defmodule Membrane.ChildEntry do
     :group,
     initialized?: false,
     ready?: false,
-    terminating?: false
+    terminating?: false,
+    playback: :stopped
```
mat-hek, line 45: why exactly do we need that?

lib/membrane/core/parent/lifecycle_controller.ex
```diff
@@ -43,19 +43,29 @@ defmodule Membrane.Core.Parent.LifecycleController do
 
     activate_syncs(state.children)
 
-    Enum.each(state.children, fn {_name, %{pid: pid, ready?: ready?}} ->
-      if ready?, do: Message.send(pid, :play)
-    end)
+    {pinged_children, state} =
+      Enum.flat_map_reduce(state.children, state, fn
+        {child, %{ready?: true, terminating?: false, playback: :stopped, pid: pid}}, state ->
+          Message.send(pid, :play)
```
mat-hek, line 49: here we make the children play, but we don't exec the handle_child_playing callback

lib/membrane/child_entry.ex
FelonEkonom: Before my changes we were sending `:play` to each child between 1 and 2 times. This field allow me distinct, if I have already done it, because I don't want to execute callback `handle_child_playing` more than once for a single child

lib/membrane/core/parent/lifecycle_controller.ex
FelonEkonom: We do it in line 68

FelonEkonom: It is done this way to:
1. Ping children
2. Exec `handle_playing`
3. Exec `handle_child_playing`

In this particular order

mat-hek (approved):

lib/membrane/child_entry.ex
mat-hek: I think we never send `play` two times. If we do, it seems like a bug.

mat-hek (changes requested):

lib/membrane/child_entry.ex
```diff
@@ -21,7 +21,7 @@ defmodule Membrane.ChildEntry do
           module: module,
           options: struct | nil,
           component_type: :element | :bin,
-          playback: Membrane.Playback.t(),
+          # playback: Membrane.Playback.t(),
```
mat-hek, line 24: leftover

lib/membrane/child_entry.ex
```diff
@@ -41,7 +41,7 @@ defmodule Membrane.ChildEntry do
     :group,
     initialized?: false,
     ready?: false,
-    terminating?: false,
-    playback: :stopped
+    terminating?: false
+    # playback: :stopped
```
mat-hek, line 45: leftover

mat-hek (approved):

PR #809, opened 2024-05-22
Fix doc for `handle_start_of_stream`

takumihara:


lib/membrane/element/with_input_pads.ex
```diff
@@ -32,7 +32,7 @@ defmodule Membrane.Element.WithInputPads do
                  | Membrane.Element.Action.forward()
                ], Element.state()}
   @doc """
-  Callback invoked when element receives `Membrane.Event.StartOfStream` event.
+  Callback invoked when element receives `Membrane.Core.Events.StartOfStream` event.
```
FelonEkonom, line 35: I agree that this doc is not very helpful. 
Modules prefixed with `Membrane.Core` are not part of public API, 
```suggestion
  Callback invoked when element receives the first buffer from the specific pad, before `c:#{__MODULE__}.handle_buffer/4`. 
```

FelonEkonom (changes requested):

FelonEkonom (approved):

PR #810, opened 2024-05-23
Modify 5s linking timeout constraint

FelonEkonom:
Solves https://github.com/membraneframework/membrane_core/issues/786


lib/membrane/core/bin/pad_controller.ex
```diff
@@ -102,15 +96,19 @@ defmodule Membrane.Core.Bin.PadController do
...
-  @spec handle_linking_timeout(Pad.ref(), State.t()) :: :ok | no_return()
+  @spec handle_linking_timeout(Pad.ref(), State.t()) :: State.t() | no_return()
   def handle_linking_timeout(pad_ref, state) do
-    case PadModel.get_data(state, pad_ref) do
-      {:ok, %{endpoint: nil} = pad_data} ->
-        raise Membrane.LinkError,
-              "Bin pad #{inspect(pad_ref)} wasn't linked internally within timeout. Pad data: #{inspect(pad_data, pretty: true)}"
+    case Map.fetch(state.linking_timeout_counters, pad_ref) do
+      {:ok, 1} ->
+        Map.update!(state, :linking_timeout_counters, &Map.delete(&1, pad_ref))
```
mat-hek, line 103: Remember we talked about something similar, but couldn't we store the reference to the timer in `pad_data` and compare it when the timeout arrives?

test/membrane/integration/linking_test.exs
```diff
@@ -353,6 +353,7 @@ defmodule Membrane.Integration.LinkingTest do
     Membrane.Pipeline.terminate(pipeline)
   end
 
+  @tag :xd
```
mat-hek, line 356: xd

mat-hek (changes requested):

lib/membrane/core/bin/pad_controller.ex
FelonEkonom: It is not a good idea, this approach generates some weird race conditions and it's implementation would be far more complex. I don't se any pros of it, only cons

FelonEkonom: Conclusion from out last talk was that we don't want to do this

mat-hek (changes requested):
As we spoke, the best approach would be to create the pad data earlier and store the timer reference there

lib/membrane/core/bin/pad_controller.ex
```diff
@@ -50,8 +49,7 @@ defmodule Membrane.Core.Bin.PadController do
     state =
       case PadModel.get_data(state, pad_ref) do
         {:error, :unknown_pad} ->
-          init_pad_data(pad_ref, pad_info, state)
-          |> Map.update!(:pad_refs, &[pad_ref | &1])
```
mat-hek, line 54: Why remove `pad_refs`? In case you forgot, it was added by yourself in this PR: https://github.com/membraneframework/membrane_core/pull/641 :P

FelonEkonom: Ok, I didn't remember it 🙈 There was 1 place, where the pad was removed from `pads_data` but not from `pad_refs`, so without this PR value stored under this field may be incorrect and I saw that nothing relies on the value of this field. I guess that the reason of this field was to have more readable logs, but has anybody ever used it in this purpose?

mat-hek: I didn't, but hard to say in general. I have the following idea:
- call it `inspect_pad_refs` and assign to it a function that will accept the state and return the pad refs
- in `Membrane.Core.Inspect`, if a field starts with `inspect_`, we'll call the function in it instead of printing it directly

I don't like it much, but should be better than what we have. Otherwise I'd just remove this field, but it should be removed from the element as well

FelonEkonom: there is no such a field in the element state

mat-hek (approved):

PR #811, opened 2024-05-23
Reshape put_in/update_in calls in crashgroups

FelonEkonom:
1. In FishJam tests error logs didn't show up, so I added printing them in debug logs in one place.
2. For some reason there was a bug occurring in statements like`put_in(state, [:crash_groups, id, :members], ...)` despite de fact that module `CrashGroup` contains definitions of proper functions, so i rewrote it to statements like `put_in(state.crash_groups[id].members, ...)`

### DIFF
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -209,6 +209,9 @@ defmodule Membrane.Core.CallbackHandler do
     handler_module.handle_end_of_actions(state)
   end
 
+  # We log it, because sometimes, for some reason, crashing process doesn't cause
+  # printing error logs on stderr, so this debug log allows us to get some info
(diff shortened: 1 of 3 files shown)


lib/membrane/core/callback_handler.ex
```diff
@@ -198,10 +200,26 @@ defmodule Membrane.Core.CallbackHandler do
...
+            log_debug_orginal_error(action, e, __STACKTRACE__)
+
             reraise e, __STACKTRACE__
         end
       end)
 
     handler_module.handle_end_of_actions(state)
   end
+
+  defp log_debug_orginal_error(action_or_actions, error, stacktrace) do
```
mat-hek, line 212: let's add a comment why we do that

mat-hek (approved):

PR #822, opened 2024-06-10
Ensure error logs are always printed

FelonEkonom:


lib/membrane/core/element.ex
```diff
@@ -198,7 +209,9 @@ defmodule Membrane.Core.Element do
       :erlang.process_info(self(), :message_queue_len) |> elem(1)
     )
 
-    do_handle_info(message, state)
+    Macros.log_on_error do
+      do_handle_info(message, state)
+    end
```
varsill: why `Telemetry.report_metric` is outside of the `Macros.log_on_error`?

lib/membrane/core/macros.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Membrane.Core.Macros do
+  @moduledoc false
+
+  # For some reason GenServer processes sometimes don't print logs about crash, so
+  # we add this macro, to ensure that error logs wil be always logged
```
varsill, line 5:
```suggestion
  # we add this macro, to ensure that error logs are always printed
```

lib/membrane/core/macros.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Membrane.Core.Macros do
```
mat-hek, line 1:
```suggestion
defmodule Membrane.Core.Utils do
```

mat-hek (approved):

varsill (changes requested):
How about adding it to other GenServers that we define in Core as well (like ResourceGuard etc.?)

varsill (approved):
Initially I just thought of adding it only in other processes that can execute users code (just like `ResourceGuard`, which executes registered user's lambda), but since you have added logging to all GenServers it's even better, as it will be easier for us to debug if we made some mistakes :D

PR #852, opened 2024-08-13
Fix table rendering for raw audio and video

harrisi:
`EarmarkParser` expects a blank line before tables. It seems to handle a header immediately before a table correctly, so I left those as is.


mat-hek (approved):

mat-hek:
Thanks!

PR #876, opened 2024-09-27
Add max_instances option for dynamic pads

FelonEkonom:
Closes https://github.com/membraneframework/membrane_core/issues/870


lib/membrane/bin/pad_data.ex
```diff
@@ -23,6 +24,7 @@ defmodule Membrane.Bin.PadData do
           availability: Membrane.Pad.availability(),
           direction: Membrane.Pad.direction(),
           name: Membrane.Pad.name(),
+          max_cardinality: Membrane.Pad.max_cardinality() | nil,
```
mat-hek, line 27: What about `max_instances`? :P

mat-hek (commented):
Let's add it to the docs generation too https://github.com/membraneframework/membrane_core/blob/a9b052d10f9014612bb163fad3f81d3c7e84268a/lib/membrane/core/child/pads_specs.ex#L274

lib/membrane/pad.ex
```diff
@@ -102,6 +102,12 @@ defmodule Membrane.Pad do
   """
   @type accepted_format :: module() | (pattern :: term())
 
+  @typedoc """
+  Describes maximal number of instances of dynamic pad (`availability: :on_request`) that
+  can occur within single component.
```
mat-hek, line 107:
```suggestion
  can occur simultaneously within a single component.
```

lib/membrane/pad.ex
```diff
@@ -126,7 +135,8 @@ defmodule Membrane.Pad do
            accepted_format: accepted_format(),
            flow_control: flow_control(),
            options: Keyword.t(),
-           demand_unit: Buffer.Metric.unit()}
+           demand_unit: Buffer.Metric.unit(),
+           max_cardinality: max_cardinality()}
```
varsill, line 139: It's probably not the scope of this PR, but is this `element_spec` a proper typespec? I think that it doesn't highlight the fact that some of these fields are mandatory, while the others are optional

varsill: BTW. - Perhaps it would be good to write somewhere that `max_cardinality` defaults to `:infinity` for dynamic pads? I don't think there is a way to tell that without checking the code.

varsill (changes requested):

mat-hek (approved):

PR #879, opened 2024-09-30
Additional telemetry events

FelonEkonom:


lib/membrane/bin.ex
```diff
@@ -348,6 +348,10 @@ defmodule Membrane.Bin do
 
       Membrane.Core.Child.PadsSpecs.ensure_default_membrane_pads()
 
+      @doc false
+      @spec membrane_component_type() :: :pipeline | :bin | :element
```
varsill, line 352: [NIT] Technically speaking, `Membrane.Bin.membrane_component_type()` always returns `:bin`, so the typespec should be slightly different

PR #900, opened 2024-10-29
Generate warnings related to pads with `:auto` flow control 

FelonEkonom:
Closes https://github.com/membraneframework/membrane_core/issues/868
Closes https://github.com/membraneframework/membrane_core/issues/869

### DIFF
--- a/lib/membrane/core/child/pads_specs.ex
+++ b/lib/membrane/core/child/pads_specs.ex
@@ -207,11 +207,11 @@ defmodule Membrane.Core.Child.PadsSpecs do
 
     if auto_input_pads_possible_number >= 2 and auto_output_pads_possible_number >= 2 do
       Membrane.Logger.warning("""
-      This element might have multiple input pads with `flow_control: :auto` and multiple output pads with \
-      `flow_control: :auto` at the same time. Notice, that lack of demand on any of output pads with \
+      #{inspect(env.module)}  might have multiple input pads with `flow_control: :auto` and multiple output \
+      pads with `flow_control: :auto` at the same time. Notice, that lack of demand on any of output pads with \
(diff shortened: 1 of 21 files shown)


mat-hek (changes requested):
I'm not sure that warnings are the best way to inform a user about behaviour that may be completely desired. IMO warnings should indicate that even though something won't necessarily break, it's definitely wrong, which is not the case here. That said, I have no idea for a better way to solve this :P

lib/membrane/core/child/pads_specs.ex
```diff
@@ -150,6 +154,74 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
+  @spec generate_flow_control_related_warnings(
+          pads :: [{Pad.name(), Pad.description()}],
+          env :: Macro.Env.t()
+        ) :: :ok
+  defp generate_flow_control_related_warnings(pads, env) do
+    used_module =
+      env.module
+      |> Module.get_attribute(:__membrane_element_type__, :bin)
+      |> case do
+        :bin -> Membrane.Bin
```
mat-hek, line 166: shouldn't this apply to elements only?

lib/membrane/endpoint.ex
```diff
@@ -25,6 +25,9 @@ defmodule Membrane.Endpoint do
 
   Options:
     - `:bring_pad?` - if true (default) requires and aliases `Membrane.Pad`
+    - `:flow_control_related_warnings?` - if true (default) generates compile-time warnings \
```
mat-hek, line 28: What about

```suggestion
    - `:flow_control_hints?` - if true (default) generates compile-time warnings \
```

mat-hek (approved):

PR #908, opened 2024-11-19
Rename ForwardingFilter to Connector

FelonEkonom:


lib/membrane/forwarding_filter.ex
mat-hek, line 1: rename file

mat-hek (approved):

PR #909, opened 2024-11-21
Detect diamonds

### LINKED ISSUE #873: Traverse pipeline topology looking for the flow control stuck vunerability
Every randomized time interval check if there is a diamond structure in the pipeline topology with a root having `:auto` flow control and `pull` effective flow control (and proper flow control on both paths to the "leaf"). 

If it occurs, add a debug/warning log informing about possibility of getting stuck, if the data will be unbalanced and possible solutions (setting `auto_demand_size` to a bigger number, modifying pipeline topology).

Another possible approach is to auto-magically-bump `auto_demand_size` in case of getting stuck

FelonEkonom:
Closes #873


lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,207 @@
...
+
+  defp send_after_to_self(message, seconds \\ 10) do
+    send_after_time = Membrane.Time.seconds(seconds) |> Membrane.Time.as_milliseconds(:round)
+    self() |> Process.send_after(message, send_after_time)
+    :ok
+  end
+
+  defp get_component_path() do
+    # adding @component_path_prefix to component path causes that component path
+    # always has more than 64 bytes, so it won't be copied during sending a message
```
mat-hek, line 194: I'm not convinced it's worth the hassle. Did you do any benchmarks? BTW, it seems we construct this 'serialized' component path every time we need it - I'd do that once and keep it in state.

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,207 @@
+defmodule Membrane.Core.Element.DiamondDetectionController do
+  @moduledoc false
+
```
mat-hek, line 3: Let's describe how the algorithm works

lib/membrane/core/element.ex
```diff
@@ -281,6 +282,50 @@ defmodule Membrane.Core.Element do
...
+
+  defp do_handle_info(Message.new(:diamond_detection_trigger, trigger_ref), state) do
+    state = DiamondDetectionController.handle_diamond_detection_trigger(trigger_ref, state)
+    {:noreply, state}
+  end
+
+  defp do_handle_info(Message.new(:delete_diamond_detection_trigger_ref, trigger_ref), state) do
+    state = DiamondDetectionController.delete_diamond_detection_trigger_ref(trigger_ref, state)
+    {:noreply, state}
+  end
```
mat-hek, line 327: These messages seem quite internal to the diamond detection. I'd make them a single message and dispatch within DiamondDetectionController

lib/membrane/core/element/state.ex
```diff
@@ -46,7 +47,12 @@ defmodule Membrane.Core.Element.State do
           stalker: Membrane.Core.Stalker.t(),
           satisfied_auto_output_pads: MapSet.t(),
           awaiting_auto_input_pads: MapSet.t(),
-          resume_delayed_demands_loop_in_mailbox?: boolean()
+          resume_delayed_demands_loop_in_mailbox?: boolean(),
+          diamond_detection_ref_to_path: %{
+            optional(reference()) => PathInGraph.t()
+          },
+          diamond_detection_trigger_refs: MapSet.t(reference()),
+          diamond_detection_postponed?: boolean()
```
mat-hek, line 55: let's put these into a single `diamond_detection` map/struct

mat-hek (changes requested):

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,347 @@
...
+  # demand on the input, because one of the pads output with :auto flow control doesn't
+  # have positive demand, so the whole pipeline will get stuck and won't process more data.
+
+  # The algorithm is made of two phases: (1) triggering and (2) proper searching.
+
+  # (1) Triggering
+
+  # Let's notice that:
+  #   * a new diamond can be created only after linking a new spec
+  #   * if the new spec caused some new diamond to occur, this diamond will contain some of
```
mat-hek, line 33:
```suggestion
  #   * if the new spec created a new diamond, this diamond will contain some of
```

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,347 @@
...
+  #   * MP4 demuxer that has two output pads
+  #   * MKV muxer that is linked to both of them
+  # and let's assume that the MP4 container that is consumed by the MP4 demuxer is unbalanced.
+  # If the MKV muxer has pads working in pull mode, then demand on one pad will be satisfied,
+  # but on the other won't, because the MP4 container is unbalanced. Then, if MP4 demuxer
+  # has pads in auto flow control and its effective flow control is set to :pull, it won't
+  # demand on the input, because one of the pads output with :auto flow control doesn't
+  # have positive demand, so the whole pipeline will get stuck and won't process more data.
+
+  # The algorithm is made of two phases: (1) triggering and (2) proper searching.
```
mat-hek, line 27: why _proper_ searching? :P

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,347 @@
...
+  @component_path_prefix "__membrane_component_path_64_byte_prefix________________________"
+
+  @type diamond_detection_message() :: %{
+          :type =>
+            :start
+            | :start_trigger
+            | :diamond_detection
+            | :trigger
+            | :delete_ref
+            | :delete_trigger_ref,
```
mat-hek, line 92: I'm not a fan of these names :P there's `start`, `start_trigger` and `trigger`, hard to tell the difference. `diamond_detection` is pretty generic. They should at least correspond to the terms used in the algorithm description (proper search?), ideally be mentioned there

lib/membrane/core/element/diamond_detection_controller.ex
mat-hek: I see it's now constructed once, but I'm still unconvinced about prepending the prefix. Handling ref-counted binaries strains GC, and copying may be more efficient if the binary is small, so I wouldn't force either way, at least without benchmarks.

lib/membrane/core/element/state.ex
```diff
@@ -79,6 +86,12 @@ defmodule Membrane.Core.Element.State do
             handle_demand_loop_counter: 0,
             pads_to_snapshot: MapSet.new(),
             playback_queue: [],
+            diamond_detection: %{
+              serialized_component_path: nil,
+              ref_to_path: %{},
+              trigger_refs: MapSet.new(),
+              postponed?: false
```
mat-hek, line 93: NIT: You can move the type and initialization to the DiamondDetection module

lib/membrane/core/element/diamond_detection_controller.ex
FelonEkonom: It is supposed to be a translation of "właściwe szukanie"

mat-hek: Got it, 'proper' is more like 'correct'. The best term I know would be 'actual search', though I'd just go for 'search', since the whole process is called 'detection', not 'search', so there's no other 'search' unless I'm missing something ;)

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -30,24 +31,24 @@ defmodule Membrane.Core.Element.DiamondDetectionController do
...
+  #   * if the new spec created a new diamond, this diamond will contain some of
+  #     the links spawned in this spec
   # If the diamond contains a link, it must also contain an element whose output pad
   # is part of this link.
 
-  # After the spec status is set to :done, the parent component that returned the spec will trigger
-  # all elements whose output pads have been linked in this spec. The reference of the trigger
-  # is always set to the spec reference.
+  # After the spec status is set to :done, the parent component that returned the spec will
+  # triggerall elements whose output pads have been linked in this spec (`type: :start_trigger`).
```
mat-hek, line 40:
```suggestion
  # trigger all elements whose output pads have been linked in this spec (`type: :start_trigger`).
```

lib/membrane/core/element/diamond_detection_controller.ex
```diff
@@ -0,0 +1,312 @@
...
+  #   * the element forwards the trigger with the same reference via all input pads working
+  #     in the pull mode (`type: :trigger`)
+  #   * if the element has at least two output pads working in the pull mode, it postpones
+  #     the proper searching that will be spawned from itself (`type: :start_search`). The time
+  #     between postponing and the proper searching is one second. If during this time an element
+  #     is triggered once again with a different reference, it won't cause another postponement
+  #     of the proper searching, this means that at the time there is at most one proper searching
+  #     postponed in the single element
+
+  # (2) Proper searching:
```
mat-hek, line 54: Seems like 'proper' is still there, can we make it just 'searching'?

mat-hek (approved):
Remember to create an issue for warning only when the pipeline stucks

Issue #905, opened 2024-11-06
`:telemetry` public API in `:membrane_core`


FelonEkonom:
The goal is to have a public API describing 
 - when `membrane_core` emits `:telemetry` events
 - what are the types of these events
 - what metadata is passed to each event

`:telemetry` events should allow writing the plugs that might be used for e.g. aggregating metrics, spans or structurized logs

What needs to be done:
- [ ] Take a look on how other open-source libraries or frameworks solve this problem (Phoenix, Ecto, Bandit, Finch, Oban etc.)
- [ ] Propose what `:telemetry` events are reasonably to be emitted from `membrane_core` 
- [ ] What should be passed as the events metadata (proposition: pass there components state and create a public typespec describing the map that contains some fields from the state)
- [x] Wrap emitting `:telemetry` events into macro, that will or will not be resolved to the `:telemetry.execute(...)`, depending on config value (already partially/mostly done on the current `master`)

PR #918, opened 2024-12-06
Telemetry (#905)

wende:
Relates to #905

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -3,9 +3,10 @@ import Config
 if config_env() == :test do
   config :junit_formatter, include_filename?: true
 
-  config :membrane_core, :telemetry_flags, [
+  config :membrane_core, :telemetry_flags, include: [
     :links,
     :inits_and_terminates,
+    :spans
     # {:metrics, [:buffer, :bitrate, :queue_len, :stream_format, :event, :store, :take_and_demand]}
   ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -8,6 +8,7 @@ if config_env() == :test do
       :init,
       :setup,
       :terminate,
-      :playing
+      :playing,
+      :link
     ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,5 +14,6 @@ if config_env() == :test do
       ],
       bin: :all,
       pipeline: :all
-    ]
+    ],
+    metrics: :all
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,4 +2,10 @@ import Config
 
 if config_env() == :test do
   config :junit_formatter, include_filename?: true
+
+  config :membrane_core, :telemetry_flags, [
+    :links,
+    :inits_and_terminates,
+    # {:metrics, [:buffer, :bitrate, :queue_len, :stream_format, :event, :store, :take_and_demand]}
+  ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -16,22 +16,4 @@ if config_env() == :test do
       pipeline: :all
     ],
     datapoints: :all
-
-  # config to test legacy version of telemetry
-  # config :membrane_core, :telemetry_flags, [
-  #   :report_init,
-  #   :report_terminate,
-  #   :report_buffer,
-  #   :report_queue,
-  #   :report_link,
-  #   metrics: [
-  #     :buffer,
-  #     :bitrate,
-  #     :queue_len,
-  #     :stream_format,
-  #     :event,
-  #     :store,
-  #     :take_and_demand
-  #   ]
-  # ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -15,5 +15,23 @@ if config_env() == :test do
       bin: :all,
       pipeline: :all
     ],
-    metrics: :all
+    events: :all
+
+  # config to legacy telemetry
+  # config :membrane_core, :telemetry_flags, [
+  #   :report_init,
+  #   :report_terminate,
+  #   :report_buffer,
+  #   :report_queue,
+  #   :report_link,
+  #   metrics: [
+  #     :buffer,
+  #     :bitrate,
+  #     :queue_len,
+  #     :stream_format,
+  #     :event,
+  #     :store,
+  #     :take_and_demand
+  #   ]
+  # ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -15,7 +15,7 @@ if config_env() == :test do
       bin: :all,
       pipeline: :all
     ],
-    events: :all
+    datapoints: :all
 
   # config to test legacy version of telemetry
   # config :membrane_core, :telemetry_flags, [
--- a/config/config.exs
+++ b/config/config.exs
@@ -17,7 +17,7 @@ if config_env() == :test do
     ],
     events: :all
 
-  # config to legacy telemetry
+  # config to test legacy version of telemetry
   # config :membrane_core, :telemetry_flags, [
   #   :report_init,
   #   :report_terminate,
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,7 +7,7 @@ if config_env() == :test do
     include: [
       :init,
       :setup,
-      :terminate,
+      :terminate_request,
       :playing,
       :link
     ]
--- a/config/config.exs
+++ b/config/config.exs
@@ -9,6 +9,7 @@ if config_env() == :test do
       :setup,
       :terminate_request,
       :playing,
-      :link
+      :link,
+      :parent_notification
     ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,6 +14,5 @@ if config_env() == :test do
       ],
       bin: :all,
       pipeline: :all
-    ],
-    metrics: :all
+    ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -3,10 +3,11 @@ import Config
 if config_env() == :test do
   config :junit_formatter, include_filename?: true
 
-  config :membrane_core, :telemetry_flags, include: [
-    :links,
-    :inits_and_terminates,
-    :spans
-    # {:metrics, [:buffer, :bitrate, :queue_len, :stream_format, :event, :store, :take_and_demand]}
-  ]
+  config :membrane_core, :telemetry_flags,
+    include: [
+      :init,
+      :setup,
+      :terminate,
+      :playing
+    ]
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,12 +4,16 @@ if config_env() == :test do
   config :junit_formatter, include_filename?: true
 
   config :membrane_core, :telemetry_flags,
-    include: [
-      :init,
-      :setup,
-      :terminate_request,
-      :playing,
-      :link,
-      :parent_notification
-    ]
+    tracked_callbacks: [
+      element: [
+        :handle_init,
+        :handle_playing,
+        :handle_setup,
+        :handle_terminate_request,
+        :handle_parent_notification
+      ],
+      bin: :all,
+      pipeline: :all
+    ],
+    metrics: :all
 end
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -13,7 +13,6 @@ defmodule Membrane.Core.Bin do
     Parent,
     ProcessHelper,
     SubprocessSupervisor,
-    Telemetry,
     TimerController
   }
 
@@ -96,9 +95,7 @@ defmodule Membrane.Core.Bin do
            {:continue, :setup}}
   def init(options) do
     Utils.log_on_error do
-      Telemetry.report_span :bin, :init do
-        do_init(options)
-      end
+      do_init(options)
     end
   end
 
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -96,7 +96,9 @@ defmodule Membrane.Core.Bin do
            {:continue, :setup}}
   def init(options) do
     Utils.log_on_error do
-      do_init(options)
+      Telemetry.report_span :bin, :init do
+        do_init(options)
+      end
     end
   end
 
@@ -127,7 +129,6 @@ defmodule Membrane.Core.Bin do
     {:ok, resource_guard} =
       SubprocessSupervisor.start_utility(options.subprocess_supervisor, {ResourceGuard, self()})
 
-    Telemetry.report_init(:bin)
     path = Membrane.ComponentPath.get_formatted()
     ResourceGuard.register(resource_guard, fn -> Telemetry.report_terminate(:bin, path) end)
 
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -129,9 +129,6 @@ defmodule Membrane.Core.Bin do
     {:ok, resource_guard} =
       SubprocessSupervisor.start_utility(options.subprocess_supervisor, {ResourceGuard, self()})
 
-    path = Membrane.ComponentPath.get_formatted()
-    ResourceGuard.register(resource_guard, fn -> Telemetry.report_terminate(:bin, path) end)
-
     state =
       %State{
         module: module,
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -128,7 +128,8 @@ defmodule Membrane.Core.Bin do
       SubprocessSupervisor.start_utility(options.subprocess_supervisor, {ResourceGuard, self()})
 
     Telemetry.report_init(:bin)
-    ResourceGuard.register(resource_guard, fn -> Telemetry.report_terminate(:bin) end)
+    path = Membrane.ComponentPath.get_formatted()
+    ResourceGuard.register(resource_guard, fn -> Telemetry.report_terminate(:bin, path) end)
 
     state =
       %State{
--- a/lib/membrane/core/bin.ex
+++ b/lib/membrane/core/bin.ex
@@ -20,7 +20,7 @@ defmodule Membrane.Core.Bin do
 
   require Membrane.Core.Utils, as: Utils
   require Membrane.Core.Message
-  require Membrane.Core.Telemetry
+  require Membrane.Core.LegacyTelemetry, as: LegacyTelemetry
   require Membrane.Logger
 
   @type options :: %{
@@ -125,6 +125,9 @@ defmodule Membrane.Core.Bin do
 
     {:ok, resource_guard} =
       SubprocessSupervisor.start_utility(options.subprocess_supervisor, {ResourceGuard, self()})
+
+    LegacyTelemetry.report_init(:bin)
+    ResourceGuard.register(resource_guard, fn -> LegacyTelemetry.report_terminate(:bin) end)
 
     state =
       %State{
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -127,7 +127,7 @@ defmodule Membrane.Core.CallbackHandler do
   end
 
   @spec exec_callback(callback :: atom, args :: list, handler_params, state) ::
-          {list, internal_state}
+          {list, internal_state} | no_return()
   defp exec_callback(
          callback,
          args,
@@ -137,37 +137,39 @@ defmodule Membrane.Core.CallbackHandler do
     context = context_fun.(state)
     args = args ++ [context, internal_state]
 
-    callback_result =
-      try do
-        fn -> apply(module, callback, args) end
-        |> report_telemetry(callback, args, state, context)
-      rescue
-        e in UndefinedFunctionError ->
-          _ignored =
-            with %{module: ^module, function: ^callback, arity: arity} <- e do
-              reraise CallbackError,
-                      [
-                        kind: :not_implemented,
-                        callback: {module, callback},
-                        arity: arity,
-                        args: args
-                      ],
-                      __STACKTRACE__
-            end
-
-          reraise e, __STACKTRACE__
+    try do
+      fn ->
+        apply(module, callback, args)
+        |> validate_callback_result!(module, callback)
       end
+      |> report_telemetry(callback, args, state, context)
+    rescue
+      e in UndefinedFunctionError ->
+        _ignored =
+          with %{module: ^module, function: ^callback, arity: arity} <- e do
+            reraise CallbackError,
+                    [
+                      kind: :not_implemented,
+                      callback: {module, callback},
+                      arity: arity,
+                      args: args
+                    ],
+                    __STACKTRACE__
+          end
 
-    case callback_result do
-      {actions, _state} when is_list(actions) ->
-        callback_result
-
-      _result ->
-        raise CallbackError,
-          kind: :bad_return,
-          callback: {module, callback},
-          value: callback_result
+        reraise e, __STACKTRACE__
     end
+  end
+
+  defp validate_callback_result!({actions, _state} = callback_result) when is_list(actions) do
+    callback_result
+  end
+
+  defp validate_callback_result!(callback_result, module, callback) do
+    raise CallbackError,
+      kind: :bad_return,
+      callback: {module, callback},
+      value: callback_result
   end
 
   defp report_telemetry(f, callback, args, state, context) do
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -7,6 +7,8 @@ defmodule Membrane.Core.CallbackHandler do
 
   use Bunch
 
+  require Membrane.Core.Telemetry
+  alias Membrane.Core.Telemetry
   alias Membrane.CallbackError
 
   require Membrane.Logger
@@ -136,7 +138,9 @@ defmodule Membrane.Core.CallbackHandler do
 
     callback_result =
       try do
-        apply(module, callback, args)
+        Telemetry.report_span :element, :terminate do
+          apply(module, callback, args)
+        end
       rescue
         e in UndefinedFunctionError ->
           _ignored =
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -143,7 +143,7 @@ defmodule Membrane.Core.CallbackHandler do
           fn ->
             res = {_actions, new_internal_state} = apply(module, callback, args)
 
-            Telemetry.state_result(res, internal_state, new_internal_state, state)
+            Telemetry.state_result(res, args, internal_state, new_internal_state, state)
           end
         )
       rescue
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -133,7 +133,8 @@ defmodule Membrane.Core.CallbackHandler do
          %{context: context_fun},
          %{module: module, internal_state: internal_state} = state
        ) do
-    args = args ++ [context_fun.(state), internal_state]
+    context = context_fun.(state)
+    args = args ++ [context, internal_state]
 
     callback_result =
       try do
@@ -143,7 +144,13 @@ defmodule Membrane.Core.CallbackHandler do
           fn ->
             res = {_actions, new_internal_state} = apply(module, callback, args)
 
-            Telemetry.state_result(res, args, internal_state, new_internal_state, state)
+            Telemetry.state_result(res, %{
+              args: args,
+              module: module,
+              internal_state_before: internal_state,
+              internal_state_after: new_internal_state,
+              component_context: context
+            })
           end
         )
       rescue
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -7,11 +7,10 @@ defmodule Membrane.Core.CallbackHandler do
 
   use Bunch
 
-  require Membrane.Core.Telemetry
-  alias Membrane.Core.Telemetry
   alias Membrane.CallbackError
 
   require Membrane.Logger
+  require Membrane.Core.Telemetry, as: Telemetry
 
   @type state :: %{
           :module => module,
@@ -138,9 +137,14 @@ defmodule Membrane.Core.CallbackHandler do
 
     callback_result =
       try do
-        Telemetry.report_span :element, :terminate do
-          apply(module, callback, args)
-        end
+        Telemetry.report_span(
+          state.__struct__,
+          callback,
+          fn ->
+            apply(module, callback, args)
+            |> Telemetry.state_result(internal_state, state)
+          end
+        )
       rescue
         e in UndefinedFunctionError ->
           _ignored =
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -173,7 +173,7 @@ defmodule Membrane.Core.CallbackHandler do
   end
 
   defp report_telemetry(f, callback, args, state, context) do
-    Telemetry.track_component_callback(
+    Telemetry.track_callback_handler(
       f,
       state.__struct__,
       callback,
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -8,6 +8,7 @@ defmodule Membrane.Core.CallbackHandler do
   use Bunch
 
   alias Membrane.CallbackError
+  alias Membrane.ComponentPath
 
   require Membrane.Logger
   require Membrane.Core.Telemetry, as: Telemetry
@@ -138,21 +139,8 @@ defmodule Membrane.Core.CallbackHandler do
 
     callback_result =
       try do
-        Telemetry.component_span(
-          state.__struct__,
-          callback,
-          fn ->
-            res = {_actions, new_internal_state} = apply(module, callback, args)
-
-            Telemetry.state_result(res, %{
-              args: args,
-              module: module,
-              internal_state_before: internal_state,
-              internal_state_after: new_internal_state,
-              component_context: context
-            })
-          end
-        )
+        fn -> apply(module, callback, args) end
+        |> report_telemetry(callback, args, state, context)
       rescue
         e in UndefinedFunctionError ->
           _ignored =
@@ -180,6 +168,23 @@ defmodule Membrane.Core.CallbackHandler do
           callback: {module, callback},
           value: callback_result
     end
+  end
+
+  defp report_telemetry(f, callback, args, state, context) do
+    Telemetry.span_component_callback(
+      f,
+      state.__struct__,
+      callback,
+      %{
+        callback: callback,
+        callback_args: args,
+        callback_context: context,
+        component_type: state.module,
+        component_path: ComponentPath.get(),
+        internal_state_before: state.internal_state,
+        internal_state_after: nil
+      }
+    )
   end
 
   @spec handle_callback_result(
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -161,7 +161,8 @@ defmodule Membrane.Core.CallbackHandler do
     end
   end
 
-  defp validate_callback_result!({actions, _state} = callback_result) when is_list(actions) do
+  defp validate_callback_result!({actions, _state} = callback_result, _module, _cb)
+       when is_list(actions) do
     callback_result
   end
 
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -137,14 +137,10 @@ defmodule Membrane.Core.CallbackHandler do
 
     callback_result =
       try do
-        Telemetry.report_span(
-          state.__struct__,
-          callback,
-          fn ->
-            apply(module, callback, args)
-            |> Telemetry.state_result(internal_state, state)
-          end
-        )
+        Telemetry.report_span(state.__struct__, callback, fn ->
+          res = {_actions, new_internal_state} = apply(module, callback, args)
+          Telemetry.state_result(res, internal_state, new_internal_state, state)
+        end)
       rescue
         e in UndefinedFunctionError ->
           _ignored =
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -8,7 +8,6 @@ defmodule Membrane.Core.CallbackHandler do
   use Bunch
 
   alias Membrane.CallbackError
-  alias Membrane.ComponentPath
 
   require Membrane.Logger
   require Membrane.Core.Telemetry, as: Telemetry
@@ -174,19 +173,16 @@ defmodule Membrane.Core.CallbackHandler do
   end
 
   defp report_telemetry(f, callback, args, state, context) do
-    Telemetry.span_component_callback(
+    Telemetry.track_component_callback(
       f,
       state.__struct__,
       callback,
-      %{
-        callback: callback,
-        callback_args: args,
-        callback_context: context,
-        component_path: ComponentPath.get(),
-        component_type: state.module,
-        internal_state_before: state.internal_state,
-        internal_state_after: nil
-      }
+      Telemetry.callback_meta(
+        state,
+        callback,
+        args,
+        context
+      )
     )
   end
 
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -179,8 +179,8 @@ defmodule Membrane.Core.CallbackHandler do
         callback: callback,
         callback_args: args,
         callback_context: context,
-        component_type: state.module,
         component_path: ComponentPath.get(),
+        component_type: state.module,
         internal_state_before: state.internal_state,
         internal_state_after: nil
       }
--- a/lib/membrane/core/callback_handler.ex
+++ b/lib/membrane/core/callback_handler.ex
@@ -137,10 +137,15 @@ defmodule Membrane.Core.CallbackHandler do
 
     callback_result =
       try do
-        Telemetry.report_span(state.__struct__, callback, fn ->
-          res = {_actions, new_internal_state} = apply(module, callback, args)
-          Telemetry.state_result(res, internal_state, new_internal_state, state)
-        end)
+        Telemetry.component_span(
+          state.__struct__,
+          callback,
+          fn ->
+            res = {_actions, new_internal_state} = apply(module, callback, args)
+
+            Telemetry.state_result(res, internal_state, new_internal_state, state)
+          end
+        )
       rescue
         e in UndefinedFunctionError ->
           _ignored =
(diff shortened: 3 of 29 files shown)


lib/membrane/core/bin.ex
```diff
@@ -96,7 +96,9 @@ defmodule Membrane.Core.Bin do
            {:continue, :setup}}
   def init(options) do
     Utils.log_on_error do
-      do_init(options)
+      Telemetry.report_span :bin, :init do
+        do_init(options)
+      end
```
FelonEkonom, line 101: Since we will have an appropriate span for `handle_init` in `CallbackHandler`, there is no need for span for `:init` here

FelonEkonom:
I took a brief look at your changes and I commented things that caught my attention

wende:
@FelonEkonom, could you provide an example of how the .config.exs file should look for the end user? This would help me a lot with separation of concerns.
```
config :membrane_core, :telemetry_flags,
    handlers: [
      element: [
        :init,
        :setup,
        :terminate_request,
        :playing,
        :parent_notification
      ],
      bin: :all,
      pipeline: :all
    ],
    metrics: :all
 ```
 Is this example requiring excessive detail from the end user? Or would you consider it an adequate level of granularity?

**Comment by FelonEkonom** [comment]
It could look like 
```

config :membrane_core, :telemetry_flags,
    tracked_callbacks: [
      element: [
        :handle_init,
        :handle_playing,
        :handle_demand
      ],
      bin: :all,
      pipeline: :all
    ],
    metrics: :all
```
or simply
```
config :membrane_core, :telemetry_flags,
    tracked_callbacks: :all,
    metrics: :all
```

@varsill @mat-hek @Noarkhh WDYT guys?

File: lib/membrane/core/telemetry.ex [hunk 2]
```diff
@@ -1,173 +1,196 @@
...
+            raise """
+              Invalid telemetry flag: #{inspect(event)}.
+              Possible values for #{component} are: #{inspect(@possible_handlers[component])}
+            """
+          end
+
+          def handler_measured?(unquote(component), unquote(event)), do: true
+        end
+    end
+  end
```
wende, line 117: It might be a little bit over-engineered, but it gives the performance benefit of head-matching instead of a run-time lookup.

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,196 @@
...
+               %{
+                 old_internal_state: old_intstate,
+                 old_state: sanitize_state_data(old_state),
+                 new_internal_state: new_intstate,
+                 log_metadata: Logger.metadata(),
+                 path: ComponentPath.get()
+               }}
+
+            _other ->
+              raise "Unexpected telemetry span result. Use Telemetry.state_result/3 instead"
```
wende, line 171: A bit more lenient solution would be to issue a warning that telemetry won't be reported and return whatever the end-user returned themselves. 
Use your best judgment which one you prefer

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,196 @@
...
-      quote do
-        _fn = fn ->
-          _unused = unquote(event_name)
-          _unused = unquote(measurement)
-          _unused = unquote(metadata)
-        end
-
-        :ok
-      end
+      measurement
```
wende, line 193: This one is currently unused. There used to be some metrics measured by legacy code of this purpose, but since it wasn't tested anywhere I treated it as dead code and removed it
A list of metrics that you want to measure would be appreciated.

lib/membrane/telemetry.ex
```diff
@@ -61,15 +69,16 @@ defmodule Membrane.Telemetry do
   * `:take_and_demand` - reports the current size of a input buffer when taking buffers and making a new demand
   """
```
wende, line 71: This list is outdated as of this PR since it wasn't in any way tested. Let me know if these are really the metrics you'd like to have measured; as well as any additional ones if applicable

wende:
This is an MVP version of this PR. I'm going to need a list of desirable metrics to be measured, as well as some guidance on how to document `Membrane.Telemetry` and `Membrane.Core.Telemetry` in relation to each other

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,201 @@
...
+    end
+  end
 
-    report_event(
-      event,
-      value,
-      Enum.find(@telemetry_flags, &(&1 == :inits_and_terminates)) != nil
-    )
+  defp sanitize_state_data(state = %struct{}) do
+    Map.take(state, @public_state_keys[struct])
```
FelonEkonom, line 187: It can't be done this way, I believe it would affect the performance too much. 
Let's create a type being part of a public API that will contain some fields from the private state that we want to make public, e.g.

```elixir 
@type pipeline_metadata :: %{
          module: module,
          playback: Membrane.Playback.t(),
          internal_state: Membrane.Pipeline.state() | nil,
          children: ChildrenModel.children(),
          links: %{Link.id() => Link.t()},
          crash_groups: %{CrashGroup.name() => CrashGroup.t()},
          initialized?: boolean(),
          terminating?: boolean(),
          resource_guard: Membrane.ResourceGuard.t(),
          setup_incomplete?: boolean(),
          subprocess_supervisor: pid(),
          optional(any()) => any()
}
```

We will declare that we pass `pipeline_metadata` (you can rename it to sth different) to the `:telemetry` event, but in fact we will pass there `Membrane.Pipeline.Core.State.t()`. It can be done this way because `Membrane.Pipeline.Core.State.t()` always matches type `pipeline_metadata`.

To be more certain, what is the impact of a specific PR on the performance of the framework, we have a job on the CI called `test_performance`, that needs to be manually triggered on CIrcleCI website to be run. I have triggered it, but it looks that it has some problems, that I see for the first time. There is the link to the job I am talking about: https://app.circleci.com/pipelines/github/membraneframework/membrane_core/4128/workflows/536725be-8718-4158-90a5-b7b28ec10189/jobs/14079

FelonEkonom (changes requested):
I haven't finished my review, but I am publishing some of the comments that I've made.

Further review soon 🔜

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,201 @@
...
-            8 *
-              Enum.reduce(
-                List.wrap(unquote(buffers)),
-                0,
-                &(Membrane.Payload.size(&1.payload) + &2)
-              )
-        }
-      end
+      nil ->
+        def handler_measured?(unquote(component), _), do: false
```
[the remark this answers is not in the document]
wende: Good catch. The catch-all case at the bottom was added later

lib/membrane/core/telemetry.ex
wende: So what you mean is to pass the entire state to the end user, but document only the fields that are considered public?

FelonEkonom: Yes

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+          [:membrane, :element, :handle_parent_notification, :exception]
+        ]
+
+      :telemetry.attach_many(ref, spans, &TelemetryListener.handle_event/4, %{
+        dest: self(),
+        ref: ref
+      })
+
+      pid = Testing.Pipeline.start_link_supervised!(spec: child_spec)
+      Process.flag(:trap_exit, true)
```
FelonEkonom, line 167: Use `Testing.Pipeline.start_supervised!/1` and remove `Process.flag(:trap_exit, true)`.

I recommend you to avoid using `Process.flag(:trap_exit, true)` when it is possible, the bugs that might be caused by it sometimes are hard to be found  🐛

test/test_helper.exs
```diff
@@ -3,5 +3,4 @@ System.cmd("epmd", ["-daemon"])
 Node.start(:"my_node@127.0.0.1", :longnames)
 
 ExUnit.configure(formatters: [ExUnit.CLIFormatter, JUnitFormatter])
-# ExUnit.start(exclude: [:long_running], assert_receive_timeout: 500)
```
FelonEkonom, line 6: 👍

lib/membrane/telemetry.ex
FelonEkonom: If we had some `:telemetry` metrics, that were emitted, I wouldn't remove them because of the backwards compatibility, but IMO the main goal is to have new, well designed and documented metrics that will allow us to plug something more or less like [this](https://github.com/membraneframework-labs/membrane_opentelemetry_plugs)

FelonEkonom (commented):
It would be nice to have a piece of documentation that describes
 - how to enable metrics in config
 - what events are emitted and when
 - what is passed as events metadata
but you can wait with it until the code and API are more or less stable within this PR

Another thing is, please take a look at the CI: `test_performance` job doesn't work at all for now, but `test` and `lint` should pass

varsill:
@wende @FelonEkonom performance tests should be fixed, you just need to merge changes from current master branch for them to run properly

lib/membrane/core/callback_handler.ex
```diff
@@ -136,7 +137,14 @@ defmodule Membrane.Core.CallbackHandler do
...
       try do
-        apply(module, callback, args)
+        Telemetry.report_span(
+          state.__struct__,
+          callback,
+          fn ->
+            apply(module, callback, args)
+            |> Telemetry.state_result(internal_state, state)
+          end
+        )
```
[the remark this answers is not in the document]
wende: The purpose of `Telemetry.state_result` was exactly that to avoid circular dependency. I did incorrectly use it after all though. 
Take a look at the most recent version. I don't believe there is a way to avoid circ dependency with a macr. Using this `state_result/2` workaround while not ideal achieves better decoupling.

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+  Test suite for Membrane telemetry public API. It tests if telemetry events are reported
+  properly for all metrics and spans upon attaching to the :telemetry system.
+  """
+
+  use ExUnit.Case, async: false
+  alias Membrane.Testing
+  import Membrane.ChildrenSpec
+  require Logger
+
+  @moduletag :telemetry
```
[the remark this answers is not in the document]
wende: None. Deleted :+1:

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+  end
+
+  defmodule TelemetryListener do
+    def handle_event(name, value, metadata, %{dest: pid, ref: ref}) do
+      pid |> send({ref, :telemetry_ack, {name, value, metadata}})
+    end
+  end
+
+  setup do
+    ref = make_ref()
```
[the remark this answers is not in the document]
wende: Done

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+
+    child_spec =
+      child(:source, %Testing.Source{output: [~c"a", ~c"b", ~c"c"]})
+      |> child(:filter, %TestFilter{target: self()})
```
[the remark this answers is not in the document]
wende: Fixed

wende:
> If we had some :telemetry metrics, that were emitted, I wouldn't remove them because of the backwards compatibility, but IMO the main goal is to have new, well designed and documented metrics that will allow us to plug something more or less like [this](https://github.com/membraneframework-labs/membrane_opentelemetry_plugs)

 I am planning to restore these in accordance with new changes. Please keep in mind though that these were calls at the very heart of the core functionality while being 100% uncovered by tests. I want to make sure these are properly TDDed before they are reintroduced

lib/membrane/telemetry.ex
```diff
@@ -61,15 +56,15 @@ defmodule Membrane.Telemetry do
...
+  @type instrument :: :element | :bin | :pipeline
+  @type event_name :: [atom() | list(atom())]
   @typedoc """
   * component_path - element's or bin's path
   * metric - metric's name
   * value - metric's value
   """
   @type metric_event_value :: %{
-          component_path: String.t(),
+          component_path: Membrane.ComponentPath.path(),
```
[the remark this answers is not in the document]
wende: The change is valid, but right now I realized that it would break backwards compatibility, so I'll reverse it back to the original

lib/membrane/telemetry.ex
wende: I addressed this ☝️ in the main comment below

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,201 @@
...
+      :crash_groups,
+      :initialized?,
+      :terminating?,
+      :resource_guard,
+      :setup_incomplete?,
+      :subprocess_supervisor
+    ]
+  }
+
+  # Verify at compile time that every key is actually present in Core.*.State
```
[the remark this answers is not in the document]
wende: done

lib/membrane/core/telemetry.ex
wende: done

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+          [:membrane, :element, event, step]
+        end
+
+      :telemetry.attach_many(ref, spans, &TelemetryListener.handle_event/4, %{
+        dest: self(),
+        ref: ref
+      })
+
+      pid = Testing.Pipeline.start_link_supervised!(spec: child_spec)
+      :ok = Testing.Pipeline.terminate(pid)
```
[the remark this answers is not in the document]
wende: done

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,189 @@
...
+
+      assert_receive {^ref, :telemetry_ack,
+                      {[:membrane, :pipeline, :handle_setup, :start], value, %{}}}
+
+      assert value.monotonic_time != 0
+
+      assert_receive {^ref, :telemetry_ack,
+                      {[:membrane, :pipeline, :handle_setup, :stop], value, %{}}}
+
+      assert value.duration > 0
```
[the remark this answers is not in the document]
wende: done

test/membrane/telemetry_test.exs
wende: done

lib/membrane/core/element.ex
```diff
@@ -42,12 +42,12 @@ defmodule Membrane.Core.Element do
...
   @type options :: %{
-          module: module,
+          module: module(),
           name: Membrane.Element.name(),
-          node: node | nil,
+          node: node() | nil,
           user_options: Membrane.Element.options(),
           sync: Sync.t(),
-          parent: pid,
+          parent: pid(),
```
wende, line 50: I changed it across core/element, core/bin and core/pipeline to follow elixir style guidelines and be consistent with the rest of the codebase. Let me know if putting it in this PR is overstepping

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,247 @@
...
+
+      assert_receive {^ref, :telemetry_ack, {[:membrane, :metric, :event], _value, _metadata}}
+    end
+
+    test "Queue Length", %{child_spec: child_spec} do
+      ref = setup_pipeline_for(:queue_len, child_spec)
+
+      assert_receive {^ref, :telemetry_ack, {[:membrane, :metric, :queue_len], _value, _metadata}}
+    end
+  end
```
wende, line 231: These tests are a placeholder. I plan on extending them to cover idiosyncrasies of each individual metric

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,179 @@
...
-      # A hack to suppress the 'unused variable' warnings
-      quote do
-        _fn = fn ->
-          _unused = unquote(event_name)
-          _unused = unquote(measurement)
-          _unused = unquote(metadata)
-        end
+  def report_bitrate(buffers, meta \\ %{}) do
+    if metric_measured?(:bitrate) do
+      value = Enum.reduce(List.wrap(buffers), 0, &(&2 + Membrane.Payload.size(&1.payload)))
```
wende, line 147: This currently crashes in tests, because some teasts pass integers as buffer. Which, although allowed by documentation and typespecs, does not implement standard functions like size/1 causing this to crash.

Let me know should I account for integer buffers here or add an implementation of `Membrane.Payload` for integers, or change the tests that are failing

lib/membrane/core/telemetry.ex
wende: I reimplemented metrics for: `[link, buffer, bitrate, queue, stream_format, event, store]`

lib/membrane/telemetry.ex
```diff
@@ -51,6 +53,12 @@ defmodule Membrane.Telemetry do
   Additionally one can control which metric values should get measured by passing an option of format :
   `{:metrics, [list of metrics]}`
 
+  Available spans are:
+  * `:init` - Time spent on executing components initialization handler
+  * `:terminate` - Time spent on executing components termination handler
+  * `:setup` - Time spent on executing components setup handler
+  * `:playing` - Time spent on executing components playing handler
+
```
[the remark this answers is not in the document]
wende: No longer relevant

lib/membrane/core/callback_handler.ex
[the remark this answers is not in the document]
wende: Not to mention there is little to no benefit of using a macro here, I don't think it's warranted but i'm open to discussion

lib/membrane/core/legacy_telemetry.ex
wende, line 1: This file is almost entirely identical to what it was before this PR. Small adjustments were made to make it more resilient, discontinued telemetry events are rerouted to this module. Events that appear in both implementations are routed based on telemetry config indicating legacy format or the new one.

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,277 @@
...
+        assert_receive {^ref, :telemetry_ack,
+                        {[:membrane, :element, ^event, :stop], meta, %{path: [_, ^element_type]}}},
+                       1000
+
+        assert meta.duration > 0
+      end
+    end
+  end
+
+  describe "Telemetry reports pipelines'" do
```
wende, line 94: There's unfortunately no tests for bin's reports since i couldn't find anything like Testing.Bin consistent with Testing.Pipeline and Testing.Element

lib/membrane/core/element.ex
FelonEkonom: It is fine

lib/membrane/telemetry.ex
```diff
@@ -4,84 +4,151 @@ defmodule Membrane.Telemetry do
 
   Membrane uses [Telemetry Package](https://hex.pm/packages/telemetry) for instrumentation and does not store or save any measurements by itself.
 
-  It is user's responsibility to use some sort of metrics reporter
-  that will be attached to `:telemetry` package to consume and process generated measurements.
+  It is user's responsibility to use some sort of event consumer and metric reporter
```
FelonEkonom, line 7:
```suggestion
  It is user's responsibility to use some sort of event handler and metric reporter
```

lib/membrane/telemetry.ex
```diff
@@ -4,84 +4,151 @@ defmodule Membrane.Telemetry do
...
   The following events are published by Membrane's Core with following measurement types and metadata:
 
-    * `[:membrane, :metric, :value]` - used to report metrics, such as input buffer's size inside functions, incoming events and received stream format
-        * Measurement: `t:metric_event_value/0`
-        * Metadata: `%{}`
+    * `[:membrane, :element | :bin | :pipeline, handler, :start | :stop | :exception]` -
+    where handler is any possible handle_* function defined for a component.
+      * `:start` - when the handler begins execution
+      * `:stop` - when the handler finishes execution
+      * `:exception` - when the handler crashes during execution
```
FelonEkonom, line 17: `handler` in `:telemetry` documentation is used as a name of a function that is attached to an event, so I would avoid using this word in our documentation to describe things different than that 

https://hexdocs.pm/telemetry/telemetry.html

lib/membrane/telemetry.ex
```diff
@@ -4,84 +4,151 @@ defmodule Membrane.Telemetry do
...
-  Available metrics are:
+  Available events are:
+  * `:link` - reports the number of links created in the pipeline
   * `:buffer` - number of buffers being sent from a particular element
-  * `:bitrate` - total number of bits carried by the buffers mentioned above
   * `:queue_len` - number of messages in element's message queue during GenServer's `handle_info` invocation
   * `:stream_format` - indicates that given element has received new stream format, value always equals '1'
   * `:event` - indicates that given element has received a new event, value always equals '1'
   * `:store` - reports the current size of a input buffer when storing a new buffer
-  * `:take_and_demand` - reports the current size of a input buffer when taking buffers and making a new demand
```
FelonEkonom, line 61: If I understand correctly, these two should be restored

FelonEkonom (changes requested):
Please fix `lint` CI job. 
Take a look at `test_performance` - there is a decrease in performance on this branch. Optimising code around callback execution should bring the biggest performance boost.

Please, check if you can request performance tests on CI on your own and let me know if you will meet any problems with it.

Add a very short description of this PR on the top of `CHANGELOG.md` file.

lib/membrane/core/telemetry.ex
FelonEkonom: Generating a metric only if a payload is binary is OK (like you did it now)

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,238 @@
...
+  alias Membrane.Pad
+  alias Membrane.Element.WithOutputPads
+  alias Membrane.Element.WithInputPads
   alias Membrane.ComponentPath
 
-  require Membrane.Pad
+  require Membrane.Core.LegacyTelemetry, as: LegacyTelemetry
 
-  @telemetry_flags Application.compile_env(:membrane_core, :telemetry_flags, [])
+  @type telemetry_result() :: {:telemetry_result, {any(), list(), map(), map(), map()}}
```
mat-hek, line 22: 🤔

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,238 @@
...
+                 path: ComponentPath.get()
+               }}
+
+            _other ->
+              raise "Unexpected telemetry span result. Use Telemetry.state_result/3 instead"
+          end
+        end
+      )
+    else
+      {:telemetry_result, {result, _, _, _, _}} = f.()
```
mat-hek, line 181: This wraps every callback with an additional lambda call even if telemetry is disabled. Since the `:telemetry.span` function basically emits `start` and `stop` events, wouldn't it be better to emit them manually?

lib/membrane/telemetry.ex
```diff
@@ -4,84 +4,151 @@ defmodule Membrane.Telemetry do
...
+          internal_state: Pipeline.state() | nil,
+          children: ChildrenModel.children(),
+          links: %{Link.id() => Link.t()},
+          crash_groups: %{CrashGroup.name() => CrashGroup.t()},
+          initialized?: boolean(),
+          terminating?: boolean(),
+          resource_guard: ResourceGuard.t(),
+          setup_incomplete?: boolean(),
+          subprocess_supervisor: pid()
         }
```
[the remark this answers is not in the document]
mat-hek: I see some fields here that are not part of the callback context, like `pads_info`. Others have different names, such as `pads_data`. I don't think we should make too many internal fields public, as they may change

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,277 @@
...
+                      {[:membrane, :element, :handle_parent_notification, :start], _meta,
+                       %{path: [_, ":filter"]}}},
+                     1000
+
+      assert_receive {^ref, :telemetry_ack,
+                      {[:membrane, :element, :handle_parent_notification, :exception], meta,
+                       %{path: [_, ":filter"]}}},
+                     1000
+
+      assert meta.duration > 0
```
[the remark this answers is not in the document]
wende: It's calculated on monotonic time so it's impossible to be 0. But i don't have a strong feeling about changing it to >=

lib/membrane/core/telemetry.ex
wende: That's true, i didn't consider it to be deal-breaking since the lambda calls are virtually free. I could change it to a macro to make it more inlined at compile time, didn't want to prematurely optimize though

lib/membrane/core/telemetry.ex
wende: Yeah, it's not very self explanatory, is it 😅

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,238 @@
...
-                &(Membrane.Payload.size(&1.payload) + &2)
-              )
-        }
-      end
+  if @legacy? do
+    Logger.warning("""
+      Legacy telemetry is deprecated and will be removed in the next major release.
+      Please update your configuration to use the new telemetry system.
+    """)
+  end
```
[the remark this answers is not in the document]
wende: It can be tested by uncommenting what's in the config. I did run it to confirm it works as intended and it didn't seem to have a problem with that

test/membrane/telemetry_test.exs
wende: Yeah I recovered it when doing some trial and error tests. Forgot to remove it after all. On it

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,126 @@
...
+    for element_type <- @paths,
+        event <- @events do
+      test "#{element_type}/#{event}", %{ref: ref} do
+        element_type = unquote(element_type)
+        event = unquote(event)
+
+        assert_receive {^ref, :telemetry_ack,
+                        {[:membrane, :element, ^event, :start], meta, %{path: [_, ^element_type]}}}
+
+        assert meta.system_time > 0
```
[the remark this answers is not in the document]
wende: This is returned from `:telemetry.span` so I don't have control over the contents of this result. It does however return both system_time and monotonic_time

lib/membrane/core/telemetry.ex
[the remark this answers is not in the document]
wende: No longer relevant

config/config.exs
```diff
@@ -2,4 +2,36 @@ import Config
...
+  #   metrics: [
+  #     :buffer,
+  #     :bitrate,
+  #     :queue_len,
+  #     :stream_format,
+  #     :event,
+  #     :store,
+  #     :take_and_demand
+  #   ]
+  # ]
```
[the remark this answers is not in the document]
wende: I think I want to leave it in some documentation since it makes it much easier to manually test behaviour on legacy settings. Any idea where I could put it?

lib/membrane/core/callback_handler.ex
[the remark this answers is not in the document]
wende:
```
def apply_within_span(lambda, callback, callback_args, internal_state, state) do
  report_span(
    state.__struct__,
    callback,
    fn -> apply(lambda) |> state_result(internal_state, state) end
  )
end
```

This solution has the problem that internal state passed this way would be an old internal state, instead of the new one

test/membrane/telemetry_test.exs
wende: Done

lib/membrane/core/element.ex
```diff
@@ -206,10 +206,7 @@ defmodule Membrane.Core.Element do
   @impl GenServer
   def handle_info(message, state) do
     Utils.log_on_error do
-      Telemetry.report_metric(
-        :queue_len,
-        :erlang.process_info(self(), :message_queue_len) |> elem(1)
-      )
+      Telemetry.report_queue_len(self())
```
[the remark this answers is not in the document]
wende: All calls that were removed go directly to LegacyTelemetry so that they can be easily identified to be removed later on. Those that are both in the old and the new one go to Telemetry that redirects the calls based on the config having old or new format. When the legacy telemetry is removed this will allow it only to delete lines with LegacyTelemetry as well as the small block of redirection in new Telemetry

lib/membrane/telemetry.ex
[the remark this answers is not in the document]
wende: This is a contentious issue. Because internal state is available both before and after the call . Higher state however is not, so merging them together may be confusing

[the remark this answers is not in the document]
wende: @mat-hek Could you revise the metadata above to remove some fields that are unnecessary? I'm not familiar with state of the components enough to make that decision myself

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,277 @@
...
+
+      assert_receive {^ref, :telemetry_ack,
+                      {[:membrane, :pipeline, :handle_setup, :start], meta, %{}}}
+
+      assert meta.monotonic_time != 0
+
+      assert_receive {^ref, :telemetry_ack,
+                      {[:membrane, :pipeline, :handle_setup, :stop], meta, %{}}}
+
+      assert meta.duration > 0
```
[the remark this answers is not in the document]
wende: Good catch. I made the change without verifying its validity. Thanks 👍

lib/membrane/telemetry.ex
wende: 1. Take was removed by me incorrectly. Which would break backwards compatibility so good catch
2. Demand is an interesting metric because although defined it was actually never implemented anywhere in the codebase. Should I add it myself?

config/config.exs
```diff
@@ -2,4 +2,36 @@ import Config
...
+        :handle_init,
+        :handle_playing,
+        :handle_setup,
+        :handle_terminate_request,
+        :handle_parent_notification
+      ],
+      bin: :all,
+      pipeline: :all
+    ],
+    events: :all
```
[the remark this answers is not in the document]
wende: The word :metrics is terribly confusing. Since there is no metric. They are telemetry logs or telemetry events. Metrics would be something that aggregates them, which we don't do. Not to mention the whole Telemetry.Metrics confusion on top of that.

> Do I understand correctly:
Legacy API is still supported, but there is a new API that covers new callback events and legacy events as well?
Correct me if I am wrong

Depending on the format of configuration, new telemetry will reroute calls to the old module if it discovers configuration to be of old telemetry system.
```elixir
...
if @legacy? do
      do_legacy_telemetry(event_name, lazy_block)
else
...
```

If the format is consistent with new changes it will use the new Telemetry instead

lib/membrane/core/legacy_telemetry.ex
```diff
@@ -0,0 +1,185 @@
...
+      quote do
+        %{
+          component_path: ComponentPath.get_formatted() <> "/",
+          metric: "bitrate",
+          value:
+            8 *
+              Enum.reduce(
+                List.wrap(unquote(buffers)),
+                0,
+                &(Membrane.Core.LegacyTelemetry.get_payload_size(&1.payload) + &2)
```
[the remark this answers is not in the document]
wende: done

lib/membrane/core/legacy_telemetry.ex
```diff
@@ -0,0 +1,185 @@
...
+        %{
+          parent_path: Membrane.ComponentPath.get_formatted(),
+          from: inspect(unquote(from).child),
+          to: inspect(unquote(to).child),
+          pad_from:
+            Membrane.Core.LegacyTelemetry.__get_public_pad_name__(unquote(from).pad_ref)
+            |> inspect(),
+          pad_to:
+            Membrane.Core.LegacyTelemetry.__get_public_pad_name__(unquote(to).pad_ref)
+            |> inspect()
```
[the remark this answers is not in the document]
wende: done

test/membrane/telemetry_test.exs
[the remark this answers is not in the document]
FelonEkonom: so please assert on `monotonic_time`

lib/membrane/core/callback_handler.ex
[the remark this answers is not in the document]
FelonEkonom: You can get the new one from the value returned from `apply(lambda)`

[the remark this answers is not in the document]
FelonEkonom: For me it is overcomplicated and I still think that it is possible to write functions in `Membrane.Core.Telemetry` in a way that they will be much more intuitive to read. We can do pair-coding session to do so

test/membrane/telemetry_test.exs
```diff
@@ -0,0 +1,279 @@
...
+      test "#{element_type}/#{event}", %{ref: ref} do
+        element_type = unquote(element_type)
+        event = unquote(event)
+
+        assert_receive {^ref, :telemetry_ack,
+                        {[:membrane, :element, ^event, :start], results,
+                         %{component_path: [_, ^element_type]}}},
+                       1000
+
+        assert results.monotonic_time < 0
```
FelonEkonom, line 85: Why do you check, if `monotonic_time` is smaller than zero? Shouldn't it be a positive number?

lib/membrane/telemetry.ex
FelonEkonom: It is still undone

lib/membrane/telemetry.ex
FelonEkonom: Don't add it if it hasn't been implemented before, leave it as it was

lib/membrane/telemetry.ex
```diff
@@ -168,6 +105,28 @@ defmodule Membrane.Telemetry do
...
+
+  @typedoc """
+  Metadata included with each telemetry event
+  """
+  @type event_metadata :: %{
+          event: event_type(),
+          component_path: ComponentPath.path(),
+          component_metadata: any()
+        }
+
```
FelonEkonom, line 129: This is for legacy events, right?

lib/membrane/core/element/stream_format_controller.ex
```diff
@@ -33,7 +33,7 @@ defmodule Membrane.Core.Element.StreamFormatController do
     withl pad: {:ok, data} <- PadModel.get_data(state, pad_ref),
           playback: %State{playback: :playing} <- state do
       %{direction: :input} = data
-      Telemetry.report_stream_format(stream_format, %{pad_ref: inspect(pad_ref)})
+      Telemetry.report_stream_format(stream_format, inspect(pad_ref))
```
FelonEkonom, line 36: Shouldn't it be in legacy telemetry?

lib/membrane/telemetry.ex
wende: I changed it to "component handler" to distinguish between the two

lib/membrane/core/telemetry.ex
```diff
@@ -1,173 +1,284 @@
...
-        %{
-          component_path: ComponentPath.get_formatted() <> "/" <> (unquote(log_tag) || ""),
-          metric: Atom.to_string(unquote(metric)),
-          value: unquote(value)
+  @type telemetry_callback_metadata() :: %{
+          args: list(any()),
+          module: module(),
+          internal_state_before: any(),
+          internal_state_after: any(),
+          component_context: Membrane.Telemetry.component_context()
```
FelonEkonom, line 29: please add `module`, component_path and `setup_incomplete?` (the last one is the value stored in state) to `Membrane.Element/Bin/Pipeline.CallbackContext.t()` and to values returned from functions defined in  `Membrane.Core.Element/Bin/Pipeline.CallbackContext`

Since `module` will be inside callback_context, there will be no need to have it explicitly in `telemetry_callback_metadata`

lib/membrane/telemetry.ex
```diff
@@ -4,95 +4,152 @@ defmodule Membrane.Telemetry do
...
   """
-  @type metric_event_value :: %{
-          component_path: String.t(),
-          metric: String.t(),
-          value: integer()
-        }
+  @type component_type :: :element | :bin | :pipeline
+
+  @type component_context ::
+          Element.CallbackContext.t() | Bin.CallbackContext.t() | Pipeline.CallbackContext.t()
```
FelonEkonom, line 58: Rename it to `callback_context`

test/membrane/telemetry_test.exs
wende: On systems I checked at the monotonic time was always negative. But I did some research and it appears it can be positive on some systems. I'll change it to just be different than 0 to make sure it doesn't fail randomly for some

lib/membrane/telemetry.ex
FelonEkonom: Add a mention, that `events` will be removed in `v2.0`

lib/membrane/telemetry.ex
```diff
@@ -4,95 +4,152 @@ defmodule Membrane.Telemetry do
...
+        }
+
+  @typedoc """
+  Metadata included with each telemetry event
+  """
+  @type event_metadata :: %{
+          event: event_type(),
+          component_path: ComponentPath.path(),
+          component_metadata: any()
+        }
```
FelonEkonom, line 128: Do you have any idea what prefix could be added to names of these two types, to avoid confusion if it has anything to do with newly added callback events?

FelonEkonom (changes requested):

lib/membrane/core/callback_handler.ex
```diff
@@ -136,50 +137,57 @@ defmodule Membrane.Core.CallbackHandler do
...
+      callback,
+      %{
+        callback: callback,
+        callback_args: args,
+        callback_context: context,
+        component_path: ComponentPath.get(),
+        component_type: state.module,
+        internal_state_before: state.internal_state,
+        internal_state_after: nil
+      }
```
FelonEkonom, line 189: Let's move creating this map into `Telemetry.span_component_callback/4`

lib/membrane/telemetry.ex
```diff
@@ -54,84 +56,82 @@ defmodule Membrane.Telemetry do
...
-          component_module: module(),
+  @type callback_span_metadata :: %{
+          callback: atom(),
+          callback_args: [any()],
+          callback_context: callback_context(),
           component_path: ComponentPath.path(),
+          component_type: component_type(),
           internal_state_before: Element.state() | Bin.state() | Pipeline.state(),
-          internal_state_after: Element.state() | Bin.state() | Pipeline.state()
+          internal_state_after: Element.state() | Bin.state() | Pipeline.state() | nil
```
FelonEkonom, line 79: Let's rename fields `internal_state_before` and `internal_state_after` to `state_before_callback` and `state_after_callback`.

The name `internal_state` occurs only inside `membrane_core` private modules and nowhere beyond them. In every other place it is called just `state`

config/config.exs
```diff
@@ -15,7 +15,7 @@ if config_env() == :test do
       bin: :all,
       pipeline: :all
     ],
-    events: :all
+    datapoints: :all
 
   # config to test legacy version of telemetry
   # config :membrane_core, :telemetry_flags, [
```
FelonEkonom, line 21: I would remove the legacy config

lib/membrane/core/telemetry.ex
```diff
@@ -166,108 +160,71 @@ defmodule Membrane.Core.Telemetry do
...
-          any()
-  def component_span(component_type, callback, f) do
-    component_type = state_module_to_atom(component_type)
+  @spec span_component_callback(
+          (-> CallbackHandler.callback_return() | no_return()),
+          module(),
+          atom(),
+          Telemetry.callback_span_metadata()
+        ) :: CallbackHandler.callback_return() | no_return()
+  def span_component_callback(f, component_module, callback, meta) do
```
FelonEkonom, line 171:
```suggestion 
 def apply_callback_within_span(callback_lambda, component_module, callback, meta) do
```
IMO name of this function is little weird

lib/membrane/telemetry.ex
```diff
@@ -8,36 +8,38 @@ defmodule Membrane.Telemetry do
...
 
   ## Instrumentation
-  The following events are published by Membrane's Core with following measurement types and metadata:
+  The following telemetric events are published by Membrane's Core components:
 
     * `[:membrane, :element | :bin | :pipeline, handler, :start | :stop | :exception]` -
     where handler is any possible handle_* function defined for a component.
       * `:start` - when the handler begins execution
       * `:stop` - when the handler finishes execution
       * `:exception` - when the component's handler crashes during execution
```
FelonEkonom, line 17:
```suggestion
      * `:start` - when the components begins callback's execution
      * `:stop` - when the components finishes callback's execution
      * `:exception` - when the component crashes during callback's execution
```

FelonEkonom (changes requested):
@wende are you planning on making any major changes to the documentation?

wende:
> @wende are you planning on making any major changes to the documentation?
Not unless they're needed. I want to have this merged as soon as possible and the documentation can always be improved upon without breaking changes

lib/membrane/core/telemetry.ex
wende: It uses span as a verb, but I'll find a better name for it

wende: What do you think of "track_callback_handler" it has less ambiguous verb, it still specifies it's for a callback and also adds handler to be more descriptive and consistent with "callback_tracked?/1" function name

config/config.exs
```diff
@@ -2,4 +2,18 @@ import Config
...
+        :handle_init,
+        :handle_playing,
+        :handle_setup,
+        :handle_terminate_request,
+        :handle_parent_notification
+      ],
+      bin: :all,
+      pipeline: :all
+    ],
+    datapoints: :all
```
mat-hek, line 18: IIRC, this enables the old, deprecated metrics. Why do we add a new config key for a deprecated feature?

wende: It's not for deprecated metrics. It's for those that were renewed. So:
link, buffer, queue_len, stream_format, event, store and take (no bitrate, demand, init or terminate)

And it's only for those to be tested

mat-hek (approved):

lib/membrane/telemetry.ex
FelonEkonom: > Add a mention, that events will be removed in v2.0

Undone

lib/membrane/telemetry.ex
```diff
@@ -4,95 +4,152 @@ defmodule Membrane.Telemetry do
...
 
   ## Instrumentation
-  The following events are published by Membrane's Core with following measurement types and metadata:
+  The following telemetric events are published by Membrane's Core components:
 
-    * `[:membrane, :metric, :value]` - used to report metrics, such as input buffer's size inside functions, incoming events and received stream format
-        * Measurement: `t:metric_event_value/0`
-        * Metadata: `%{}`
+    * `[:membrane, :element | :bin | :pipeline, handler, :start | :stop | :exception]` -
+    where handler is any possible handle_* function defined for a component.
```
FelonEkonom, line 14:
```suggestion
    * `[:membrane, :element | :bin | :pipeline, callback, :start | :stop | :exception]` -
    where `callback` is any possible handle_* function defined for a component.
```

FelonEkonom (approved):
Fix these two things and it should be fine

PR #922, opened 2024-12-18
Incorporate Funnel, Tee and Fake.Sink

FelonEkonom:
Closes https://github.com/membraneframework/membrane_core/issues/871


lib/membrane/funnel.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Membrane.Funnel do
+  @moduledoc """
+  Element that can be used for collecting data from multiple inputs and sending it through one
+  output. When a new input connects in the `:playing` state, the funnel sends
+  `#{inspect(__MODULE__)}.NewInputEvent` via output.
```
mat-hek, line 5:
```suggestion
  Element that can be used for collecting data from multiple inputs and sending it through one
  output.
  
  When a new input connects in the `:playing` state, the funnel sends
  `#{inspect(__MODULE__)}.NewInputEvent` via output.
```

lib/membrane/tee.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Membrane.Tee do
+  @moduledoc """
+  Element for forwarding buffers to at least one output pad
+
+  It has one input pad `:input` and 2 output pads:
+  * `:output` - is a dynamic pad which is always available and works in pull mode
+  * `:output_copy` - is a dynamic pad that can be linked to any number of elements (including 0) and works
```
mat-hek, line 7: While `copy` made (a bit of) sense along with the `master` pad, now it doesn't IMO. Maybe `push_output`?

lib/membrane/fake_sink.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Membrane.Fake.Sink do
```
mat-hek, line 1: What about

```suggestion
defmodule Membrane.FakeSink do
```

FelonEkonom: In analogical places in core (`Testing.Sink`, `Debug.Filter`) we use dots

PR #962, opened 2025-04-04
Fix order of handle_child_terminated and handle_crash_group_down callback invocation

varsill:
This PR:
* Makes sure that all `handle_child_terminated` callbacks for children in crash group are called before `handle_crash_group_down` of that group
* Adds test checking that assertions
* Fix a formatting issue in documentation for the testing pipeline

TODO:
- [x] Make sure that terminated children is not in the context of `handle_child_terminated`

### DIFF
--- a/lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
+++ b/lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
@@ -52,14 +52,7 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
     end
   end
 
-  def handle_crash_group_member_death(child_name, %CrashGroup{} = group, crash_reason, state) do
-    state =
-      if group.detonating? do
-        state
-      else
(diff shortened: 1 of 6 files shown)


mat-hek (approved):

lib/membrane/core/parent/child_life_controller/crash_group_utils.ex
```diff
@@ -71,7 +64,19 @@ defmodule Membrane.Core.Parent.ChildLifeController.CrashGroupUtils do
...
+
+  def maybe_detonate_crash_group(
+        _child_name,
+        %CrashGroup{detonating?: true} = _group,
+        _crash_reason,
+        state
+      ),
+      do: state
+
+  def maybe_detonate_crash_group(crash_initiator, %CrashGroup{} = group, crash_reason, state) do
```
FelonEkonom, line 79: you can match on `detonating?: false` and add a guard `reason != :normal` to have 2 cases instead of 3

FelonEkonom (approved):

PR #1012, opened 2025-10-23
Plugins docs

Noarkhh:


scripts/elixir/update_packages_list.exs
```diff
@@ -250,39 +278,80 @@ header = """
...
+
+        markdown_link = "[#{package.name}](#{package.url})"
+
+        readme_result = """
+        #{if acc.is_header_present, do: "", else: header}\
+        | #{markdown_link} | #{prefix}#{package.description} | #{hex_badge} #{hexdocs_badge} |\
+        """
+
+        docs_result = """
+        ## #{package.name}
```
mat-hek: let's add a comment that the file is automatically generated and shouldn't be edited

scripts/elixir/update_packages_list.exs
```diff
@@ -307,4 +332,53 @@ File.read!(readme_path)
...
+
+      file_path =
+        Path.join(packages_docs_path, prefix <> filename <> ".md") |> String.replace(" ", "_")
+
+      if filename != "Plugins" do
+        File.write!(
+          file_path,
+          "<!-- Generated code, do not edit. See `scripts/elixir/update_packages_list.exs`. -->\n"
+        )
+      end
```
mat-hek: Couldn't we instead make this `reduce` generate a list of 'files', where each file is a name and content, and then iterate over it and write to actual files? It would be cleaner and we could just skip empty files

PR #1013, opened 2025-10-24
Include demos in docs

Noarkhh:


mix.exs
```diff
@@ -97,14 +98,42 @@ defmodule Membrane.Mixfile do
...
+
+    demo_url =
+      Path.join(demos_url, Path.basename(demo_path))
+
+    {demo_title, [url: demo_url]}
+  end
+
   defp reformat_tutorial_title(filename) do
     {first_letter, rest} =
       filename
```
mat-hek: I missed this previously, but you can use `String.capitalize` instead

mat-hek: The docs for `String.capitalize` say you should use `:string.titlecase` in that case :P

mat-hek (approved):

PR #1030, opened 2025-11-19
Only search packages that have hexdocs

Noarkhh:


scripts/elixir/update_packages_list.exs
```diff
@@ -170,7 +178,7 @@ header = """
 """
 
 generated_code_comment =
-  "<!-- Generated code, do not edit. See `scripts/elixir/update_packages_list.exs`. -->"
+  "<!-- #{generated_code_comment} -->"
```
mat-hek: let's not reassign the variable, just add the markers in line 209

mat-hek (approved):

PR #1033, opened 2025-12-03
Timestamps guide

Noarkhh:


mat-hek (commented):
Sounds great. I'm missing some summary/guidelines, that would emphasise that:
- DTS should always be monotonic, while PTS for video with B-frames can be non-monotonic
- Filters should always forward timestamps
- If a filter doesn't use timestamps, it should forward them, but not rely on them being set
- Sources should attach timestamps to buffers whenever they're known
- Whenever possible, elements should rely on timestamps instead of framerate or audio duration calculated from the stream
- If an element queues buffers in its state (or uses a library that does so), it should make sure that the timestamps for the output buffers are the same as for the corresponding input buffers
- You should make sure that calculations on timestamps don't introduce an accumulating error. Prefer using rationals (Ratio library) to floats.
- Elements should generate deterministic output timestamps for better testability
- If an element transforms N input buffers into M output buffers, each of the output buffers should have either:
  - the timestamp of the first of the input buffers (even if only a part of it was used to construct the output buffers)
  - more precise timestamps, if it's possible to calculate them
- Timestamps are harder than they seem and are the source of many bugs, including:
  - Audio/video desynchronization
  - Stream hanging (due to waiting indefinitely to process/play a buffer because of a wrong timestamp)
  - Stream stalls (e.g. due to processing a real-time stream slightly faster than real time)
  - Memory leaks (e.g. due to processing a real-time stream slightly slower than real time and indefinite buffering)
  - Video flickering (due to incorrect handling of B-frames)
  - Audio cracking
  Therefore, the operations on timestamps should be given a lot of care and be well-tested

guides/useful_concepts/timestamps.md
```diff
@@ -0,0 +1,137 @@
...
+* If a filter doesn't use timestamps, it should forward them, but not rely on
+them being set.
+* Sources should attach timestamps to buffers whenever they're known.
+* Whenever possible, elements should rely on timestamps instead of
+framerate or audio duration calculated from the stream.
+* If an element queues buffers in its state (or uses a library that does so), it
+should make sure that the timestamps for the output buffers are the same as for
+the corresponding input buffers.
+* You should ensure that calculations on timestamps don't introduce an
+accumulating error. Prefer using rationals (Ratio library) to floats.
```
FelonEkonom: maybe some link to `ratio`?

FelonEkonom (approved):

.circleci/config.yml
```diff
@@ -70,6 +70,8 @@ jobs:
 
     steps:
       - checkout
+      - run: git submodule sync
+      - run: git submodule update --init
```
varsill: Why do we need to download the submodules on CI? 
And if so, shouldn't we also cache them (these repos are quite large)

guides/useful_concepts/timestamps.md
```diff
@@ -0,0 +1,137 @@
...
+for streams with B-frames can be non-monotonic.
+
+## Tips and guidelines
+
+Dealing with timestamps can be complicated and very different depending on the
+use case, so here is some advice on dealing with them:
+
+* Filters should always forward timestamps.
+* If a filter doesn't use timestamps, it should forward them, but not rely on
+them being set.
```
varsill: It's a little bit vague to me, how could a filter "rely on" timestamps being set it and not use them?

mat-hek: This was my idea, and indeed it doesn't make much sense :P

```suggestion
* If a filter doesn't use timestamps, it should still forward them.
* If an element relies on timestamps and they are not set, it should raise a meaningful error.
```

PR #1038, opened 2025-12-10
Update to elixir 1.19

Noarkhh:
I'll suppress warnings from problematic functions, the bug is described here: https://github.com/elixir-lang/elixir/issues/14576

### DIFF
--- a/lib/membrane/core/element.ex
+++ b/lib/membrane/core/element.ex
@@ -103,7 +103,7 @@ defmodule Membrane.Core.Element do
   end
 
   # Suppress false positive dialyzer warnings resulting from 
-  # broken handling of opaque types
+  # broken handling of opaque types - in this case MapSet.t() 
   # https://github.com/elixir-lang/elixir/issues/14576
   @dialyzer {:nowarn_function, do_init: 1}
   defp do_init(options) do
--- a/lib/membrane/core/element/auto_flow_controller.ex
+++ b/lib/membrane/core/element/auto_flow_controller.ex
@@ -219,7 +219,7 @@ defmodule Membrane.Core.Element.AutoFlowController do
   end
 
   # Suppress false positive dialyzer warnings resulting from 
-  # broken handling of opaque types
+  # broken handling of opaque types - in this case MapSet.t() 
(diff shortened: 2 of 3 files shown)


lib/membrane/core/child/pad_spec_handler.ex
```diff
@@ -14,6 +14,7 @@ defmodule Membrane.Core.Child.PadSpecHandler do
   Initializes pads info basing on element's or bin's pads specifications.
   """
   @spec init_pads(Element.State.t()) :: Element.State.t()
+  # @spec init_pads(Element.State1.t()) :: Element.State1.t()
```
mat-hek: leftover

lib/membrane/core/element/state.ex
```diff
@@ -15,38 +15,38 @@ defmodule Membrane.Core.Element.State do
...
+          pads_to_snapshot: MapSet.t(),
+          playback_queue: Membrane.Core.Element.PlaybackQueue.t(),
+          diamond_detection_state: DiamondDatectionState.t(),
+          pads_data: PadModel.pads_data() | nil,
           satisfied_auto_output_pads: MapSet.t(),
           awaiting_auto_input_pads: MapSet.t(),
-          resume_delayed_demands_loop_in_mailbox?: boolean(),
-          diamond_detection_state: DiamondDatectionState.t()
+          auto_input_pads: [Pad.ref()],
+          resume_delayed_demands_loop_in_mailbox?: boolean()
```
mat-hek: what exactly has changed here? :P

lib/membrane/core/element.ex
```diff
@@ -102,6 +102,9 @@ defmodule Membrane.Core.Element do
     end
   end
 
+  # Suppress false positive dialyzer warnings
+  # https://github.com/elixir-lang/elixir/issues/14576
```
mat-hek: Mention which type causes the error whenever suppressing warnings

lib/membrane/core/element/state.ex
Noarkhh: I was debugging the dialyzer and reordered the fields so that they're in the same order in defstruct and spec for convenience

PR #1059, opened 2026-01-29
Mix task for creating component templates

### LINKED ISSUE #1016: Add mix tasks for easier start with membrane
It would be nice to have some mix tasks that create some simple project that the user can build on top of, like `mix.phx.new` or `mix.nerves.new`.

Some of the conepts are:
- Tasks that generate a template for an element of given type with necessary callbacks and things like that.
- Tasks that generate our demos.
- Tasks that create a simple pipeline based on passed arguments what should be the input and the output. It would create a pipeline like the one boombox creates, but that could be then tinkered with.

Noarkhh:


templates/pipeline.ex
```diff
@@ -0,0 +1,108 @@
...
+  # end
+
+  # @impl true
+  # def handle_terminate_request(_ctx, state) do
+  #   {[terminate: :normal], state}
+  # end
+
+  # --------------------------
+  # --- OPTIONAL CALLBACKS ---
+  # --------------------------
```
mat-hek: Why are these callbacks marked as optional? AFAIK all of them are optional 🤔

mat-hek (commented):
I have mixed feelings about generating commented-out code. I think it would be better to use comments to mark things as optional, but no strong preference

lib/mix/tasks/membrane.gen.component.ex
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    defp infer_path_from_module(module) do
+      module
+      |> inspect()
+      |> String.downcase()
+      |> String.split(".")
+      |> List.update_at(-1, &"#{&1}.ex")
+      |> Path.join()
+    end
```
FelonEkonom: Will it convert `AlaMaKota` into `alamakota.ex` or `ala_ma_kota.ex`? I think we want the second option

lib/mix/tasks/membrane.gen.component.ex
```diff
@@ -0,0 +1,85 @@
+[Filter, Endpoint, Sink, Source, Bin, Pipeline]
+|> Enum.map(fn component_type ->
+  component_name = inspect(component_type)
+
+  defmodule Module.concat(Mix.Tasks.Membrane.Gen, component_type) do
+    @shortdoc "Generates a template for a Membrane #{component_name}"
```
FelonEkonom: I am suprised that you can use `component_name` without `unquote`

templates/bin.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Membrane.TemplateBin do
+  @moduledoc """
+  This is a generated template for a Bin. Uncomment the snippets as necessary.
+  """
+  use Membrane.Bin
+
+  # def_input_pad :input,
+  #   accepted_format: _any,
+  #   availability: :on_request | :always,
```
FelonEkonom: What about 
```suggestion
  #   availability: :on_request | :always,  # defaults to :always
```
Here and in other places?

templates/bin.ex
```diff
@@ -0,0 +1,143 @@
...
+
+  # @impl true
+  # def handle_child_terminated(_child, _ctx, state) do
+  #   {[], state}
+  # end
+
+  # --------------------------
+  # --- OPTIONAL CALLBACKS ---
+  # --------------------------
+
```
FelonEkonom: Do we mention anywhere in our docs, that there are optional callbacks in membrane components? If not, we shouldn't mention it here as well. Moreover, it gives an impression that callbacks above should be implemented in every bin, while optional and non-optional callbacks are indistinguishable from the user perspective, because we provide a default implementations of almost all callbacks

FelonEkonom (changes requested):

mat-hek (approved):

test/mix/tasks/membrane.gen_test.exs
```diff
@@ -0,0 +1,39 @@
...
+  use ExUnit.Case
+  alias Mix.Tasks.Membrane.Gen
+
+  Enum.each([Filter, Endpoint, Sink, Source, Bin, Pipeline], fn component_type ->
+    @tag :tmp_dir
+    test "#{inspect(component_type)} template generates correctly", %{tmp_dir: tmp} do
+      Module.concat(Gen, unquote(component_type)).do_run(tmp, [inspect(unquote(component_type))])
+
+      {{:module, module, _contents, __result}, []} =
+        Code.eval_file(Path.join(tmp, Macro.underscore(unquote(component_type)) <> ".ex"))
```
FelonEkonom: It could be such a beautiful `|>` sequence (with `Kernel.<>(".ex")`) 😞

FelonEkonom (approved):

lib/mix/tasks/membrane.gen.component.ex
```diff
@@ -0,0 +1,80 @@
...
+            Module name not provided.
+
+            This task expects a module name, which the created #{unquote(component_type)} will have:
+
+              $ mix membrane.gen.#{String.downcase(unquote(component_name))} My#{unquote(component_name)}
+              
+            """)
+
+          [module_name | _rest] ->
+            Module.concat([module_name]) |> inspect()
```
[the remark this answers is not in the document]
varsill: It works due to a kind of a "side effect" of how `Module.concat/1` treats the strings (it runs `String.to_atom()` under the hood) 🤔 If we were to leave it this way, it definitelly requires a comment.

I did some research on how to determine that a String is a valid alias and that's what I found:
* this is how IEx searches for a valid module names within autocomplete (it resolves alias "pieces" one by one):
https://github.com/elixir-lang/elixir/blob/d33fd8e413fe98e0e54dde7ef94fadb272b1ef67/lib/iex/lib/iex/autocomplete.ex#L511
* We could also use `Code.string_to_quoted()` and check if the returned value is an alias:
```
def proper_module_name?(name) do
  case Code.string_to_quoted(name) do
    {:ok, {:__aliases__, _, _}} -> true
    _ -> false
  end
end
```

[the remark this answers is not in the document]
Noarkhh: did the second thing

varsill (approved):
🥇

PR #1067, opened 2026-02-05
Add task for downloading membrane demos

### LINKED ISSUE #1016: Add mix tasks for easier start with membrane

Noarkhh:
`mix membrane.demo` task pulls specific or all demos available. It can also list available demos with `-l` option and the "manifest" of available demos it the `membrane_demo` readme.


mix.exs
```diff
@@ -46,7 +46,7 @@ defmodule Membrane.Mixfile do
     opts = [
       plt_local_path: "priv/plts",
       flags: [:error_handling, :unmatched_returns],
-      plt_add_apps: [:mix]
+      plt_add_apps: [:mix, :req]
```
varsill: why do we need `:req` here?

varsill (changes requested):

FelonEkonom (commented):
What about adding a test? 🙂

FelonEkonom (changes requested):

FelonEkonom (approved):
Awesome 🎉

lib/mix/tasks/membrane.demo.ex
```diff
@@ -0,0 +1,137 @@
...
+  @shortdoc "Download Membrane demos and examples"
+  @moduledoc """
+  Download Membrane demos and examples into the current directory. Requires `git` installed.
+
+    $ mix membrane.demo [-a] [-l] [-d <repo_dir>] [<demos> ...]
+
+  ## Options
+  * `-l, --list` - List all demos available and their brief descriptions.
+  * `-a, --all` - Pull the repository with all demos. 
+  * `-d, --directory` - Specify a directory where the demos should be placed in.
```
varsill: So what is the default directory? :D

Noarkhh: cwd, it's in the moduledoc

varsill (approved):
🥇

PR #1095, opened 2026-03-09
Fix compilation warnings in test files

FelonEkonom:
Squeeze Klaudiusz like a 🍋


test/membrane/integration/auto_demands_test.exs
```diff
@@ -88,10 +88,9 @@ defmodule Membrane.Integration.AutoDemandsTest do
         Enum.flat_map(payloads, &Enum.map(1..Integer.pow(factor, filters), fn _i -> &1 end))
 
       {in_payloads, out_payloads} =
-        case direction do
-          :up -> {payloads, mult_payloads}
-          :down -> {mult_payloads, payloads}
-        end
+        if direction == :up,
+          do: {payloads, mult_payloads},
+          else: {mult_payloads, payloads}
```
mat-hek, line 93: Why is this needed?

mat-hek (approved):

test/membrane/integration/auto_demands_test.exs
FelonEkonom: I guess the previous version caused some warnings

mat-hek: This seems more like a hack than a solution to me. I'd try figuring out the actual reason for the warnings.

PR #1103, opened 2026-03-19
Improve docs

### LINKED ISSUE #1089: Deprecate `Membrane.FilterAggregator`
This filter was suggested by Gemini when asked what's the go-to way of stacking filters into one, but the context called for a simple Bin.

### LINKED ISSUE #1091: [docs] What's the difference between `Filter` and `Endpoint`?
It's very hard to find explicit information about how a `Membrane.Endpoint` differs from a `Membrane.Filter`. We decided that the best modification to the docs in this case would be to add a section in `Membrane.Endpoint`'s docs that compares it to `Membrane.Filter` and discusses the differences.

### LINKED ISSUE #1092: [docs] More info on terminating pipelines gracefully
I've had trouble searching through the docs for info on how to gracefully terminate a pipeline after all its elements received an EOS. Example:

```
Mix.install([
  {:membrane_generator_plugin, ">= 0.0.0"},
  {:membrane_core, "~> 1.0"}
])

defmodule MyPipeline do
  use Membrane.Pipeline

  @impl true
  def handle_init(_ctx, _opts) do
    spec = [
      child(
        :generator,
        %Membrane.SilenceGenerator{
          stream_format: %Membrane.RawAudio{
            channels: 1,
            sample_rate: 16_000,
            sample_format: :s16le
          },
          duration: Membrane.Time.milliseconds(100)
        }
      )
      |> child(:sink, Membrane.Fake.Sink)
    ]

    {[spec: spec], nil}
  end
end

{:ok, supervisor, pipeline} = Membrane.Pipeline.start_link(MyPipeline, [])
Process.monitor(supervisor)

defmodule Loop do
  def loop() do
    receive do
      {:DOWN, _ref, :process, _pid, _reason} ->
        IO.puts("Pipeline down")
      other ->
        IO.inspect(other, label: "received unknown message")
    end

    loop()
  end
end

Loop.loop()
```

The solution was to include `def handle_element_end_of_stream(:sink, _ctx, state), do: {[terminate: :normal], state}` - this would be triggered when the sink receives an EOS, which for this simple case would mean `:end_of_stream` messages covered the entire pipeline. This might not be the case for more complex pipeline topologies. I was under the impression the framework would automatically detect and terminate when all elements receive EOS, so a section in the guides going through the process of adding termination mechanisms would provide DX value.

Noarkhh:


lib/membrane/filter_aggregator.ex
```diff
@@ -13,6 +13,8 @@ defmodule Membrane.FilterAggregator do
   * their pads cannot use manual demands
   * the first filter must make demands in buffers
   """
+  @moduledoc deprecated: "This module is deprecated, use Membrane.Bin instead"
```
varsill: [NIT] I am not sure, but I feel like it requires a little bit more of a context, something like: "if you need to aggregate children into a single component, use `Membrane.Bin`.
(the point is that' I am not sure if it's always a valid clue to use `Membrane.Bin` when one tried using the FilterAggregator)

lib/membrane/exceptions.ex
```diff
@@ -112,8 +112,21 @@ defmodule Membrane.ActionError do
     "This action cannot be returned from the #{callback} callback."
   end
 
+  defp format_reason({:invalid_element, element_type}) do
+    "This action cannot be returned by a #{element_type |> Atom.to_string() |> String.capitalize()}."
+  end
+
+  defp format_reason({:invalid_element_callback_combination, element_type, callback}) do
```
varsill: Shouldn't it be more like `component_type`, to make it generic for elements/bins/pipelines?

lib/membrane/endpoint.ex
```diff
@@ -8,14 +8,24 @@ defmodule Membrane.Endpoint do
...
+
+  Although Endpoints may seem similair to [Filters](`m:Membrane.Filter`), there are 
+  some important differences. While Filters can be thought as parts of a Pipeline that 
+  take in data, process it in some way, and then pass it along, Endpoints create 
+  "holes" in the pipeline. They behave more like a [Sink](`m:Membrane.Sink`) and a 
+  [Source](`m:Membrane.Source`) in a single element - media they consume and produce are
+  parts of different streams. The main consequence of this is the fact that 
+  they separate the flow control of the pipeline ahead of them and behind them, 
+  and in turn their input pads behave like ones of a Sink, and their output pads behave
+  like ones of a Source.
```
varsill:
```suggestion
  Although Endpoints may seem similair to [Filters](`m:Membrane.Filter`), there are 
  some important differences. While Filters can be thought as parts of a Pipeline that 
  take in data, process it in some way, and then pass it along, Endpoints create 
  "holes" in the pipeline. They behave more like a [Sink](`m:Membrane.Sink`) and a 
  [Source](`m:Membrane.Source`) combined in a single element - media they consume and produce are parts of different streams. The main consequence of this is the fact that 
  they separate the flow control of the pipeline ahead of them and behind them, 
  as their input pads behave like those of a Sink, and their output pads behave
  like those of a Source.
```

lib/membrane/core/child/pads_specs.ex
```diff
@@ -325,6 +338,67 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
+              ":max_instances can only be set for pads with :availability set to :on_request"
+
+            other_config_key ->
+              "#{inspect(other_config_key)} is not a valid config key"
+          end)
+
+        "Invalid config keys:\n" <> invalid_keys_reasons
+
+      other ->
+        "Invalid pad config: #{other}"
```
varsill:
```suggestion
        "Invalid pad config: #{inspect(other)}"
```

lib/membrane/core/child/pads_specs.ex
```diff
@@ -325,6 +338,67 @@ defmodule Membrane.Core.Child.PadsSpecs do
...
+  defp format_parsing_error(reason, direction, component) do
+    case reason do
+      {:config_not_keyword, _config} ->
+        "Pad config not a keyword list"
+
+      {:config_duplicates, duplicates} ->
+        "Duplicate keys in pad config: #{inspect(duplicates)}"
+
+      {:config_field, {:key_not_found, :flow_control}}
+      when component != :filter and direction == :output ->
```
varsill, line 356: [JUST A THOUGHT :D] It looks as if we are repeating the conditions specified in `parse_pad_spec()` - perhaps `Bunch.Config.parse()` should accept error message as one of its arguments?

lib/membrane/core/element/action_handler.ex
```diff
@@ -139,20 +141,34 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+  def handle_action(
+        {:buffer, {pad_ref, buffers}} = action,
+        _cb,
+        _params,
+        %State{type: type} = state
+      )
+      when is_pad_ref(pad_ref) do
+    if type == :sink do
+      raise ActionError, action: action, reason: {:invalid_element, type}
+    else
```
varsill: It looks as if there is a plenty of places where you check if `type == <some element type>` and then you raise `{:invalid_element, type}` - wouldn't it be possible to add guards checking expected types of elements in these `handle_action` clauses (just as it use to be) and raise `invalid_element` in fallback `handle_action` implementation?

varsill (changes requested):

lib/membrane/core/child/pads_specs.ex
Noarkhh: yeah, but our conditions in `parse_pad_spec` were simple conds. What could that error message argument look like anyway?

lib/membrane/endpoint.ex
```diff
@@ -8,14 +8,23 @@ defmodule Membrane.Endpoint do
...
   provide more complex solutions.
+
+  Although Endpoints may seem similair to [Filters](`m:Membrane.Filter`), there are 
+  some important differences. While Filters can be thought as parts of a Pipeline that 
+  take in data, process it in some way, and then pass it along, Endpoints create 
+  "holes" in the pipeline. They behave more like a [Sink](`m:Membrane.Sink`) and a 
+  [Source](`m:Membrane.Source`) combined in a single element - media they consume and produce are parts of different streams. The main consequence of this is the fact that 
+  they separate the flow control of the pipeline ahead of them and behind them, 
+  as their input pads behave like those of a Sink, and their output pads behave
+  like those of a Source.
```
FelonEkonom: I sounds nice, however I would modify it in some places:
1. Tell also why Filters and Endpoints look similar (because they both can have input and output pads).
2. Tell what intuition stands for Filters and what intuition stands for Endpoints and how each of them are supposed to work.
3. Tell what are actual, concrete differences between these two, it means: Filter can accept auto flow control on output pads, while Endpoint cannot and Filter cannot redemand in `handle_demand`, while Endpoints can. You could also add a reason why it is done such way.

Without the distinction between what intuitively each of component types does and what are actual, concrete differences this paragraph can look like a small hocus-pocus for a newbie, at least IMO. If you disagree, tell me WDYT

lib/membrane/pipeline.ex
```diff
@@ -47,7 +48,16 @@ defmodule Membrane.Pipeline do
...
-  ![Observer graph](assets/images/observer_graph.png).
+  ![Observer graph](assets/images/observer_graph.png)
+
+  ## Terminating
+
+  The Pipeline won't terminate automatically - it has to be terminated explicitly 
+  by executing a `t:Membrane.Pipeline.Action.terminate/0` action. The reason for 
+  this is the fact that there is no objective way to tell when a Pipeline should terminate.
+  Even if all of it's children have terminated, it doesn't mean that it should too - more
+  children can be spawned later on.
```
FelonEkonom: We could mention we recommend to terminate a pipeline when a sink (or all sinks) receives EoS

FelonEkonom (changes requested):

lib/membrane/endpoint.ex
```diff
@@ -8,14 +8,34 @@ defmodule Membrane.Endpoint do
...
+
+  Although Endpoints may seem similair to [Filters](`m:Membrane.Filter`) - they both have
+  input and output pads - there are some important differences. While Filters can
+  be thought as parts of a Pipeline that 
+  take in data, process it in some way, and then pass it along, Endpoints create 
+  "holes" in the pipeline. They behave more like a [Sink](`m:Membrane.Sink`) and a 
+  [Source](`m:Membrane.Source`) combined in a single element - media they consume
+  and produce are parts of different streams. 
+
+  For example an example Filter would only modify
```
varsill:
```suggestion
  For instance an exemplary Filter would only modify
```

varsill (approved):
🥇

lib/membrane/core/element/action_handler.ex
```diff
@@ -141,34 +141,35 @@ defmodule Membrane.Core.Element.ActionHandler do
...
   @impl CallbackHandler
-  def handle_action(
-        {:buffer, {pad_ref, buffers}} = action,
-        _cb,
-        _params,
-        %State{type: type} = state
-      )
+  def handle_action({:buffer, {pad_ref, buffers}}, _cb, _params, %State{type: type} = state)
+      when is_pad_ref(pad_ref) and type != :sink do
+    state = %{state | uninterrupted_redemands: 0}
```
FelonEkonom: Shouldn't `uninterrupted_redemands` be in `pad_data`?

FelonEkonom (approved):
Noice

lib/membrane/core/element/action_handler.ex
Noarkhh: It is, but I haven't rebased this branch yet

PR #1106, opened 2026-03-19
Add :broadcast action

varsill:


lib/membrane/core/element/action_handler.ex
```diff
@@ -287,6 +308,17 @@ defmodule Membrane.Core.Element.ActionHandler do
     raise ActionError, action: action, reason: {:unknown_action, Membrane.Element.Action}
   end
 
+  defp broadcast_to_action(:end_of_stream, pad_ref), do: {:end_of_stream, pad_ref}
+  defp broadcast_to_action(%Buffer{} = buffer, pad_ref), do: {:buffer, {pad_ref, buffer}}
+
+  defp broadcast_to_action(data, pad_ref) do
```
FelonEkonom: Add a guard checking if `data` is a struct...

lib/membrane/core/element/action_handler.ex
```diff
@@ -287,6 +308,17 @@ defmodule Membrane.Core.Element.ActionHandler do
     raise ActionError, action: action, reason: {:unknown_action, Membrane.Element.Action}
   end
 
+  defp broadcast_to_action(:end_of_stream, pad_ref), do: {:end_of_stream, pad_ref}
+  defp broadcast_to_action(%Buffer{} = buffer, pad_ref), do: {:buffer, {pad_ref, buffer}}
```
FelonEkonom: let's put `%Buffer{}` in the first match

lib/membrane/core/element/action_handler.ex
```diff
@@ -214,6 +215,26 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+  @impl CallbackHandler
+  def handle_action({:broadcast, data}, cb, params, state) do
+    output_pads =
+      Enum.flat_map(state.pads_data, fn
+        {pad_ref, %{direction: :output}} -> [pad_ref]
+        _other -> []
+      end)
+
+    Enum.reduce(output_pads, state, fn pad_ref, state ->
+      handle_action(broadcast_to_action(data, pad_ref), cb, params, state)
```
FelonEkonom: Please use `|>` operator

lib/membrane/core/element/action_handler.ex
```diff
@@ -214,6 +215,26 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+    end)
```
FelonEkonom: We can join these enums with `|>`

lib/membrane/core/element/action_handler.ex
```diff
@@ -214,6 +215,26 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+  end
+
+  @impl CallbackHandler
+  def handle_action({:broadcast, data}, cb, params, state) do
+    output_pads =
+      Enum.flat_map(state.pads_data, fn
+        {pad_ref, %{direction: :output}} -> [pad_ref]
+        _other -> []
+      end)
```
FelonEkonom: If somebody returns `[broadcast: [many_buffers]]`, we will perform this `Enum.flat_map/2` many times, while we could do it only once. Refactor it, because its a part of the Membrane hot path

lib/membrane/core/element/action_handler.ex
```diff
@@ -287,6 +308,17 @@ defmodule Membrane.Core.Element.ActionHandler do
...
+  defp broadcast_to_action(%Buffer{} = buffer, pad_ref), do: {:buffer, {pad_ref, buffer}}
+
+  defp broadcast_to_action(data, pad_ref) do
+    if Event.event?(data) do
+      {:event, {pad_ref, data}}
+    else
+      {:stream_format, {pad_ref, data}}
+    end
+  end
+
```
FelonEkonom, line 337: ... and another match capturing e.g. maps and raising a meaningful error

FelonEkonom (changes requested):
Please, ensure there is no need to update any guide because of this change

lib/membrane/core/element/action_handler.ex
varsill: done

lib/membrane/core/element/action_handler.ex
varsill: done

lib/membrane/core/element/action_handler.ex
varsill: done

lib/membrane/core/element/action_handler.ex
varsill: done

lib/membrane/core/element/action_handler.ex
varsill: done

mat-hek (commented):
let's deprecate `forward` action and remove it from elements that live in core

FelonEkonom (approved):
🥇

mat-hek (approved):

PR #1113, opened 2026-04-13
Migrate CI from CircleCI to GitHub Actions

khamilowicz:
Replace the CircleCI elixir orb with reusable workflows from
membraneframework/membrane_actions.

## Changes

- Remove `.circleci/config.yml`
- Add `.github/workflows/ci.yml` with `build_test`, `test`, and `lint` jobs
- Add `.github/workflows/release.yml` with the same jobs plus `hex_publish`
- Add `.github/workflows/performance.yml` for benchmark comparisons

CI runs on every push and can be manually triggered via `workflow_dispatch`
(used for PRs from outside the organization). Hex publish triggers on `v*`
tags using the `HEX_API_KEY` secret.

Submodules are checked out recursively to support the `guides/membrane_tutorials`
and `guides/membrane_demo` submodules. Tests run with `--include long_running`
across 4 parallel shards.

Also includes a fix for demo path resolve logic — when `demo_dir` is `.`, the
sparse-checkout path was incorrectly constructed as `./demo_name` instead of
just `demo_name`, causing sparse-checkout to fail for demos located at the
repo root.

## Performance test

The performance workflow replaces the CircleCI manual approval gate with
`workflow_dispatch` — trigger it manually from the Actions tab on the branch
you want to benchmark. It compares benchmark results between the feature branch
and master using Docker.

Since this is a public repository, GitHub by default does not spin self-hosted runners. It has to be explicitly allowed in settings.

## Omissions

- **Parallel test splitting by timings** — the original config used CircleCI's
  timing-based test splitting. Tests are now split by round-robin file distribution
  across 4 shards.


FelonEkonom:
The version on master is now 1.3.0, so this PR should bump it to 1.3.1

khamilowicz:
@FelonEkonom I also added the GitHub release job, we'll see if the result is satisfactory.

varsill (approved):

mat-hek (approved):

PR #1125, opened 2026-04-22
Postpone handle child terminated

varsill:
This PR:
* makes sure that `handle_child_terminated` is only called when given components subprocess supervisor is already dead


lib/membrane/core/subprocess_supervisor.ex
```diff
@@ -250,22 +250,33 @@ defmodule Membrane.Core.SubprocessSupervisor do
...
+    state
   end
 
   defp handle_exit(%{role: :component} = data, reason, state) do
     Process.exit(data.supervisor_pid, :shutdown)
-    Message.send(state.parent_component, :child_death, [data.name, reason])
+
+    update_in(state, [:children, data.supervisor_pid], fn supervisor_data ->
+      Map.put(supervisor_data, :pending_child_death, {data.name, reason})
+    end)
```
mat-hek: How about

```suggestion
    put_in(state.children[data.supervisor_pid].pending_child_death, {data.name, reason})
```

Also, initialise the `pending_child_death` to `nil` in `handle_call(:start_component, ...)`

lib/membrane/core/subprocess_supervisor.ex
```diff
@@ -250,22 +250,33 @@ defmodule Membrane.Core.SubprocessSupervisor do
...
 
       :error ->
-        :ok
+        case data do
+          %{pending_child_death: {child_name, death_reason}} ->
+            Message.send(state.parent_component, :child_death, [child_name, death_reason])
+
+          _no_pending_death ->
+            :ok
+        end
```
mat-hek: Let's move the `case` out of the parent `case`. Both can also be refactored to `with`s

mat-hek (approved):

lib/membrane/core/subprocess_supervisor.ex
varsill: done

lib/membrane/core/subprocess_supervisor.ex
varsill: done

PR #1142, opened 2026-06-01
Macro escape default pad option. Use :httpc instead of Req in Mix task membrane.demo. Bump to v1.3.3

varsill:


lib/membrane/core/options_specs.ex
```diff
@@ -127,8 +127,10 @@ defmodule Membrane.Core.OptionsSpecs do
             quote(do: &inspect/1)
           )
 
+        default_ast = opt_definition[:default]
```
mat-hek, line 130: don't we want to escape it as well?

lib/mix/tasks/membrane.demo.ex
```diff
@@ -51,11 +51,24 @@ defmodule Mix.Tasks.Membrane.Demo do
   end
 
   defp list_available_demos() do
-    {:ok, _apps} = Application.ensure_all_started(:req)
-    response = Req.get!(@demos_readme_url)
+    {:ok, _apps} = Application.ensure_all_started(:inets)
+    {:ok, _apps} = Application.ensure_all_started(:ssl)
```
mat-hek:
```suggestion
    {:ok, _apps} = Application.ensure_all_started([:inets, :ssl])
```

mat-hek (approved):

PR #1145, opened 2026-06-08
Adjust to flat structure of membrane_demo

Noarkhh:
membrane_demo no longer has `livebooks` directory, so we can just assume every demo is at the root of the repo.


lib/mix/tasks/membrane.demo.ex
```diff
@@ -120,20 +118,15 @@ defmodule Mix.Tasks.Membrane.Demo do
 
   defp copy_demo(target_dir, demo_name) do
     repo_dir = Path.join(target_dir, "membrane_demo")
```
mat-hek: let's rename to `repo_path` for consistency

mat-hek (approved):

PR #1146, opened 2026-06-10
Fix 1.20.0 warnings

kidq330:


lib/membrane/core/bin/pad_controller.ex
```diff
@@ -85,23 +85,21 @@ defmodule Membrane.Core.Bin.PadController do
...
-
-      Pad.is_dynamic_pad_ref(pad_ref) ->
-        Message.send(state.parent_pid, :child_pad_removed, [state.name, pad_ref])
-        PadModel.delete_data!(state, pad_ref)
+  def remove_pad(pad_ref, state) when Pad.is_dynamic_pad_ref(pad_ref) do
+    if PadModel.assert_instance(state, pad_ref) == :ok do
+      Message.send(state.parent_pid, :child_pad_removed, [state.name, pad_ref])
+      PadModel.delete_data!(state, pad_ref)
+    else
+      state
```
mat-hek, line 95:
```suggestion
      # This is to handle the case when a bin pad is removed and then the bin removes its child linked to this pad
      state
```

lib/membrane/core/element/diamond_detection_controller/path_in_grapth.ex
```diff
@@ -3,7 +3,7 @@ defmodule Membrane.Core.Element.DiamondDetectionController.PathInGraph do
 
   defmodule Vertex do
     @moduledoc false
-    require Membrane.Pad, as: Pad
+    alias Membrane.Pad, as: Pad
```
mat-hek: Applies to all `alias A.B.C, as: C` cases

```suggestion
    alias Membrane.Pad
```

test/membrane/telemetry_test.exs
```diff
@@ -62,8 +62,10 @@ defmodule Membrane.TelemetryTest do
     end
   end
 
+  defp identity(term), do: term
+
   setup do
-    if Telemetry.legacy?() do
+    if Telemetry.legacy?() |> identity() do
```
mat-hek:
```suggestion
    if Telemetry.legacy?() |> Function.identity() do
```

lib/membrane/core/component.ex
```diff
@@ -26,13 +26,11 @@ defmodule Membrane.Core.Component do
...
-    alias Membrane.Core.{Bin, Element, Pipeline}
-
     callback_context_module =
       case state do
-        %Element.State{} -> Element.CallbackContext
-        %Bin.State{} -> Bin.CallbackContext
-        %Pipeline.State{} -> Pipeline.CallbackContext
+        %Membrane.Core.Element.State{} -> Membrane.Core.Element.CallbackContext
+        %Membrane.Core.Bin.State{} -> Membrane.Core.Bin.CallbackContext
+        %Membrane.Core.Pipeline.State{} -> Membrane.Core.Pipeline.CallbackContext
```
FelonEkonom, line 33: why?

kidq330: `alias Membrane.Core.{Bin, Element, Pipeline}` is an expression that returns `[Bin, Element, Pipeline]`. Dialyzer was complaining that the value returned by this expression is unused

mat-hek (approved):

